PR #23, opened 2021-12-02
Clarify source of key index, from sylabs 456

rhattersley:
This pulls in sylabs PR

* sylabs/singularity#456

> This is purely a change to the documentation to clarify that the `--keyidx` key index parameter for the `singularity sign ...` command relates to the _private_ keys, so the `--secret` argument is necessary when obtaining the index from `singularity key list`.


cclerget:
superseded by #24

DrDaveD:
Thanks, @rhattersley, this PR was perfectly fine, but in parallel @edytuk had imported your sylabs PR and yours happened to hit a transient e2e test service error and the other one didn't.

PR #36, opened 2021-12-04
Dockerfile to support running of various apptainer test suites locally on a variety of hosts.

### LINKED ISSUE #33: support running of various apptainer test suites locally on a variety of hosts
# where we are at

current test suites need to run as privileged user, with mixed privileges between root and sudo access
to drivers and other resources on an Ubuntu Host in the GitHub CI process.

# desired state

developers need to be able to run each element of the test suite locally, regardless of their host OS, and
preferably without managing their own guest OS via virtualization

PaulCharlton:
Dockerfile to support running of various apptainer test suites locally on a variety of hosts.

YMMV -- this is a good start and likely needs work from additional contributors.

The current readme is in comments at the end of the Dockerfile

Fixes #33
Ultimately, most of the tests need to be audited and separated by those which need sudo or root privilege

### DIFF
--- a/tools/Dockerfile.local.testing
+++ b/tools/Dockerfile.local.testing
@@ -1,0 +1,46 @@
+FROM ubuntu:20.04
+ENV TZ=Pacific/Honolulu
+RUN ln -snf "/usr/share/zoneinfo/${TZ}" /etc/localtime && echo "${TZ}" > /etc/timezone
+RUN apt-get update && apt-get -y install tzdata
+ENV LC_ALL=en_US.UTF-8
+ENV LANG=en_US.UTF-8
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential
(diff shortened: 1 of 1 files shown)


gmkurtzer (approved):

DrDaveD:
> The current readme is in comments at the end of the Dockerfile

What do you mean by that?   I think there does need to be some sort of reference to this file somewhere along with some description of how to make use of it or what still needs to be done with it.

PaulCharlton:
@DrDaveD re: `I think there does need to be some sort of reference to this file somewhere along with some description of how to make use of it or what still needs to be done with it.`.  the material described is embedded in the file, as is common practice with brief Dockerfile

ikaneshiro (approved):

DrDaveD:
Ok I'm sorry I missed that, but still it seems like there should be a reference to this file somewhere else, otherwise how will people find it?

PR #72, opened 2021-12-14
Fixes #71 - remove dead, unused and unreachable code

### LINKED ISSUE #71: remove dead or unreachable code from repo
remove dead or unreachable code from repo

PaulCharlton:
Fixes #71 - remove dead, unused and unreachable code


pkg/util/loop/loop.go
```diff
@@ -31,36 +31,17 @@ type Device struct {
 const (
 	FlagsReadOnly  = 1
 	FlagsAutoClear = 4
-	FlagsPartScan  = 8
-	FlagsDirectIO  = 16
-)
 
 // Loop device encryption types
```
PaulCharlton, line 35: will remove

e2e/internal/e2e/apptainercmd.go
```diff
@@ -159,21 +159,6 @@ func ExpectOutput(mt MatchType, pattern string) ApptainerCmdResultOp {
 	}
 }
 
-// ExpectOutputf tests if the command output stream match the
-// formatted string pattern based on the type of match.
-func ExpectOutputf(mt MatchType, formatPattern string, a ...interface{}) ApptainerCmdResultOp {
```
ikaneshiro, line 164: We should leave all of these unused functions in `e2e/internal/` because they are part of our `e2e` test toolkit and may be valuable in the future

internal/pkg/cgroups/config_linux.go
```diff
@@ -191,13 +191,3 @@ func LoadConfig(confPath string) (config Config, err error) {
 	err = toml.Unmarshal(b, &config)
 	return
 }
-
-// PutConfig takes the content of a CgroupsConfig struct and Marshals it to file
-func PutConfig(config Config, confPath string) (err error) {
```
ikaneshiro, line 196: We should leave this for completeness of the `Get/Set` style api for this struct. This is actually used by this example plugin: https://github.com/apptainer/apptainer/blob/main/examples/plugins/config-plugin/main_linux.go#L54

pkg/image/driver.go
```diff
@@ -66,19 +64,6 @@ type Driver interface {
 // drivers holds all registered image drivers
 var drivers = make(map[string]Driver)
 
-// RegisterDriver registers an image driver by name.
-func RegisterDriver(name string, driver Driver) error {
```
ikaneshiro, line 70: This is important for this package API to allow plugins to register different types of image drivers: https://github.com/apptainer/apptainer/blob/main/examples/plugins/ubuntu-userns-overlay-plugin/main.go#L42

pkg/util/loop/loop.go
```diff
@@ -31,36 +31,15 @@ type Device struct {
 const (
 	FlagsReadOnly  = 1
 	FlagsAutoClear = 4
-	FlagsPartScan  = 8
```
ikaneshiro, line 34: I dont see a reason to remove all of these constants. From my perspective they benefit the package API by having completeness in options available

ikaneshiro (changes requested):

internal/pkg/cgroups/config_linux.go
PaulCharlton: will preserve

e2e/internal/e2e/apptainercmd.go
PaulCharlton: I prefer not to keep things around simply because `someone might use it`

pkg/image/driver.go
PaulCharlton: will preserve -- interesting that the dependency walker I use did not detect this.

pkg/util/loop/loop.go
PaulCharlton: having these constants in our project is a slippery slope at best.    can we find a more authoritative golang package, maintained by kernel developers, that provides these constants?

e2e/internal/e2e/apptainercmd.go
PaulCharlton: can we find a separate GOLANG e2e test framework and leave the maintenance to THAT community ?

ikaneshiro: all of our e2e framework is made ourselves to make sure we can run tests and inspect containers effectively. I am not aware of an e2e test framework in golang that will allow us to do what we are currently doing in our e2e tests and dont think it is worth migrating to a different framework to avoid mainlining something that has been effective and stable to date

pkg/util/loop/loop.go
ikaneshiro: im not aware of any loop package implementation in golang maintained by kernel developers, google or oci container maintainers would be our best hope, but they do not use loop devices in their container lifecycle. We were on our own when developing a lot of this code base because the operations against the OS is not very common in the go community

DrDaveD:
So @PaulCharlton can you please restore the pieces that Ian requested?

e2e/internal/e2e/apptainercmd.go
PaulCharlton: reverting

DrDaveD:
I believe that some of this may not be dead code after #85 is implemented so I think it would be better to hold this PR for that.

pkg/util/loop/loop.go
PaulCharlton: restoring

PR #83, opened 2021-12-17
Update minimum golang version to 1.16.12

DrDaveD:
This updates the minimum golang version to 1.16.12 as needed for security purposes.  It removes the golang-1.15 patch that was applied for rpm builds when needed, and instead if mconfig finds a too-old version in the PATH it expects a golang source tarball to be available to the rpm and compiles that using the older golang.

Based on apptainer/singularity#6227.


DrDaveD:
I'm going to close this for now, until I have fixed CI tests.

PR #89, opened 2021-12-21
Fixes #63 - remove references to remote building

### LINKED ISSUE #63: Remove remaining references to "remote build" in code and man page
After #53 removed remote building, I still see some references to ["remote build"](https://github.com/apptainer/apptainer/search?q=%22remote+build%22&type=code) in the code including "Usage" strings and also in `man apptainer-build`.

Attn: @PaulCharlton

PaulCharlton:
Fixes #63 - remove references to remote building

### DIFF
--- a/cmd/internal/cli/build.go
+++ b/cmd/internal/cli/build.go
@@ -13,7 +13,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"runtime"
 
 	"github.com/apptainer/apptainer/docs"
 	"github.com/apptainer/apptainer/internal/pkg/util/fs"
@@ -28,7 +27,6 @@ var buildArgs struct {
 	sections      []string
 	bindPaths     []string
 	mounts        []string
-	arch          string
 	libraryURL    string
 	keyServerURL  string
 	webURL        string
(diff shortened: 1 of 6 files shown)


cmd/internal/cli/build.go
```diff
@@ -105,7 +105,7 @@ var buildArchFlag = cmdline.Flag{
 	Value:        &buildArgs.arch,
 	DefaultValue: runtime.GOARCH,
 	Name:         "arch",
-	Usage:        "architecture for remote build",
+	Usage:        "architecture for {obsolete}",
```
DrDaveD, line 108: Could we please remove use of the --arch flag instead?  Otherwise the Usage should be "obsolete" and if somebody tries to use it it should probably print a deprecated warning or an error.

DrDaveD (changes requested):

cmd/internal/cli/build.go
PaulCharlton: re: `remove use of the --arch flag instead` -- that may break existing users with no explanation.

1. it is not `deprecated` ... which implies "soon to be removed", rather than "removed"
2. in light of existing scripts, should error exit with appropriate error

PaulCharlton:
@DrDaveD fixed and green CI. -- mitigated e2e race condition for overlay tests

DrDaveD (approved):

PR #69, opened 2021-12-13
Fixes #61 - remove commercial library references from code

PaulCharlton:
Fixes #61 - remove commercial library references from code

PR #101, opened 2021-12-28
Restore library api

### LINKED ISSUE #85: Restore library:// URI
As discussed on the [mailing list](https://groups.google.com/u/1/a/lbl.gov/g/singularity/c/e-t6Ag-srBQ), support for the library protocol that was removed in #69 can be restored.

DrDaveD:
This PR restores support for the library:// API.  It begins by reverting most of #69 and then renaming the use of sylabs/scs-client-library to apptainer/container-client-library.   It replaces the SylabsCloud default remote with one called DefaultRemote which has https://keys.openpgp.org as the key server, and no library server.  It also changes the pull and push behavior to not use library:// by default but to instead require that a URI transport be supplied.

I suggest reviewing the 3 commits separately so it is more clear what's new and what's old.

- Fixes #85


ikaneshiro (approved):

DrDaveD:
@PaulCharlton Do you want to review this before merging?

examples/library/Apptainer
```diff
@@ -1,4 +1,5 @@
 Bootstrap: library
+Library: https://library.sylabs.io
```
PaulCharlton, line 2: @DrDaveD @gmkurtzer do we need to remove commercial server from our CI tests?

internal/pkg/remote/endpoint/client_test.go
```diff
@@ -18,6 +18,12 @@ func init() {
 	useragent.InitValue("apptainer", "0.1.0")
 }
 
+const (
+	testCloudURI     = "cloud.sylabs.io"
+	testLibraryURI   = "https://library.sylabs.io"
+	testKeyserverURI = "https://keys.sylabs.io"
+)
+
```
PaulCharlton, line 26: @DrDaveD @gmkurtzer do we need to remove commercial server from our CI tests?

cmd/internal/cli/search.go
```diff
@@ -16,8 +16,8 @@ import (
 	"github.com/apptainer/apptainer/internal/pkg/client/library"
 	"github.com/apptainer/apptainer/pkg/cmdline"
 	"github.com/apptainer/apptainer/pkg/sylog"
+	"github.com/apptainer/container-library-client/client"
```
PaulCharlton, line 19: @DrDaveD for consistency, if one import has a name override, all of the imports in different files should have that same name override.

at the moment, some files have `client.` while others have `libClient.`

cmd/internal/cli/delete.go
```diff
@@ -90,20 +94,53 @@ var deleteImageCmd = &cobra.Command{
...
-		pullFrom := args[len(args)-1]
-		transport, ref := uri.Split(pullFrom)
-		if ref == "" {
-			sylog.Fatalf("Bad URI %s", pullFrom)
+		imageRef, err := library.NormalizeLibraryRef(args[0])
+		if err != nil {
+			sylog.Fatalf("Error parsing library ref: %v", err)
 		}
 
-		switch transport {
```
PaulCharlton, line 99: @DrDaveD this switch can still be used to identify transports which do not yet support `delete` functionality

cmd/internal/cli/pull.go
```diff
@@ -180,9 +185,9 @@ func pullRun(cmd *cobra.Command, args []string) {
 		pullTo = args[0]
 		if len(args) == 1 {
 			if transport == "" {
-				pullTo = uri.GetName("oras://" + pullFrom)
+				pullTo = uri.GetName("library://" + pullFrom)
```
PaulCharlton: @DrDaveD @gmkurtzer do we really want `library://` to be default ?

cmd/internal/cli/search.go
```diff
@@ -12,28 +12,27 @@ package cli
...
 	"github.com/apptainer/apptainer/pkg/cmdline"
 	"github.com/apptainer/apptainer/pkg/sylog"
 	"github.com/spf13/cobra"
+	"github.com/sylabs/scs-library-client/client"
 )
 
 var (
-	// SearchRegistryURI holds the base URI to a Registry API instance
-	SearchRegistryURI string
+	// SearchLibraryURI holds the base URI to a Sylabs library API instance
```
PaulCharlton, line 24: @DrDaveD library/registry API ?

cmd/internal/cli/search.go
```diff
@@ -75,15 +74,18 @@ var SearchCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
 	Args:                  cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		searchFor := args[len(args)-1]
-		transport, _ := uri.Split(SearchRegistryURI)
-		switch transport {
```
PaulCharlton, line 80: @DrDaveD this switch can still provide useful insight if incompatible transport

e2e/testdata/regressions/issue_5399.def
```diff
@@ -1,5 +1,5 @@
-bootstrap: oras
```
PaulCharlton, line 1: @DrDaveD @ikaneshiro we need Oras E2E tests ... not sure that we want to revert all to `library://`

internal/pkg/client/progress.go
```diff
@@ -97,7 +97,7 @@ func CopyWithContext(ctx context.Context, dst io.Writer, src io.Reader) error {
 	return err
 }
 
-// DownloadProgressBar is used for chunked library downloads.
+// DownloadProgressBar is used for chunked scs-library-client downloads.
```
PaulCharlton, line 100: @DrDaveD reference should be to the new community library

internal/pkg/remote/endpoint/client_test.go
```diff
@@ -112,7 +110,7 @@ func TestKeyserverClientOpts(t *testing.T) {
 		{
 			name: "Fake cloud",
 			endpoint: &Config{
-				URI: "cloud.nonexistent-xxxx-domain.io",
+				URI: "cloud.inexistent-xxxx-domain.io",
```
PaulCharlton, line 119: @DrDaveD @ikaneshiro this is misspelled and also not the canonical IETF recommended dummy domain of `example.com`

internal/pkg/remote/endpoint/client_test.go
```diff
@@ -136,74 +134,68 @@ func TestKeyserverClientOpts(t *testing.T) {
 }
 
 //nolint:dupl
-func TestRegistryClientConfig(t *testing.T) {
+func TestLibraryClientConfig(t *testing.T) {
```
PaulCharlton, line 143: @DrDaveD @gmkurtzer not sure all of our unit tests should depend on 3rd party commercial server -- this also regresses our newer GHCR server for containers

internal/pkg/remote/endpoint/service.go
```diff
@@ -28,18 +28,18 @@ const defaultTimeout = 10 * time.Second
 // Default Sylabs cloud service endpoints.
 const (
 	SCSDefaultCloudURI     = "cloud.sylabs.io"
-	DefaultApptainerHost   = "ghcr.io"
-	DefaultRegistryURI     = "https://ghcr.io"
+	SCSDefaultLibraryURI   = "https://library.sylabs.io"
```
PaulCharlton: @DrDaveD @gmkurtzer not sure all of our unit tests should depend on 3rd party commercial server -- this also regresses our newer GHCR server for containers

internal/pkg/remote/endpoint/service.go
```diff
@@ -188,22 +188,18 @@ func (ep *Config) GetAllServices() (map[string][]Service, error) {
 }
 
 // GetServiceURI returns the URI for the service at the specified SCS endpoint
-// Examples of services: consent, registry, key, token
+// Examples of services: consent, build, library, key, token
```
PaulCharlton, line 191: @DrDaveD there is no longer a remote build

internal/pkg/util/uri/uri.go
```diff
@@ -53,7 +56,7 @@ func IsValid(source string) (valid bool, err error) {
 }
 
 // GetName turns a transport:ref URI into a name containing the top-level identifier
-// of the image. For example, docker://sylabs.io/lolcow returns lolcow
+// of the image. For example, docker://sylabsio/lolcow returns lolcow
```
PaulCharlton, line 59: @DrDaveD this is not a proper hostname

PaulCharlton (changes requested):
@DrDaveD a few comments for consideration by yourself, Greg and Ian

PaulCharlton:
@DrDaveD @gmkurtzer as I recall the issue with 3rd party commercial servers used for Unit tests and CI is `who is paying for the service`, and is that fair under the new community/organization structure.   The default should be that LF/Apptainer is paying for all such test servers

cmd/internal/cli/pull.go
DrDaveD: That default was removed by a later commit

e2e/testdata/regressions/issue_5399.def
DrDaveD: This was changed back by a later commit

cmd/internal/cli/search.go
DrDaveD: I think that's outside the scope of this PR.  If that's to be a consistent policy, someone should examine the whole codebase for that.  Make an issue for that if you lie.

examples/library/Apptainer
gmkurtzer: Similar to #99, if there is a publicly available service we can leverage, I'd prefer to use that service.

It seems as though Singularity Hub is still up, but I don't recall that it is using the Library API (I think it is using SHUB API), what about any public hosted versions of Singularity Registry?

DrDaveD: The singularity hub is shut down.  The URL forwards to an archive of some sort, certainly not using Library API.  I agree I would prefer to not use any commercial server for testing, but I don't think there are any.  @vsoch told me there weren't any publicly available sregistry servers.

gmkurtzer: It does not seem that there are any other alternatives to validate the Library API so I think this might be the only option.

PaulCharlton: @gmkurtzer @DrDaveD was there any mention in the new lib-client community (Vanessa, et al) of standing up a publicly available registry?

Did anything happen with artifactory? https://www.jfrog.com/jira/browse/RTFACT-15479?jql=text%20~%20%22singularity%22

vsoch: Most people that use sregistry do so because they want a private registry. It’s easy to use an OCI registry if you just need public.

gmkurtzer: @vsoch The point and question is regarding the Library API, there is no OCI registry that implements that protocol which is why Paul asked if there will be a publicly available implementation of it.

If not, we will probably have to use the Sylabs implementation, or remove the CI testing.

My feeling is we should be testing the Library API, and if there is no non-commercial endpoint for it, then we should use the commercial endpoint, and hope that we don't get ourselves blocked or rate limited (which has much precedent at this point).

vsoch: Yep that makes sense to me! I was just providing further comment on @PaulCharltons's question about a publicly available registry (unlikely to exist).

gmkurtzer:
> @DrDaveD @gmkurtzer as I recall the issue with 3rd party commercial servers used for Unit tests and CI is `who is paying for the service`, and is that fair under the new community/organization structure. The default should be that LF/Apptainer is paying for all such test servers

Apptainer does not have a "bank account" at this point, so it is hard for the project to "pay" for services, which is why we really should rely 100% (or as much as possible) on open and freely available standards and implementations.

If there is rate limiting or we are blocked from using any particular commercial services to maintain compatibility, then unfortunately we won't be able to test those interfaces. So let's do the best we can, while options exist.

cmd/internal/cli/search.go
DrDaveD: I removed 'Sylabs'.  The function does not support general registry searches, only library searches.

cmd/internal/cli/search.go
DrDaveD: I don't think it does provide anything useful in this function.  The option that fills in the SearchLibraryURI variable is called `--library`.  As it is, if somebody tries to put a transport on that for example by doing
```
apptainer search --library oras://ghcr.io/apptainer lolcow
```
It currently gives an error
```
FATAL:   Error initializing library client: unsupported protocol scheme "oras"
```

internal/pkg/remote/endpoint/client_test.go
DrDaveD: Replaced with cloud.example.com

internal/pkg/remote/endpoint/client_test.go
DrDaveD: A remote endpoint is really almost all about the library interface, with the rest little bit for key servers.  (It used to also be about build services, but that's removed).  It doesn't do anything related to a general registry service.  Those are always accessed directly by URL, they are not registered endpoints.

internal/pkg/remote/endpoint/service.go
DrDaveD: Fixed

internal/pkg/util/uri/uri.go
DrDaveD: Right but it is a proper docker URI.  `apptainer pull` with the listed URI works and not with the old one.  sylabs.io does not support the docker API.

internal/pkg/client/progress.go
DrDaveD: Fixed

cmd/internal/cli/delete.go
DrDaveD: I added a check to catch if someone tries to use an unsupported protocol scheme in a URI passed in as a delete imageRef.  Without that it was giving an incomprehensible error message.

cmd/internal/cli/search.go
DrDaveD: I changed my mind, because I decided somebody might accidentally include a URI on the search query argument.  I added an error if any URI protocol scheme is provided there.

PR #105, opened 2021-12-30
Support more key server variation allowed for in HKP spec

### LINKED ISSUE #104: fixes to key server response parser for v1.0
...

_Originally posted by @PaulCharlton in https://github.com/apptainer/apptainer/pull/99#pullrequestreview-840666657_

DrDaveD:
This supports more possible variations in key server API that is allowed for in HKP spec and some other code improvements, as suggested in @PaulCharlton's [pr #99 review](https://github.com/apptainer/apptainer/pull/99#pullrequestreview-840666657).

- Fixes #104

### DIFF
--- a/pkg/sypgp/sypgp.go
+++ b/pkg/sypgp/sypgp.go
@@ -612,13 +612,15 @@ func formatMROutput(mrString string, longOutput bool) (int, []byte, error) {
 	gotName := false
 	for _, l := range lines {
 		fields := strings.Split(strings.TrimSpace(l), ":")
-		if fields[0] == "info" {
+		switch fields[0] {
+
+		case "info":
 			var err error
 			numKeys, err = strconv.Atoi(fields[2])
 			if err != nil {
 				return -1, nil, fmt.Errorf("unable to check number of keys")
 			}
-		} else if fields[0] == "pub" {
+		case "pub":
 			if !first {
 				if !gotName {
 					// there was a pub without uid; end line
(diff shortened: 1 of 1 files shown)


pkg/sypgp/sypgp.go
```diff
@@ -636,17 +638,32 @@ func formatMROutput(mrString string, longOutput bool) (int, []byte, error) {
...
 				keyStatus := ""
-				if fields[6] == "r" {
-					keyStatus = "[revoked]"
-				} else if fields[6] == "d" {
-					keyStatus = "[disabled]"
-				} else if fields[6] == "e" {
-					keyStatus = "[expired]"
-				} else {
+				if len(fields) >= 6 {
+					flags := fields[6]
```
PaulCharlton, line 655: @DrDaveD I believe this should have `strings.ToLower`

pkg/sypgp/sypgp.go
```diff
@@ -612,13 +612,15 @@ func formatMROutput(mrString string, longOutput bool) (int, []byte, error) {
 	gotName := false
 	for _, l := range lines {
 		fields := strings.Split(strings.TrimSpace(l), ":")
-		if fields[0] == "info" {
+		switch fields[0] {
+
+		case "info":
 			var err error
 			numKeys, err = strconv.Atoi(fields[2])
```
PaulCharlton, line 619: @DrDaveD this may need the `if len(fields) >=2 && fields[2] != ""`

PaulCharlton (changes requested):

pkg/sypgp/sypgp.go
DrDaveD: I didn't see anything the spec allowing it to be uppercase, did you?

pkg/sypgp/sypgp.go
DrDaveD: I see the spec saying that the "info" line is optional but I don't read it to say that the count is optional if the line is there.

PaulCharlton: the last paragraph of section 5 states:
`Note that empty fields are allowed.  For example, a key with no
   expiration date would have the <expirationdate> field empty.  Also,
   a keyserver that does not track a particular piece of information
   may leave that field empty as well.  Colons for empty fields on the
   end of each line may be left off, if desired.`

I read it as the count is option if there are no keys returned for a given search.

PaulCharlton: see also, the section on `info` wherein `If this optional line is not included, __or the version information is
   not supplied__, the version number is assumed to be 1`

DrDaveD: If there are no results then the keyserver definitely knows that so I think it is not an optional piece of information, it should return zero.

ikaneshiro: The wording of the spec seems to indicate that any field could be empty and assuming all following fields are empty as well, colon splitting and accessing a slice at an index could cause an out of bounds panic. It doesnt make sense to me for an implementation to do this, but is appears to be allowed. I think we should do the index check here and in the `uid` case

pkg/sypgp/sypgp.go
ikaneshiro: I don't see anything in the spec allowing for characters other than `r`, `d`, and `e`. So i think we are ok here

pkg/sypgp/sypgp.go
DrDaveD: It doesn't make sense for some fields to ever be blank, including this one.  I didn't put checks on fingerprint, keybits, or the uid name either.  How paranoid do we have to be?  There's not a security risk.

PaulCharlton: @DrDaveD @ikaneshiro given the spec is ambiguous, better to be safe and defensive rather than sorry -- not to argue `It doesn't make sense for some fields to ever be blank` ... making sense is not part of the spec.   ambiguous is part of the spec.  there's not much extra code involved to be safe.

DrDaveD: Ok, I'm out of time on this for at least the next couple of weeks and am not motivated to continue it, so feel free to replace or build on this PR as you wish.

PaulCharlton:
can be closed in favor of #113

PR #107, opened 2022-01-01
Fixes #59  - update example go modules and plugin go modules files

### LINKED ISSUE #59: update all example `go.mod` to refer to apptainer packages
update all example `go.mod` to refer to apptainer packages

prerequisite for v1.0.0

PaulCharlton:
Fixes #59  - update example go modules and plugin go modules files


internal/pkg/plugin/create.go
```diff
@@ -20,11 +20,11 @@ import (
 
 const goMod = `module %s
 
-go 1.13
+go 1.16
 
 require github.com/apptainer/apptainer v0.0.0
 
-replace github.com/apptainer/apptainer => ./apptainer_source
+replace github.com/apptainer/apptainer => ../../../../apptainer
```
cclerget, line 27: `apptainer_source` is a symlink pointing to the source tree and it is created automatically, since the plugin skeleton could be created at an arbitrary location, changing it as "relative to" would break `plugin create` command, this change must be reverted

cclerget (changes requested):

internal/pkg/plugin/create.go
PaulCharlton: @cclerget and what if we put symlinks into the repo itself ... they should be rewritten by `plugin create` ?

cclerget: Those are symlinks outside of the repo itself. By example `apptainer plugin create /tmp/mynewplugin github.com/something/mynewplugin` is creating a directory skeleton with a templated plugin, it will create the symlink in this skeleton directory to point the plugin to the same source tree used to compile the current apptainer binary, this is an internal mechanism used to compile plugin against the same source tree, and actually necessary as we are not versioning our public package, ideally the source code of apptainer should be provided by the binary itself with the embed package or something.

PaulCharlton: @cclerget new code submitted

cclerget (approved):

ikaneshiro:
@PaulCharlton Can you please rebase this on main now that https://github.com/apptainer/apptainer/pull/117 is merged so we don't need to play CI retry games to get this passing?

ikaneshiro (approved):

PR #110, opened 2022-01-01
Reduce changelog version to low number so debchange will run and set a valid number

DrDaveD:
I think this will fix the errors seen in the debian ci test that cause it to fail due to an invalid version number starting with a hexadecimal letter.

### DIFF
--- a/dist/debian/DEBIAN_PACKAGE.md
+++ b/dist/debian/DEBIAN_PACKAGE.md
@@ -97,16 +97,17 @@ clean copy of the debian directory before you switch to (checkout) an
 older version.
 
 Usually `debchange` is configured by the environment variables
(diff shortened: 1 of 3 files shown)


PaulCharlton:
@DrDaveD IDK -- may want to look at `${gitroot}/scripts/get-version`

DrDaveD:
This works because in [rules](https://github.com/apptainer/apptainer/blob/main/dist/debian/rules#L29) `NEW_VERSION` is only set when the value of `scripts/get_version` is newer than the version in `control`, and debchange is only called when `NEW_VERSION` is set.

PaulCharlton (approved):

ikaneshiro (approved):

PR #111, opened 2022-01-03
Loop device fix

### LINKED ISSUE #115: make loop device more reliable
need for apptainer 1.0.0

cclerget:

Sleep retry logic was restored to 5 retries and a sleep interval of 200ms.
Additionally after this period of 1s we proceed with a cache flush:
- `BLKFLSBUF` ioctl command is issued for image file located on a block device
- or by calling `sync` syscall

And after that if we keep seeing `EAGAIN` error, a grace period of one second is added before returning an error.

To reduce lock contention introduce by all those sleeping periods, the global lock on `/dev` was removed, instead an exclusive lock is associated to a single loop device only when shared loop devices is enabled.

- Fixes #115

### DIFF
--- a/pkg/util/loop/loop.go
+++ b/pkg/util/loop/loop.go
@@ -13,10 +13,12 @@ import (
 	"fmt"
 	"os"
 	"syscall"
+	"time"
 	"unsafe"
 
 	"github.com/apptainer/apptainer/pkg/sylog"
 	"github.com/apptainer/apptainer/pkg/util/fs/lock"
+	"github.com/apptainer/apptainer/pkg/util/fs/proc"
 	"golang.org/x/sys/unix"
 )
 
@@ -81,6 +83,16 @@ type Info64 struct {
 	Init           [2]uint64
 }
 
+// loop status retry related constants
+const (
+	sleepRetries      = 5
+	sleepInterval     = 200 * time.Millisecond
+	flushGracePeriod  = 1 * time.Second
+	errStatusTryAgain = syscall.EAGAIN
+)
+
+type retryStatusFn func(string, int) error
+
 // AttachFromFile attempts to find a suitable loop device to use for the specified image.
 // It runs through /dev/loopXX, up to MaxLoopDevices to find a free loop device, or
 // to share a loop device already associated to file (if shared loop devices are enabled).
@@ -88,8 +100,6 @@ type Info64 struct {
 // If a usable loop device is not found, and this is due to a transient EAGAIN / EBUSY error,
 // then it will retry up to maxRetries times, retryInterval apart, before returning an error.
 func (loop *Device) AttachFromFile(image *os.File, mode int, number *int) error {
-	var err error
-
 	if image == nil {
 		return fmt.Errorf("empty file pointer")
 	}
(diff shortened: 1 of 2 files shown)


PaulCharlton (approved):

pkg/util/loop/loop.go
```diff
@@ -122,29 +128,25 @@ func (loop *Device) AttachFromFile(image *os.File, mode int, number *int) error
...
 
 		// Try to open an existing loop device, but don't create a new one
-		loopFd, err := openLoopDev(device, mode, false)
+		loopFd, releaseLock, err := openLoopDev(device, mode, false, true)
 		if err != nil {
 			if !os.IsNotExist(err) {
-				sylog.Debugf("Couldn't open loop device %d: %v\n", device, err)
+				sylog.Debugf("Couldn't open loop device %d: %s\n", device, err)
 			}
 			continue
```
PaulCharlton, line 145: @cclerget are release lock and err guaranteed to be mutually exclusive?

pkg/util/loop/loop.go
```diff
@@ -154,8 +156,7 @@ func (loop *Device) shareLoop(imageIno, imageDev uint64, mode int, number *int)
 			// be sure that the loop device won't be released between this
 			// check and the mount of the filesystem
 			sylog.Debugf("Sharing loop device %d", device)
-			loop.fd = new(int)
-			*loop.fd = loopFd
+			loop.fd = &loopFd
```
PaulCharlton, line 159: @cclerget loop.fd is not thread-safe, and not protected here.

pkg/util/loop/loop.go
```diff
@@ -168,82 +169,48 @@ func (loop *Device) shareLoop(imageIno, imageDev uint64, mode int, number *int)
...
-		clearDevice := false
 
 		// Try to open the loop device, creating the device node if needed
-		loopFd, err := openLoopDev(device, mode, true)
+		loopFd, releaseLock, err := openLoopDev(device, mode, true, loop.Shared)
 		if err != nil {
-			sylog.Debugf("couldn't openLoopDev loop device %d: %v", device, err)
+			sylog.Debugf("Couldn't open loop device %d: %v", device, err)
 			return err
 		}
```
PaulCharlton, line 191: @cclerget what about `defer releaseDevice...`

pkg/util/loop/loop.go
```diff
@@ -253,39 +220,157 @@ func (loop *Device) attachLoop(image *os.File, mode int, number *int) error {
...
+			continue
+		}
+		// trigger block device flush command
+		f, err := os.Open(e.Source)
+		if err != nil {
+			return fmt.Errorf("while opening %s: %s", e.Source, err)
+		}
+		defer f.Close()
+
+		_, _, esys := syscall.Syscall(syscall.SYS_IOCTL, f.Fd(), unix.BLKFLSBUF, 0)
```
PaulCharlton, line 364: @cclerget something to consider ... bd_flush may not flush at the FS level, nor stop FS I/o in flight.   Noting that loop device is rare use-case where kernel internally actually captures an FD.    Seems like there should be some ioctl at the mount point which performs fs.flush properly

pkg/util/loop/loop.go
cclerget: releaseLock is valid only if the loop device was opened successfully

pkg/util/loop/loop.go
cclerget: `loop` is a short-live object not a singleton object and `AttachFromFile` is not intended to be called in parallel for the same instance

pkg/util/loop/loop.go
cclerget: stacking defer inside loop is not ideal here as it would lead to side effects like hitting the file descriptor limit

pkg/util/loop/loop.go
cclerget: I didn't find more granular than this call available from user space, this is called after the sleep period of 1s, so we are already in a worst case scenario, this is not 100% reliable and it's provided as best-effort, but as 5.8 kernel introduced `LOOP_CONFIGURE` ioctl allowing to set the file descriptor and the offset/sizelimit in a single ioctl command, this issue would disappear. I won't push other effort on fixing this one, I restored the old sleeping behavior and added some flushing helper, I'm done

.github/workflows/ci.yml
```diff
@@ -209,6 +209,13 @@ jobs:
           E2E_PARALLEL: 8
         run: make -C ./builddir e2e-test
 
+      - name: Loop device debug
```
ikaneshiro, line 212: Lets remove this before merging

ikaneshiro (approved):
lgtm

PR #131, opened 2022-01-12
Fixes #128 - no warning if ENV override with same value

### LINKED ISSUE #128: Allow both APPTAINER and SINGULARITY env variables to be set without warning
On the current main branch, after #108, a warning is printed whenever a SINGULARITY prefix variable is set, even if the corresponding APPTAINER prefix variable is also set to the same value.  This behavior would prohibit people from writing scripts that work with both apptainer and singularity without having to ignore or suppress warnings.  Change the behavior to only print a warning if the value of the two variables are different, and to print only a debug message if they are the same.

PaulCharlton:
Fixes #128 - no warning if ENV override with same value

if override with new value - emit warning
if override with same value - emit debug message

### DIFF
--- a/cmd/internal/cli/apptainer.go
+++ b/cmd/internal/cli/apptainer.go
@@ -482,8 +482,9 @@ func Init(loadPlugins bool) {
 	apptainerCmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
 		persistentPreRun(cmd, args)
 		var err error
-		for prefixIndex := range env.ApptainerPrefixes {
-			err = cmdManager.UpdateCmdFlagFromEnv(cmd, prefixIndex)
+		foundKeys := make(map[string]string)
+		for precedence := range env.ApptainerPrefixes {
+			err = cmdManager.UpdateCmdFlagFromEnv(cmd, precedence, foundKeys)
 			if nil != err {
 				break
 			}
--- a/internal/pkg/util/env/create.go
+++ b/internal/pkg/util/env/create.go
@@ -44,10 +44,17 @@ var alwaysOmitKeys = map[string]bool{
 type envKeyMap = map[string]string
 
 // setKeyIfNotAlreadyOverridden sets a value for key if not already overridden
-func setKeyIfNotAlreadyOverridden(g *generate.Generator, envKeys envKeyMap, prefixedKey, key, value string) {
+func setKeyIfNotAlreadyOverridden(g *generate.Generator, envKeys envKeyMap, precedence int, prefixedKey, key, value string) {
 	if oldValue, ok := envKeys[key]; ok {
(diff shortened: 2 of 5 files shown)


DrDaveD (changes requested):
The warning that I see when setting both APPTAINER_TMPDIR and SINGULARITY_TMPDIR to the same value is
```
WARNING: Environment Prefix [SINGULARITY_] has been deprecated in favor of [APPTAINER_] for [SINGULARITY_TMPDIR]
```
I don't see you changing that message here.

PaulCharlton:
@DrDaveD we need to be explicit as to the difference between `APPTAINER_` and `APPTAINERENV_` ... they are two entirely different things

PaulCharlton:
this PR addresses the "APPTAINERENV_" only

PaulCharlton:
added code for `APPTAINER_` prefix
1) warning if deprecated prefix used with different value or standalone
2) debug message if deprecated prefix used with identical value

@DrDaveD

DrDaveD (changes requested):
I compiled the current PR branch and found some problems. 

1. I still see the deprecated warning with SINGULARITY_ & APPTAINER_ prefixes setting the same value:
```
$ SINGULARITY_TMPDIR=/scratch/tmp APPTAINER_TMPDIR=/scratch/tmp apptainer exec docker://alpine true
WARNING: Environment Prefix [SINGULARITY_] has been deprecated in favor of [APPTAINER_] for [SINGULARITY_TMPDIR], using deprecated value [/scratch/tmp]
```

2. With a SINGULARITYENV_ in the deprecated message it should not append the key to prefixedKey, it gets a duplicate of the name:
```
$ SINGULARITYENV_FOO=bar apptainer exec docker://alpine true
WARNING: DEPRECATED USAGE: forwarding SINGULARITYENV_FOOFOO as environment variable will not be supported, use APPTAINERENV_FOO instead
```
+ I also noticed that the message looks very different from the message used with SINGULARITY_ variables as above.  While you're fixing this please make it look more like the other one since I think users will expect consistent behavior for environment variables even if they are processed differently.

3. There is a new warning when an ENV_ variable is overriding something from the environment, e.g.:
```
$ APPTAINERENV_FOO=bar FOO=baz apptainer exec docker://alpine true
WARNING: Environment variable [FOO] already has value [bar], will not forward new value [baz] from parent process environment
```
+ This didn't happen before.  I'm not saying that this alone is necessarily a problem, but it even happens with the exec `-e` option which says to not import anything from the parent environment, and I think that at least is a problem.

4. There should probably be some automated tests added for this functionality.

PaulCharlton:
@DrDaveD all should be ready for review except your point 4 moved to #132

DrDaveD (approved):
Ok, this is good enough.  I have some more suggested wording changes to the messages but I'll do it as a separate PR.

PR #133, opened 2022-01-12
Make environment messages more consistent

DrDaveD:
This makes the wording of the messages from #131 regarding SINGULARITY/APPTAINER environment importing more consistent.

### DIFF
--- a/pkg/cmdline/flag.go
+++ b/pkg/cmdline/flag.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Apptainer a Series of LF Projects LLC
+// Copyright (c) 2021-2022 Apptainer a Series of LF Projects LLC
 //   For website terms of use, trademark policy, privacy policy and other
 //   project policies see https://lfprojects.org/policies
 // Copyright (c) 2019-2021, Sylabs Inc. All rights reserved.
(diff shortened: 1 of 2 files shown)


pkg/cmdline/flag.go
```diff
@@ -198,16 +198,15 @@ func (m *flagManager) updateCmdFlagFromEnv(cmd *cobra.Command, precedence int, f
 				continue
 			}
 			if precedence > 0 {
-				deprecationFormat := "DEPRECATED USAGE: Environment Prefix [%s] for environment variable will not be supported in the future, use [%s] instead for [%s],"
```
PaulCharlton, line 201: @DrDaveD I have to disagree with this change.  all 3 of the messages below are "deprecated used" due to the fact that the prefix IS NOT `APPTAINER_` ...

PaulCharlton (changes requested):

PaulCharlton:
@DrDaveD please change copyright at top to have "2021-2022"

pkg/cmdline/flag.go
DrDaveD: That's not true.  It's not deprecated to set both APPTAINER_ and SINGULARITY_ copies of the same variable; the deprecated case is when only SINGULARITY_ is set.  This set of messages is consistent with the messages on the *ENV_ variables.

ikaneshiro: I consider the main concerns wrt environment handling is to be clear when an override is occurring and inform users when their cli behavior is determined be deprecated variables. Both cases appear satisfied in this PR.

ikaneshiro (approved):

PR #160, opened 2022-01-14
Fixes #144 - sequential numeric processing of batch files across different locale/fsys/directory structures

### LINKED ISSUE #144: force sequential numeric processing of init scripts regardless of collation sequence of dir/filesystem/locale
force sequential numeric processing of init scripts regardless of collation sequence of dir/filesystem/locale

globbing such as
```
for script in /.singularity.d/env/*.sh; do
done
```

does not provide a consistent guarantee across different *NIX-like systems, filesystems, dirs as to processing in numerical order

use something like this instead:

```
for script in $(ls /.singularity.d/env/*.sh | sort -n); do
done
```

PaulCharlton:
Fixes #144 - sequential numeric processing of batch files across different locale/fsys/directory structures

### DIFF
--- a/internal/pkg/build/sources/exec.sh
+++ b/internal/pkg/build/sources/exec.sh
@@ -7,7 +7,7 @@
 # LICENSE.md file distributed with the sources of this project regarding your
 # rights to use or distribute this software.
 
-for script in /.singularity.d/env/*.sh; do
+for script in $(ls /.singularity.d/env/*.sh | sort -n); do
     if [ -f "$script" ]; then
         . "$script"
     fi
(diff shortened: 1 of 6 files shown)


cclerget (changes requested):
Blocking this PR, because it introduces silent breaking changes, it was noted that it passes `apptainer run docker://hello-world` e2e test while it shouldn't because of missing `ls` and `sort` binaries, like discussed in private, we can't rely on binaries present in the container images.

Now the embedded shell interpreter use golang `ioutil.ReadDir` https://github.com/mvdan/sh/blob/b7f83dc122bab372d3c3945f5c4c30227e6af3a5/interp/runner.go#L147 , which is returning entries sorted by filename, it is not a numerical sorting but as we are using a sysV like naming by starting script name with two digits, it works just fine.

So these changes are not a required.

PaulCharlton:
issue closed - see comment in #144

PR #188, opened 2022-01-28
Fix mount ordering between image and user binds

### LINKED ISSUE #187: Honor bind mount ordering when mixin directory/files and image bind mounts
Actually image bind mount are always done before user specified binds, so user binds always take precedence over image bind mount. Fix the mount order for consistency with user binds so that:

```
singularity exec -B /scratch,img.sqfs:/scratch/bin:image-src=/ container.sif test -f /scratch/bin/testing
```

mount the image `/scratch/bin` over `/scratch` while:

```
singularity exec -B img.sqfs:/scratch/bin:image-src=/,/scratch container.sif test -f /scratch/bin/testing
```

mount `/scratch` over `/scratch/bin`

cclerget:
- Fixes #187

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 - The `SINGULARITY_LABELS` environment variable within build definitions has
   been restored.
+- Fix mount ordering between image bind and user binds.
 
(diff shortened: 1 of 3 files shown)


DrDaveD (changes requested):
Looks good, but please add an entry in CHANGELOG.md under changes since last release.

DrDaveD (approved):

PR #191, opened 2022-02-01
Fixes #137 - add shellcheck to CI

### LINKED ISSUE #137: add CI step for shellcheck
add CI step for shellcheck

PaulCharlton:
Fixes #137 - add shellcheck to CI

At the moment, errors are ignored until we burn down the backlog of reported issues


.github/workflows/ci.yml
```diff
@@ -44,6 +44,16 @@ jobs:
...
 
+  shellcheck:
+    name: shellcheck
+    runs-on: ubuntu-20.04
+    container: koalaman/shellcheck-alpine
+    steps:
+      - uses: actions/checkout@v2
+      - name: shellcheck files that appear to be sh or bash scripts (or some cousin thereof)
+        run: |
+          shellcheck $(grep -rm1 ^ * | grep -v "^vendor|^builddir" | grep '#!/bin/*.sh' | sed -e 's|:#!.*||') || true
```
DrDaveD, line 55: I think that's supposed to be `'#!/bin/.*sh'`.  As it is, it is only matching `#!/bin/sh` and not any other shells.  The only reason it is matching that is because the `/*` is treated as zero characters and the `.` matches the `/`.

cclerget:
@PaulCharlton We could exclude `mconfig`, `mlocal` and `makeit` for a first pass to drastically reduce the number of warnings

.github/workflows/ci.yml
PaulCharlton: I should have marked it WIP -- caught that one and replaced the complex command with a shell script

PaulCharlton:
@cclerget re:  `exclude mconfig, mlocal and makeit`. I'd rather not hide any errors for now ... they are ALL suppressed at the moment.

DrDaveD:
Frankly, I don't see a point to adding a check that is completely ignored.  I would rather defer this until something is actually being checked.

PaulCharlton:
@DrDaveD `completely ignored` is a point of view.  Although a pedantic view that the automation CI does nothing with the result is _currently_ true, the result is not being ignored by developers who can start whittling away at the 700+ issues reported by the shellcheck lint -- some of them are actually serious impediments, not just style issues.    Let's start filing issues that reference the shellcheck output ... which further removes this from the arguable `ignored` status.

DrDaveD:
Developers can run shellcheck on their own so I don't see how it helps having it in a CI check.  CI is for automated checks.

DrDaveD (approved):

PR #208, opened 2022-02-03
Handle remaining legacy environments variables

### LINKED ISSUE #206: Remove remaining direct reading of APPTAINER_ variables from the environment
[This little bit of code](https://github.com/apptainer/apptainer/blob/main/cmd/internal/cli/actions_linux.go#L93) is just about the only place directly reading "APPTAINER_" prefixed environment variables, and the comment says it is for compatibility with v2.  If it really is for that purpose, it seems to me that it should be using "SINGULARITY_" prefixes.  Is it even needed at all anymore?

cclerget:
Have been removed:
- APPTAINER_LOCALCACHEDIR (legacy v2 variable)
- APPTAINER_CACHEDIR, after all this the cache directory not for image extraction (and this is confusing)

Also fix remaining legacy environments variables:
- for cache config
- for library image download

Fixes #206

### DIFF
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -24,6 +24,7 @@ import (
 	"github.com/apptainer/apptainer/internal/pkg/client/oci"
 	"github.com/apptainer/apptainer/internal/pkg/client/oras"
 	"github.com/apptainer/apptainer/internal/pkg/client/shub"
+	"github.com/apptainer/apptainer/internal/pkg/util/env"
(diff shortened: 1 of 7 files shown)


ikaneshiro (approved):

DrDaveD (changes requested):
Rather than reading APPTAINER_TMPDIR directly, it would be better to use the actionTmpDirFlag already parsed.

DrDaveD (approved):

PR #210, opened 2022-02-03
Remove subshell overhead in environment processing, from sylabs 545

DrDaveD:
This pulls from sylabs
- sylabs/singularity#545
which addressed
- sylabs/singularity#544

The original PR description was
> The action script environment processing was calling out to `getenvkey` and `unescape` go functions via subshell. This is expensive, especially with large environments, as the environment etc. must be duplicated for each subshell and other subshell setup performed.
> 
> Replace these two go functions with shell var substitution syntax.

### DIFF
--- a/internal/pkg/util/fs/files/action_script.sh
+++ b/internal/pkg/util/fs/files/action_script.sh
@@ -1,9 +1,9 @@
-#!/bin/sh
+#!/bin/bash
 # Copyright (c) Contributors to the Apptainer project, established as
 #   Apptainer a Series of LF Projects LLC.
 #   For website terms of use, trademark policy, privacy policy and other
 #   project policies see https://lfprojects.org/policies
-# Copyright (c) 2020-2021, Sylabs Inc. All rights reserved.
+# Copyright (c) 2020-2022, Sylabs Inc. All rights reserved.
 # This software is licensed under a 3-clause BSD license. Please consult the
 # LICENSE.md file distributed with the sources of this project regarding your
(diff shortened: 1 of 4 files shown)


internal/pkg/util/fs/files/action_script.sh
```diff
@@ -68,9 +68,9 @@ restore_env() {
     # defined by docker or virtual file above, empty
     # variables are also unset
     for e in ${__exported_env__}; do
-        key=$(getenvkey "${e}")
+        key=${e%%=*}
         if ! test -v "${key}"; then
-            export "$(unescape ${e})"
+            export "${e%%=*}"
```
cclerget, line 73: should be `export "${e//'\n'/$IFS}"`

cclerget (changes requested):

internal/pkg/util/fs/files/action_script.sh
DrDaveD: Yes, thanks for pinpointing the problem.

cclerget (approved):

internal/pkg/util/fs/files/action_script.sh
```diff
@@ -1,9 +1,9 @@
-#!/bin/sh
+#!/bin/bash
```
PaulCharlton, line 1: @cclerget is this interpreted by the apptainer built-in shell interpreter?

cclerget: no this is comment for the interpreter

PR #250, opened 2022-02-12
Allow unprivileged users to build images by default

### LINKED ISSUE #215: Singularity should allow any user to build from a definition file.
Continuation of apptainer/singularity#5941.

ShamrockLee:

~This patch removes several `Geteuid() != 0` checks and allow unprivileged users to build images by default.~
This patch introduces a new flag `-N|--unprivileged`, with which some `Geteuid() != 0` will be skipped.

### Motivations
*   Nowadays, package managers such as Nix and Gentoo ebuild and a lot of projects manags don't require root privileges to work, while still benefits from / rely on the customized directory trees (not available in Appimages). As a mature container format, Apptainer serves as a reliable way to wrap the build result, making it portable and runnabe on network-based file systems where I/O of small files is terrably slow.

*   As a practice of the *same privilege inside and outside* philosophy of Apptainer, it is the user's responsibility to provide enough level privilege for their toolchain. This patch also enable users to build in restricted environment where there is no way to get root privilege and that no suid-ed executables are available.

*   A program should be given the minimum level of privilege needed from the security perspective, which is known as the [Principle of Least Privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege). It makes the container building (especially the `%startup` and `%files` sections) much safer.

### Function parameter changes
*   runSectionScript(name string, script types.Script) -> runSectionScript(name string, script types.Script, unprivileged bool)

*   Full(ctx context.Context) -> Full(ctx context.Context, unprivileged bool)

*   Use b.Full(ctx, false) in ConvertOCIToSIF

### Status
This patch is compiled and run locally, and it work as expected so far, but is still a draft considering the need of discussion, consensus reaching and documentation update.

 - Fixes #215

### DIFF
--- a/cmd/internal/cli/build.go
+++ b/cmd/internal/cli/build.go
@@ -43,6 +43,7 @@ var buildArgs struct {
 	nvccli        bool
 	rocm          bool
 	writableTmpfs bool // For test section only
+	unprivileged  bool
 }
 
 // -s|--sandbox
@@ -224,6 +225,16 @@ var buildWritableTmpfsFlag = cmdline.Flag{
 	EnvKeys:      []string{"WRITABLE_TMPFS"},
 }
 
+// -N|--unprivileged
+var buildUnprivilegedFlag = cmdline.Flag{
+	ID:           "buildUnprivilegedFlag",
+	Value:        &buildArgs.unprivileged,
+	DefaultValue: false,
+	Name:         "unprivileged",
+	ShortHand:    "N",
+	Usage:        "build without root-privilege examination. Conventional package managers (e.g. apt and rpm) and system tools may fail without root privileges.",
+}
+
 func init() {
 	addCmdInit(func(cmdManager *cmdline.CommandManager) {
 		cmdManager.RegisterCmd(buildCmd)
@@ -256,6 +267,7 @@ func init() {
 		cmdManager.RegisterFlagForCmd(&buildBindFlag, buildCmd)
 		cmdManager.RegisterFlagForCmd(&buildMountFlag, buildCmd)
 		cmdManager.RegisterFlagForCmd(&buildWritableTmpfsFlag, buildCmd)
+		cmdManager.RegisterFlagForCmd(&buildUnprivilegedFlag, buildCmd)
 	})
 }
 
--- a/cmd/internal/cli/build_linux.go
+++ b/cmd/internal/cli/build_linux.go
@@ -161,8 +161,8 @@ func runBuildLocal(ctx context.Context, cmd *cobra.Command, dst, spec string) {
 		sylog.Fatalf("Failed to create an image cache handle")
 	}
 
-	if syscall.Getuid() != 0 && !buildArgs.fakeroot && fs.IsFile(spec) && !isImage(spec) {
-		sylog.Fatalf("You must be the root user, however you can --fakeroot to build from an Apptainer recipe file")
+	if syscall.Getuid() != 0 && !buildArgs.fakeroot && !buildArgs.unprivileged && fs.IsFile(spec) && !isImage(spec) {
+		sylog.Fatalf("You must be the root user, however you can --fakeroot or --unprivileged to build from an Apptainer recipe file")
 	}
 
 	err := checkSections()
@@ -263,7 +263,7 @@ func runBuildLocal(ctx context.Context, cmd *cobra.Command, dst, spec string) {
 		sylog.Fatalf("Unable to create build: %v", err)
 	}
 
-	if err = b.Full(ctx); err != nil {
+	if err = b.Full(ctx, buildArgs.unprivileged); err != nil {
 		sylog.Fatalf("While performing build: %v", err)
 	}
 }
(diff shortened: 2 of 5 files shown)


DrDaveD:
Failed lint check because of an unused variable and failed the e2e test that was checking for an error when running as an unprivileged user.

ShamrockLee:
> Failed lint check because of an unused variable

The dead-code part is caused by a function
```Go
func isImage(spec string) bool
```
that is unrelated to these changes.

Is it okay to remove?

> the e2e test that was checking for an error when running as an unprivileged user

Does that mean the check expects the privilege check to be conducted and will fail otherwise?

DrDaveD:
> > Failed lint check because of an unused variable
> 
> The dead-code part is caused by a function
> 
> ```go
> func isImage(spec string) bool
> ```
> 
> that is unrelated to these changes.

It's not unrelated, because you removed the only call to it.

> Is it okay to remove?

It would be OK if we're not going to use it anymore, but I think we're going to want these checks you have disabled to still happen by default, and only remove them if someone sets some kind of option.

> > the e2e test that was checking for an error when running as an unprivileged user
> 
> Does that mean the check expects the privilege check to be conducted and will fail otherwise?

Correct.  I expect we will want the e2e check to stay there for the default case and we'll want another one to check the optional case.

cclerget:
Hi @ShamrockLee , unfortunately it won't be as simple as removing `Geteuid() != 0`, the issue there is that the container image build is occurring in the same context as `apptainer exec` since it call it behind the scene, so as an unprivileged user you are not authorized to specify an alternate configuration to fully control what's being mounted inside the container during a build (and build is relying on this to override the config specified by an administrator in the build context), explaining why you removed the configuration check here https://github.com/apptainer/apptainer/pull/250/files#diff-15947087c721a58de48322e21fffa6a6d81e73df717c4749a39522198e27b56eL445 to force build to run correctly, but the side effect of this:
- `/etc/passwd` and `/etc/group` are mounted in the container during build
- your home directory is mounted in the container during build
- the build is subject to admin config restrictions like `user bind control = no`, `allow container dir = no`, `limit container paths`
So build may randomly failed based on the configuration set.

So far home mount could be easily fixed by using `--no-home` flag or `--no-mount=home`, `/etc/passwd` and `/etc/group` are requiring more changes to disable mount as unprivileged user, thinking about using `--no-mount` flag for that.
But for config restrictions that could be potentially set, this is an obvious blocker for security reasons, so it could be only handled through checks and aborting gracefully by informing users that it won't be possible to run build command due to restrictions set by the configuration file.

To summarize:
- I would use a flag like `--unprivileged` for build command instead of making it transparent because of all implications behind, so an end user would know what he's doing by using this flag.
- Unprivileged build requires additions regarding `--no-mount` flag by introducing `passwd`, `group` and also `resolv.conf` so unprivileged users could control that
- Return appropriate errors for unprivileged build when restrictions are set by the configuration
- e2e tests for unprivileged build

ShamrockLee:
@cclerget Considering that, I'll drop the commit that loosen the configuration privilege check.

As for the choice of flag-names, I wonder if  `--unprivileged` would imply that root privileges are dropped inside the container. Or should we check `Getuid == 0` when the flag is specified?

cclerget:
> @cclerget Considering that, I'll drop the commit that loosen the configuration privilege check.
> 
> As for the choice of flag-names, I wonder if `--unprivileged` would imply that root privileges are dropped inside the container. Or should we check `Getuid == 0` when the flag is specified?

Yes they would be dropped automatically like it is done when running `apptainer exec` unprivileged, same thing. Basically adding `--unprivileged` to build command is sufficient and should be relatively similar to fakeroot checks

ShamrockLee:
What is the recommended Go implementation to *drop* root privileges? (This is the first time for me to work on a Go project, and my knowledge toward this language is very limited.)

This SO answer implement the permission down-grading by switching to  an unprivileged user.
https://stackoverflow.com/a/48991783

cclerget:
> What is the recommended Go implementation to _drop_ root privileges? (This is the first time for me to work on a Go project, and my knowledge toward this language is very limited.)
> 
> This SO answer implement the permission down-grading by switching to an unprivileged user. https://stackoverflow.com/a/48991783

You don't need to deal with privileges at all, you are already unprivileged in the apptainer CLI, build will call `apptainer exec` for `%post` and `%test` sections

ShamrockLee:
> * Unprivileged build requires additions regarding `--no-mount` flag by introducing `passwd`, `group` and also `resolv.conf` so unprivileged users could control that

I'm currently encountering this. When Apptainer images are to be built inside the Nix expression, top-level directories such as /etc are not available.

It would be convenient to have `--no-mount-all` also.

ShamrockLee:
Is there any clue where to work on the `--no-mount` implementation?

DrDaveD:
@ShamrockLee Just search for `no-mount` in the source code and follow through on its connections.  I see that it's value is stored in a `NoMount` variable.

ShamrockLee:
Closing. See #475 #481.

PR #278, opened 2022-02-22
Only release on latest tag

### LINKED ISSUE #268: Avoid accidental production of release assets
Currently anytime somebody pushes a tag to their own apptainer or singularity repository, release assets are generated which take a lot of github action computing time.  I'm worried about what would happen if somebody did `git push origin --tags` and pushed a whole lot of them at once.

One thing that might help is to make sure that scripts/get-version matches the GITHUB_REF, since the former only looks for annotated tags.  I'm not sure it would help for those tags that were already annotated with their official release, I'll have to try it.

DrDaveD:
This adds a check to the release_assets to make sure that the value of `scripts/get-version` (which is based on the latest annotated tag) matches the given tag.  It also separates out the release_assets to a separate yml file, so tagging does not re-run all of the CI tests.  Together these changes avoids a lot of side effects if somebody does `git push origin --tags`.

- Fixes #268


.github/workflows/release.yml
```diff
@@ -0,0 +1,77 @@
+name: release
+on:
+  push:
+    branches:
+      - main
+      - 'release-*'
```
ikaneshiro, line 6: Do we need to run this on pushes to these branches if the check will only happen when a tag is pushed to the repo?

DrDaveD: I thought I read the [documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore) to say that if both were defined, both filters would have to be satisfied.  But now I see at the end of the paragraph where they're talking about the various combinations of tags & branches it all of a sudden mentions both branches and *paths* instead of branches and *tags*.  I think that must be a copy/paste error.

Having said that, I have been able to test this by just pushing a tag to my own branch, so it doesn't appear to me that the branches piece is actually doing anything.  I knew that but thought it wouldn't hurt to leave them in.  I'll take out the branches.

ikaneshiro (approved):

PR #289, opened 2022-02-28
Require binutils-gold when build is targeting SUSE

mstg:

With the current spec, builds targeting SUSE will fail with `collect2: fatal error: cannot find 'ld'` even after installing build dependencies (for example with `dnf builddep`). Adding `binutils-gold` as a build dependency fixes this issue.

This PR simply adds `binutils-gold` as a build dependency if the target vendor is SUSE.

### DIFF
--- a/dist/rpm/apptainer.spec.in
+++ b/dist/rpm/apptainer.spec.in
@@ -57,12 +57,9 @@ Provides: singularity-runtime = 1:%{version}-%{release}
 Obsoletes: singularity-runtime < 3.0
 
 %if "%{_target_vendor}" == "suse"
-%if "%{sles_version}" != "11"
-BuildRequires: go
+BuildRequires: binutils-gold
 %endif
-%else
 BuildRequires: golang
-%endif
 BuildRequires: git
 BuildRequires: gcc
 BuildRequires: make


DrDaveD:
I don't use SUSE so I'll have to trust you on this, but are you confident that it will work on all supported versions of SUSE?

mstg:
@DrDaveD The change was tested on openSUSE Leap 15.0 to 15.4. (Using the containers in https://hub.docker.com/r/opensuse/leap). `binutils-gold` seems to be available since 12 so that should cover older and newer versions. Is there a list of versions you want me to test, or does this sound good?

DrDaveD:
I'm not familiar with what versions are supported.  If that covers all supported versions, OK.  I see `%if "%{sles_version}" != "11"` in there, but googling for "sles 11 end of life" says it was March 31, 2019 so I guess it's not a thing anymore.  If you agree with that, why don't you take that `%if` out at the same time.

mstg:
Both `go` and `golang` installs the same package on Leap 15 and onwards so not having any conditionals for SUSE makes sense. I've now pushed the change that removes that `%if` statement.

DrDaveD (approved):

PR #315, opened 2022-03-11
feat: Add riscv64 support, from sylabs 631

edytuk:
This pulls in sylabs PR

- sylabs/singularity#631
which fixed
- sylabs/singularity#628

> SIF now supports riscv64. Add ELF information to avoid warning when container is examined.


DrDaveD:
Hold this until dependabot asks to include sif v2.4.0.

DrDaveD (approved):
Sif was updated in #319

PR #335, opened 2022-03-17
feat: enable native v2 unified cgroups config, from sylabs 539

edytuk:
This pulls in sylabs PR

- sylabs/singularity#539
which fixed
- sylabs/singularity#538

> Allow the `unified` key to be used in a cgroups config toml file, to directly apply resource limits using the v2 unified hierarchy, rather than v1 -> v2 translation.


DrDaveD (approved):
@cclerget FYI

PR #354, opened 2022-03-25
e2e: non-pid namespace e2e-tests for OCI/CGROUPS, from sylabs 622

edytuk:
This pulls in sylabs PR

- sylabs/singularity#622
which fixed
- sylabs/singularity#563

> The Singularity e2e-tests were previously all run in a mount and PID
> namespace, to avoid polluting the host filesystem (critical), and
> process tree (less critical). These namespaces are set up in some CGO
> init code.
> 
> The use of the PID namespace prevents testing with systemd as the
> cgroups manager, as systemd is aware of the host process tree, not the
> one in the new e2e PID namespace. This is a major omission since sylabs/singularity#540
> switched to using systemd for cgroups management by default.
> 
> This PR:
> 
> * Modifies the e2e init CGO code, so that an env var
>       `SINGULARITYE_E2E_NO_PID_NS` will cause it _not_ to create a new PID
>       namespace.
> 
> * Modifies the Makefile so that the e2e suite is called twice, once
>       with PID ns, once without.
> 
> * Moves the INSTANCE cgroups test into a new package/topic
>       e2e/cgroups, run in the e2e call without PID ns.
> 
> * Moves the OCI tests into the e2e call without PID ns.
> 
> * Modifies the CGROUPS and OCI tests so that they test with both
>       systemd and cgroupfs management, using a convenience wrapper function.


DrDaveD:
Closing & reopening to kick CI

cclerget:
This is a quicky'n dirty way to run those tests, we could use `PR_SET_CHILD_SUBREAPER` to replace PID namespace

DrDaveD:
@cclerget Could you please provide a little more detail or an example of how to use `PR_SET_CHILD_SUBREAPER`?

cclerget:
no permission to push on your branch, closing and superseded by #368

PR #355, opened 2022-03-25
feat: --no-mount for 'bind path' entries, from sylabs 626

edytuk:
This pulls in sylabs PR

- sylabs/singularity#626
which fixed
- sylabs/singularity#625

> Allow one or more `bind path` entries in `singularity.conf` to be
> ignored by specifying the absolute path of the destination as a
> `--no-mount` or `SINGULARITY_NO_MOUNT` value.
> 
> This allows finer-grained control of bind mounts than `--contain`,
> which will disable all.
> 
> It is useful in situations where e.g. a directory `/project` in a
> container clashes with a `bind path = /project` entry, but other
> default binds from the host are still required.
> 
> Note that `/etc/hosts` and `/etc/localtime` were always specially
> injected with `--contain`, but the new `--no-mount` handling now
> allows this to be overridden also.

### DIFF
--- a/pkg/runtime/engine/apptainer/config/config.go
+++ b/pkg/runtime/engine/apptainer/config/config.go
@@ -2,7 +2,7 @@
 //   Apptainer a Series of LF Projects LLC.
 //   For website terms of use, trademark policy, privacy policy and other
 //   project policies see https://lfprojects.org/policies
-// Copyright (c) 2019-2021, Sylabs Inc. All rights reserved.
+// Copyright (c) 2019-2022, Sylabs Inc. All rights reserved.
 // This software is licensed under a 3-clause BSD license. Please consult the
 // LICENSE.md file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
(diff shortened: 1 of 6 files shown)


DrDaveD:
Closing & reopening to kick off CI.

pkg/runtime/engine/apptainer/config/config.go
```diff
@@ -498,6 +499,16 @@ func (e *EngineConfig) GetNoCwd() bool {
 	return e.JSON.NoCwd
 }
 
+// SetSkipBinds sets bind paths to skip
+func (e *EngineConfig) SetSkipBinds(val []string) {
+	e.JSON.SkipBinds = val
+}
+
+// GetSkipBinds gets bind paths to skip
+func (e *EngineConfig) GetNoBinds() []string {
```
cclerget, line 508: Please rename to `GetSkipBinds` for consistency as it is also set in function comment

cclerget (changes requested):

pkg/runtime/engine/apptainer/config/config.go
edytuk: I renamed function and parameter for consistency. Is it ok?

cclerget: :+1: thanks @edytuk !

cclerget (approved):

PR #367, opened 2022-03-31
Use "binary path" config instead of separate configs for many executables

DrDaveD:
This introduces a new "binary path" configuration variable that replaces all the individual "path" variables in apptainer.conf.  If the path contains "$PATH:", then when running unprivileged it is replaced by the user's path and when running privileged it is removed.  

This was initially discussed in apptainer/singularity#5859 and suggested by @gmkurtzer in a [comment](https://github.com/apptainer/singularity/pull/5859#issuecomment-856928588).  In undertook it now in preparation for another PR that I have made that would have required adding yet another path in the configuration file without this.


cmd/starter/main_linux.go
```diff
@@ -50,7 +54,7 @@ func startup() {
 
 	// get engine operations previously registered
 	// by the above import
-	e := getEngine(jsonConfig)
+	e := getEngine(jsonConfig, sconfig)
```
cclerget, line 57:
```suggestion
        privStageOne := C.goexecute == C.STAGE1 && sconfig.GetIsSUID()
	e := getEngine(jsonConfig, privStageOne)
```

cmd/starter/main_linux.go
```diff
@@ -31,8 +31,12 @@ import (
 	_ "github.com/apptainer/apptainer/cmd/starter/engines"
 )
 
-func getEngine(jsonConfig []byte) *engine.Engine {
-	e, err := engine.Get(jsonConfig)
+func getEngine(jsonConfig []byte, sconfig *starterConfig.Config) *engine.Engine {
```
cclerget, line 34:
```suggestion
func getEngine(jsonConfig []byte, privStageOne bool) *engine.Engine {
```

internal/pkg/util/bin/bin.go
```diff
@@ -33,28 +33,42 @@ func FindBin(name string) (path string, err error) {
...
 
-// findOnPath performs a simple search on PATH for the named executable, returning its full path.
-// env.DefaultPath` is appended to PATH to ensure standard locations are searched. This
-// is necessary as some distributions don't include sbin on user PATH etc.
+// findOnPath performs a search on the configurated binary path for the
+// named executable, returning its full path.
 func findOnPath(name string) (path string, err error) {
+	cfg := apptainerconf.GetCurrentConfig()
+	if cfg == nil {
+		if strings.HasSuffix(os.Args[0], ".test") {
```
cclerget, line 50: Instead of having this directly in the code, I would put the parsing and initialization of current config directly in `_test.go` files

internal/pkg/util/bin/bin.go
```diff
@@ -68,9 +82,15 @@ func findOnPath(name string) (path string, err error) {
 func findFromConfigOrPath(name string) (path string, err error) {
 	cfg := apptainerconf.GetCurrentConfig()
 	if cfg == nil {
-		cfg, err = apptainerconf.Parse(buildcfg.APPTAINER_CONF_FILE)
-		if err != nil {
-			return "", errors.Wrap(err, "unable to parse apptainer configuration file")
+		if strings.HasSuffix(os.Args[0], ".test") {
```
cclerget, line 85: Instead of having this directly in the code, I would put the parsing and initialization of current config directly in `_test.go` files

internal/pkg/client/library/library.go
```diff
@@ -74,11 +74,15 @@ func getDownloadConfig() (libClient.Downloader, error) {
 	// get downloader parameters from config
 	conf := apptainerconf.GetCurrentConfig()
 	if conf == nil {
-		// sylog.Fatalf("Unable to get apptainer configuration")
-		var err error
-		conf, err = apptainerconf.Parse(buildcfg.APPTAINER_CONF_FILE)
-		if err != nil {
-			sylog.Fatalf("unable to parse apptainer.conf file: %s", err)
+		if strings.HasSuffix(os.Args[0], ".test") {
```
cclerget, line 77: Instead of having this directly in the code, I would put the parsing and initialization of current config directly in `_test.go` files

internal/pkg/util/bin/bin.go
DrDaveD: The problem with doing that here is that this function is used from other unit tests as well via the FindBin function, and it would have to be included in multiple `_test.go` files if it wasn't here.  So I think the loading config files for .test in this file need to stay here.

internal/pkg/client/library/library.go
DrDaveD: Likewise I found that this one was also called from other `_test.go` files in addition to its companion, so this also needs to stay.  The previous code was already loading it here; my change just makes sure that it is only applied during unit tests.

cclerget (approved):

PR #381, opened 2022-04-12
fix: check valid path for --pwd

kadern0:
Signed-off-by: Pablo Caderno <kaderno@gmail.com>

Changed default behavior of switching to a different directory if os.Chdir(e.EngineConfig.OciConfig.Process.Cwd) fails when starting the container.

Whilst this approach should fix the issue, it might be "too strict" for other cases.

 - Fixes https://github.com/apptainer/singularity/issues/6086

### DIFF
--- a/internal/pkg/runtime/engine/apptainer/process_linux.go
+++ b/internal/pkg/runtime/engine/apptainer/process_linux.go
@@ -77,8 +77,14 @@ func (e *EngineOperations) StartProcess(masterConnFd int) error {
 	bootInstance := isInstance && e.EngineConfig.GetBootInstance()
 	shimProcess := false
 
+	_, customCwd := e.EngineConfig.OciConfig.Annotations["CustomCwd"]
+
 	if err := os.Chdir(e.EngineConfig.OciConfig.Process.Cwd); err != nil {
+		if customCwd {
+			return fmt.Errorf("failed to set working directory: %s", err)
+		}
 		if err := os.Chdir(e.EngineConfig.GetHomeDest()); err != nil {
+			sylog.Debugf("Error setting the working directory. Using '/' instead: %s", err)
 			os.Chdir("/")
 		}
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 ### Changed defaults / behaviours
 
+- When starting a container, if the user has specified the cwd by using
+  the `--pwd` flag, in case of problem return error instead of defaulting to
+  a different directory.
 - Added a squashfuse image driver that enables mounting SIF files as an
   unprivileged user.  Requires the separate installation of squashfuse.
 - Added the ability to use persistent overlay (`--overlay`) and
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -89,4 +89,5 @@
 - Pranathi Locula <locula@deshaw.com>
 - Pedro Alves Batista <pedro.pesquisapb@gmail.com>
 - Chen Yiyang <cyyzero@qq.com>
+- Pablo Caderno <kaderno@gmail.com>
 ```
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -642,6 +642,10 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 		engineConfig.SetCwd(pwd)
 		if PwdPath != "" {
 			generator.SetProcessCwd(PwdPath)
+			if generator.Config.Annotations == nil {
+				generator.Config.Annotations = make(map[string]string)
+			}
+			generator.Config.Annotations["CustomCwd"] = "true"
 		} else {
 			if engineConfig.GetContain() {
 				generator.SetProcessCwd(engineConfig.GetHomeDest())


DrDaveD:
please rebase on current main and force-push

kadern0:
> please rebase on current main and force-push

Done!

DrDaveD:
We'll need input from @cclerget or @gmkurtzer about this.  Personally I'm uncomfortable with changing a default that some people may be depending on.  A warning seems more in order to me than an error.

DrDaveD:
Oh, does this apply only when a user passes a `--pwd` option?  If that was the case that would make me more comfortable with it.  However, my guess is that this also applies if the default directory does not exist in the container.  The default directory is whatever the directory is outside the container.  For example if somebody runs with `--contain` and they start from somewhere under their home directory it's likely to fail, because in that case they're given an empty home directory by default.

DrDaveD:
Yes, it is quite evident from the CI failures that this is applying also when --pwd is not passed.

cclerget:
@kadern0 if you want to modify the behavior and if a warning message is not enough, the option which may work for you without modifying the default behavior could be to add a configuration directive in `apptainer.conf` allowing you to control this behavior like:

```
container working directory = strict/root-fallback
```

Where `strict` would return an error when the current working directory is not present inside the container, and the default value `root-fallback` would keep the current behavior.

kadern0:
@DrDaveD, @cclerget my apologies for not doing a more careful job, at least I should have included the original PR message. I do agree with your comments and the "original" idea was more about discussing this default "safer" behavior. Since changing this, could (and probably would) break some dependencies, I think users should be warned so their jobs keep running but at the same time they can potentially detect any misconfiguration. On the other hand, if we add a boolean attribute to the [EngineConfig's JSONConfig](https://github.com/apptainer/apptainer/blob/831a86993cd1a646576913b65afac6c4b1dc1688/pkg/runtime/engine/apptainer/config/config.go#L73), we could check whether the user has defined a custom `Cwd` and in this case, hard fail if any issues instead of defaulting to `/`.

cclerget:
@kadern0 sorry I misunderstood that it was when a user does specify `--pwd`, so yes in this case returning an error make perfect sense, by looking at the code, an additional boolean attribute is not even necessary and you could obtain it with :
```
customCwd := true
switch e.EngineConfig.OciConfig.Process.Cwd {
case e.EngineConfig.GetCwd(), e.EngineConfig.GetHomeDest():
        customCwd = false
}
if err := os.Chdir(e.EngineConfig.OciConfig.Process.Cwd); err != nil {
        if customCwd {
	        return fmt.Errorf("failed to set working directory: %s", err)
        }
	if err := os.Chdir(e.EngineConfig.GetHomeDest()); err != nil {
                os.Chdir("/")
	}
}
```

It follows the logic in actions_linux.go : 
```
	if pwd, err := os.Getwd(); err == nil {
		engineConfig.SetCwd(pwd) // record current working directory
		if PwdPath != "" {
			generator.SetProcessCwd(PwdPath) // user specified it with --pwd
		} else {
			if engineConfig.GetContain() {
				generator.SetProcessCwd(engineConfig.GetHomeDest())
			} else {
				generator.SetProcessCwd(pwd)
			}
		}
	} else {
		sylog.Warningf("can't determine current working directory: %s", err)
	}
``

**Comment by kadern0** [comment]
@cclerget, the code in your comment should solve the issue. I would only add a warning message to the user when defaulting to "/" just in case.
Do you want me to close this PR and you open a new one with your code?

**Comment by cclerget** [comment]
> @cclerget, the code in your comment should solve the issue. I would only add a warning message to the user when defaulting to "/" just in case. Do you want me to close this PR and you open a new one with your code?

@kadern0 Not required, you can integrate the changes in this PR

File: internal/pkg/runtime/engine/apptainer/process_linux.go [hunk 1]
```diff
@@ -77,8 +77,18 @@ func (e *EngineOperations) StartProcess(masterConnFd int) error {
 	bootInstance := isInstance && e.EngineConfig.GetBootInstance()
 	shimProcess := false
 
+	customCwd := true
+	switch e.EngineConfig.OciConfig.Process.Cwd {
+	case e.EngineConfig.GetCwd(), e.EngineConfig.GetHomeDest():
+		customCwd = false
```
DrDaveD, line 83: I don't think this is a sufficient check to see if the `--pwd` option was used.  My guess is that there probably needs to be an additional parameter passed through the engineConfig when `cli.PwdPath` has been set.

DrDaveD (dismissed):

internal/pkg/runtime/engine/apptainer/process_linux.go
```diff
@@ -77,8 +77,18 @@ func (e *EngineOperations) StartProcess(masterConnFd int) error {
...
+	}
+
 	if err := os.Chdir(e.EngineConfig.OciConfig.Process.Cwd); err != nil {
+		if customCwd {
+			return fmt.Errorf("failed to set working directory: %s", err)
+		}
 		if err := os.Chdir(e.EngineConfig.GetHomeDest()); err != nil {
+			sylog.Warningf("Error setting the working directory. Using '/' instead: %s", err)
```
DrDaveD, line 91: There shouldn't be a warning here, because this is normal operation in the case of `--contain`.  A debug message could be helpful.

internal/pkg/runtime/engine/apptainer/process_linux.go
cclerget: The only case where it would not report an error is with the current working directory being the same as the one specified with `--pwd`, so yes a boolean flag would be required in this case

kadern0:
Went down a third path and made use of `Annotations` which were already available. Seems to be working as intended.

DrDaveD:
This looks reasonable to me at first glance but a lot of e2e tests are failing with a stacktrace beginning
```
panic: assignment to entry in nil map
        
        goroutine 1 [running]:
        github.com/apptainer/apptainer/cmd/internal/cli.execStarter(0x55885df2c080?, {0x7fff377c1c54, 0x21}, {0xc000088500?, 0x4, 0x4}, {0x0, 0x0})
        	github.com/apptainer/apptainer/cmd/internal/cli/actions_linux.go:645 +0x2d0a
```
It looks like the Annotations map is not initialized.

You'll also need to rebase on the current main due to a conflict in CONTRIBUTORS.md, and I would prefer a single commit (although I can always do a Squash & Merge).

internal/pkg/runtime/engine/apptainer/process_linux.go
```diff
@@ -77,8 +77,14 @@ func (e *EngineOperations) StartProcess(masterConnFd int) error {
 	bootInstance := isInstance && e.EngineConfig.GetBootInstance()
 	shimProcess := false
 
+	_, customCwd := e.EngineConfig.OciConfig.Annotations["CustomCwd"]
```
DrDaveD, line 80: This passed the tests, but I was a bit surprised.  Annotations can be nil, and I guess it's ok try to do a map lookup on nil in that form.

kadern0: Yes, you can do that. Accessing maps in Golang is a safe operation.

cclerget (approved):

PR #390, opened 2022-04-21
cgroups: rootless cgroups v2 / systemd support (run/shell/exec), from sylabs 618

edytuk:
This pulls in sylabs PR

- sylabs/singularity#618
which fixed
- sylabs/singularity#300

> Allow the `--apply-cgroups` flag to be used by non-root users so long
> as:
> 
> * The system is using cgroups v2 and the systemd cgroup manager.
> * There is a D-Bus session and XDG_RUNTIME_DIR (required for the
>       systemd cgroups manager interactions).
> 
> 
> There is no support for instances, as additional work is required to
> allow instance exec'd process to be added to the cgroup without root
> permissions.
> 
> Both the EngineCongfig code, and the E2E test code, have been modified to
> pass through XDG_RUNTIME_DIR and DBUS_SESSION_BUS_ADDRESS where needed.

### DIFF
--- a/internal/pkg/util/starter/starter.go
+++ b/internal/pkg/util/starter/starter.go
@@ -2,7 +2,7 @@
 //   Apptainer a Series of LF Projects LLC.
 //   For website terms of use, trademark policy, privacy policy and other
 //   project policies see https://lfprojects.org/policies
-// Copyright (c) 2019-2020, Sylabs Inc. All rights reserved.
+// Copyright (c) 2019-2022, Sylabs Inc. All rights reserved.
 // This software is licensed under a 3-clause BSD license. Please consult the
 // LICENSE.md file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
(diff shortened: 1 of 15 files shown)


internal/pkg/util/starter/starter.go
```diff
@@ -106,6 +106,9 @@ func Run(name string, config *config.Common, ops ...CommandOp) error {
 		return fmt.Errorf("while initializing starter command: %s", err)
 	}
 
+	sylog.Debugf("Setting GOGC=off for starter")
```
cclerget, line 109: We can remove those two lines, this is not handled by our starter

cclerget (changes requested):
Overall looks good to me, just need to remove `GOGC` lines

internal/pkg/util/starter/starter.go
edytuk: I removed `GOGC` lines

cclerget (approved):

PR #435, opened 2022-05-09
feat: Add CLI action flags to apply cgroups limits, from sylabs 721

edytuk:
This pulls in sylabs PR

- sylabs/singularity#721
which fixed
- sylabs/singularity#717

> **feat: Add CLI action flags to apply cgroups limits**
> 
> Add CLI action flags that apply cgroups limits without needing the user to write a cgroups TOML file.
> 
> Most of the flags supported by Docker are now present, with the exception of `--device-` flags which will be considered later.
> 
> Note that only the long form of these flags is supported. The short form of e.g. `--cpus` is `-c` which would clash with existing use in singularity.
> 
> To implement, a JSON serialization of the LinuxResources structures is now passed through the engine configuration. Where a TOML file is used with `--apply-cgroups` it is now read and serialized to JSON, rather than passing the path to the file through to the engine.
> 
> At a later date we can/should rationalize our handing of the structures. We should likely accept JSON files as well as TOML, avoid duplicating the LinuxResources structures in our own code, etc.
> 
> **e2e: add tests for cgroups limits CLI flags**
> 
> Run a container with cgroups limits flags, and check that the cgroup the container sits in has the limit set correctly.
> 
> Covering as much of the functionality as is straightforward on v1 and v2 systems. Note that due to v2 delegation differences between distros, systemd version differences, and changes in the resource names over time, different tests will skip on different distros.
> 
> `--blkio-weight-device` is not currently tested here, as it is not straightforward to find a block device across different environments. e.g. `/dev/sda` vs `/dev/nvme0n1p1` vs `/dev/vda` etc. The flag -> resource specification translation is, however, tested in the CLI unit tests.
> 
> CI checks cgroups v1. cgroups v2 was verified on Fedora 35 with default delegation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
   container resource usage on a system using cgroups v2 and the systemd cgroups
   manager.
 - Add instance stats command.
+- Added `--cpu*`, `--blkio*`, `--memory*`, `--pids-limit` flags to apply cgroups
+  resource limits to a container directly.
 
 ### Bug fixes
 
--- a/LICENSE_DEPENDENCIES.md
+++ b/LICENSE_DEPENDENCIES.md
@@ -659,6 +659,12 @@ The dependencies and their licenses are as follows:
 
 **License URL:** <https://github.com/rivo/uniseg/blob/master/LICENSE.txt>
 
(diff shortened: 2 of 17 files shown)


cclerget (approved):

DrDaveD:
@edytuk please rebase for conflict in cmd/internal/cli/action_flags.go

PR #460, opened 2022-05-13
fix: cgroups: ensure cgroups device limits are default allow, from sylabs 790

edytuk:
This pulls in sylabs PR

- sylabs/singularity#790
which fixed
- sylabs/singularity#787

> In singularity, cgroups device limits have always defaulted to allow-all. When a cgroups config is provided with no explicit device rules, no cgroups mediated device limits have applied.
> 
> We recently switched to runc/libcontainer/cgroups as our cgroups manager (from containerd/cgroups), and this applies a default deny rule for devices.
> 
> Revert to previous behavior by asking runc/libcontainer/cgroups to skip application of device limits if no limits are provided in the spec that has been passed.

### DIFF
--- a/e2e/cgroups/cgroups.go
+++ b/e2e/cgroups/cgroups.go
@@ -99,9 +99,6 @@ func (c *ctx) instanceStats(t *testing.T, profile e2e.Profile) {
 					e2e.ExpectOutput(e2e.ContainMatch, "MEM %"),
 					e2e.ExpectOutput(e2e.ContainMatch, "BLOCK I/O"),
 					e2e.ExpectOutput(e2e.ContainMatch, "PIDS"),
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ For older changes see the [archived Singularity change log](https://github.com/a
 ### Bug fixes
 
 - Add specific error for unreadable image / overlay file.
+- Ensure cgroups device limits are default allow per past behavior.
```
DrDaveD, line 56: Actually we don't need this CHANGELOG entry because we haven't released the code it is fixing, so please remove that.

DrDaveD (changes requested):

DrDaveD (approved):

PR #475, opened 2022-06-01
Enable unprivileged container builds

### LINKED ISSUE #447: Support unprivileged container builds
Currently containers can be built unprivileged by running these commands when unprivileged user namespaces are available:
```
unshare -rm
mount --bind ~ ~root
```
Try to do the equivalent when `apptainer build` is invoked as a non-root user.  Do it after downloading an image, because the oci conveyor packer gets confused in this mode and gets Permission denied errors.

DrDaveD:
This enables container builds by unprivileged users.  Requires the fakeroot command and unprivileged user namespaces.

Does not add e2e tests.  I will make a separate issue for that.

- Fixes #447

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,7 +154,7 @@ jobs:
           go-version: 1.18.2
 
       - name: Fetch deps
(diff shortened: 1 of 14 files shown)


cclerget (approved):
LGTM

d-a-v:
Thanks !

PR #481, opened 2022-06-02
Enable unprivileged builds without unprivileged user namespaces

DrDaveD:
This adds to #475 to also allow unprivileged container builds even when unprivileged user namespaces is not available.

Inspired by a discussion in #215.

### DIFF
--- a/pkg/util/apptainerconf/config.go
+++ b/pkg/util/apptainerconf/config.go
@@ -31,6 +31,20 @@ func GetCurrentConfig() *File {
 	return currentConfig
 }
 
+// GetBuildConfig returns the configuration to be used for building containers
+func GetBuildConfig() *File {
+	// Use default configuration except without default binds
+	config, err := Parse("")
+	if err != nil {
+		sylog.Fatalf("failure getting default config: %v", err)
+	}
+	config.BindPath = nil
+	config.ConfigResolvConf = false
+	config.MountHome = false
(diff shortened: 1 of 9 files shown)


pkg/util/apptainerconf/config.go
```diff
@@ -31,6 +31,20 @@ func GetCurrentConfig() *File {
 	return currentConfig
 }
 
+// GetBuildConfig returns the configuration to be used for building containers
+func GetBuildConfig() *File {
```
cclerget, line 35: To resume `--build-config` give the opportunity for users to fully bypass possible restrictions set by admin like this is implemented actually.
Also why did you removed --nv / --nvccli stuff ?

DrDaveD: > To resume `--build-config` give the opportunity for users to fully bypass possible restrictions set by admin like this is implemented actually.

I think I understand what you're trying to say.  It should instead start from the system config when running setuid, and hope that nothing has been so restricted that it prevents the build from working.  I don't think there's a problem with disabling these 4 things anyway.

> Also why did you removed --nv / --nvccli stuff ?

They're not needed anymore now that binaries are searched at run time.  Settings for them are no longer generated at compile time.  Anyway when I change it to start from the system configuration they'll be copied too.

cclerget (approved):

PR #495, opened 2022-06-09
Make compilation and packaging unprivileged by default

### LINKED ISSUE #416: Make no setuid the default
Continuation of apptainer/singularity#3612.

The use of unprivileged fuse has since been backported to RHEL7 as well as RHEL8.

DrDaveD:
This makes `mconfig --without-suid` the default, adds a `mconfig --with-suid` option, and moves the starter-suid binary out to a separate apptainer-suid package for rpm and debian.

- Fixes #416


cmd/starter/c/starter.c
```diff
@@ -1428,6 +1428,8 @@ __attribute__((constructor)) static void init(void) {
             }
             /* master and container processes lives in the same user namespace */
             if ( create_namespace(CLONE_NEWUSER) < 0 ) {
+                infof("Adminstrator may need to enable user namespaces, install apptainer-suid\n");
```
cclerget, line 1431: typo for `Adminstrator`

cmd/starter/c/starter.c
```diff
@@ -1413,7 +1413,7 @@ __attribute__((constructor)) static void init(void) {
         if ( sconfig->starter.isSuid ) {
             priv_escalate(true);
         } else if ( uid != 0 ) {
-            fatalf("No setuid installation found, for unprivileged installation use: ./mconfig --without-suid\n");
+            fatalf("No setuid installation found\n");
```
cclerget, line 1416: Can you revert to `No setuid installation found, for privileged installation use: ./mconfig --with-suid\n` ?

internal/pkg/util/starter/starter.go
```diff
@@ -68,6 +68,16 @@ func UseSuid(suid bool) CommandOp {
...
 }
 
+// IsSuidInstall returns true if the privileged binary is configured and exists
+func IsSuidInstall() bool {
+	if buildcfg.APPTAINER_SUID_INSTALL != 1 {
+		return false
+	}
+	path := filepath.Join(buildcfg.LIBEXECDIR, "apptainer/bin/starter-suid")
+	_, err := os.Stat(path)
+	return err == nil
```
cclerget, line 78: This could be confusing in case the setuid bit is dropped for some reason but the file exist, I think we can just return true without checking its presence and let it fail later with explicit error messages

cclerget (changes requested):

cclerget (approved):

PR #516, opened 2022-06-23
feat: add environment APPTAINER_AUTH_FILE to change registry auth file

wutz:
Signed-off-by: Taizeng Wu wutaizeng@gmail.com

In some scenarios we run Apptainer using a custom authentication file rather than changing the content of the authentication file in the default location.

So we need override the default path of the authentication file by setting the `APPTAINER_AUTH_FILE` which like podman `REGISTRY_AUTH_FILE` https://docs.podman.io/en/latest/markdown/podman-login.1.html#authfile-path

### DIFF
--- a/pkg/syfs/syfs.go
+++ b/pkg/syfs/syfs.go
@@ -81,6 +81,9 @@ func RemoteCacheDir() string {
 }
 
 func DockerConf() string {
+	if authFile := os.Getenv("APPTAINER_AUTH_FILE"); authFile != "" {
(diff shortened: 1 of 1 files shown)


DrDaveD:
It looks to me like the config file and the auth file are two different things in podman.  You're changing the config file, not the authfile.  It doesn't look like we have a corresponding auth file.  We do have a $HOME/.apptainer/docker-config.json, is that not sufficient?

Issue #476, opened 2022-06-01
Make e2e tests for unprivileged container builds


DrDaveD:
#475 needs e2e tests.

DrDaveD:
Add tests for all 3 new fakeroot modes, and also add a test for old mode "1b" which I define as a non-suid installation that uses `--fakeroot` with `/etc/sub[ug]id`.  I think the current e2e tests must not be testing that because on Ubuntu it requires the installation of the `uidmap` package which we were not requiring.  It is now required by #605.

Issue #412, opened 2022-05-02
Need e2e tests for new unprivileged functionality


DrDaveD:
Add e2e tests for mounting SIF files unprivileged added in #386 and for unprivileged persistent overlay and writable tmpfs added in #411.

PR #526, opened 2022-06-30
Generalize the --fakeroot option

### LINKED ISSUE #448: Add a pseudo-root mode with unprivileged --overlay and --writable-tmpfs
The `--overlay` and `--writable-tmpfs` run/shell/exec/instance options are supported in the main branch, but they're not very useful because the user has no permission to overwrite anything.  Add a mode that does the equivalent of
```
unshare -rm
mount --bind ~ ~root
```
It could be another option similar to `--fakeroot` but not the same, such as `--fakerroot` (just kidding) or `--pseudoroot`, or it could just be made the default for the `--overlay` and `--writable-tmpfs` options.

DrDaveD:
This PR makes the `--fakeroot` option more general in that if `/etc/subuid` and `/etc/subgid` have not been set up, it will try using a root-mapped unprivileged user namespace and/or the fakeroot command.  This is especially useful with `--overlay` and `--writable-tmpfs` options to action commands and with unprivileged builds.  The PR also then effectively makes `--fakeroot` the default for unprivileged builds, changing the existing functionality so that it will try using `/etc/sub[ug]id` before the other "fake" fake root options.

- Fixes #448

In addition, it fixes nested binds to work even if the source directory was different than the destination (since that was needed for nested fakeroot command) and changes the warning on >50 binds needed for underlay to info level (since there are now a bunch more info level messages and it made me think that that message should be at that level).

e2e tests will be added as part of #412 and #476.

### DIFF
--- a/cmd/internal/cli/build_linux.go
+++ b/cmd/internal/cli/build_linux.go
@@ -17,13 +17,13 @@ import (
 	osExec "os/exec"
 	"strconv"
 	"strings"
-	"syscall"
 
 	"github.com/apptainer/apptainer/internal/pkg/build"
 	"github.com/apptainer/apptainer/internal/pkg/cache"
+	"github.com/apptainer/apptainer/internal/pkg/fakeroot"
 	"github.com/apptainer/apptainer/internal/pkg/remote/endpoint"
 	fakerootConfig "github.com/apptainer/apptainer/internal/pkg/runtime/engine/fakeroot/config"
-	"github.com/apptainer/apptainer/internal/pkg/util/bin"
+	"github.com/apptainer/apptainer/internal/pkg/util/env"
 	"github.com/apptainer/apptainer/internal/pkg/util/fs"
 	"github.com/apptainer/apptainer/internal/pkg/util/interactive"
 	"github.com/apptainer/apptainer/internal/pkg/util/starter"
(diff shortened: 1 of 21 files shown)


cmd/internal/cli/build_linux.go
```diff
@@ -33,44 +33,69 @@ import (
...
+	uid := uint32(os.Getuid())
+	if uid != 0 && !fakeroot.IsUIDMapped(uid) {
+		sylog.Infof("User not listed in %v, trying root-mapped namespace", fakeroot.SubUIDFile)
+		os.Setenv("_APPTAINER_FAKEFAKEROOT", "1")
+		err = fakeroot.UnshareRootMapped(args)
+		if err == nil {
+			// All the work has been done by the child process
+			os.Exit(0)
+		}
+		sylog.Debugf("%v", err)
```
cclerget, line 82: Need some debug context

internal/pkg/runtime/engine/apptainer/container_linux.go
```diff
@@ -1739,6 +1746,17 @@ func (c *container) addUserbindsMount(system *mount.System) error {
 	defaultFlags := uintptr(syscall.MS_BIND | c.suidFlag | syscall.MS_NODEV | syscall.MS_REC)
 
 	for _, b := range c.engine.EngineConfig.GetBindPath() {
+		if strings.HasPrefix(b.Destination, "/.singularity.d/libs") {
```
cclerget, line 1749: please add a comment explaining context and why this is required

internal/pkg/security/seccomp/seccomp_supported.go
```diff
@@ -45,7 +45,7 @@ var scmpArchMap = map[specs.Arch]lseccomp.ScmpArch{
 }
 
 var scmpActionMap = map[specs.LinuxSeccompAction]lseccomp.ScmpAction{
-	specs.ActKill:  lseccomp.ActKill,
+	specs.ActKill:  lseccomp.ActKillThread,
```
cclerget, line 48: Hmm why ?

cmd/internal/cli/build_linux.go
cclerget: And shouldn't it be a call to `sylog.Fatal` at this point ?

cclerget (changes requested):

cmd/internal/cli/build_linux.go
DrDaveD: I thought the message from the child function would be sufficient, but I'll go ahead and add context.

It isn't fatal at this point because it hasn't yet tried the fakeroot command.  That happens later in runBuild(), and then if it hasn't yet become some kind of user id == 0 and fakeroot isn't found, at that point it fails.  The only weird thing about it was the message about --fix-perms comes before that check.  I'll put a comment here and move the message about fix-perms to after when fakeroot is found if uid != 0.

DrDaveD: Oh, but in testing it just now I realized that if UnshareRootmapped fails that also implies that a suid installation is needed in order to run a build, so I'll add a check for that there too, to fail earlier even before looking for the fakeroot command.

internal/pkg/security/seccomp/seccomp_supported.go
DrDaveD: Oh, otherwise I was getting a linting failure telling me to make this change because ActKill was deprecated.   I did hit it once in another context too.  I don't know why it isn't showing up on CI checks.  Probably this should be in a separate commit, maybe even a separate PR but I don't know that it's worth that.

internal/pkg/runtime/engine/apptainer/container_linux.go
DrDaveD: comment added

internal/pkg/security/seccomp/seccomp_supported.go
cclerget: Ok, yes that's fine in this PR

cclerget (approved):

PR #539, opened 2022-07-06
Use better messages for overlays when not root or in the suid flow

### LINKED ISSUE #446: An overlay parition in a SIF file confuses unprivileged --overlay
In the main branch which supports using the `--overlay` option unprivileged, if there's an [overlay partition in a SIF file](https://apptainer.org/docs/user/main/persistent_overlays.html#overlay-embedded-in-sif) it gets confused with an incomprehensible error message:
```
FATAL:   container creation failed: mount /proc/self/fd/3->/var/apptainer/mnt/session/overlay-images/0 error: while mounting image /proc/self/fd/3: squashfuse failed to mount /var/apptainer/mnt/session/overlay-images/0 in 2s
```
If such a partition exists, ignore it.

DrDaveD:
This ignores overlay partitions in SIF images when running unprivileged in order to avoid an incomprehensible error noted in #446.  It also gives a better error message when attempting to use --writable to use an overlay partition when unprivileged.   Builds on cases introduced in #411.

- FIxes #446

### DIFF
--- a/internal/pkg/runtime/engine/apptainer/container_linux.go
+++ b/internal/pkg/runtime/engine/apptainer/container_linux.go
@@ -973,6 +973,15 @@ func (c *container) addOverlayMount(system *mount.System) error {
 			return fmt.Errorf("while opening overlay image %s: %s", img.Path, err)
 		}
 		for _, overlay := range overlays {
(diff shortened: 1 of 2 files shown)


DrDaveD:
Superceded by #543

PR #542, opened 2022-07-07
Return helpful error message for unprivileged encrypted filesystem

### LINKED ISSUE #540: Attempting to use encrypted container with non-suid flow gives obscure error
An attempt to run an encrypted container with the `--passphrase` option when there is no suid available gives a non-helpful error:
```
FATAL:   container creation failed: mount /proc/self/fd/3->/var/apptainer/mnt/session/rootfs error: while mounting image /proc/self/fd/3: squashfuse failed to mount /var/apptainer/mnt/session/rootfs in 2s
```

Change that to indicate the suid flow is needed.

DrDaveD:
- Fixes #540


DrDaveD:
Superceded by #543

PR #549, opened 2022-07-18
Ignore new markdown lint error

DrDaveD:
This ignores a new markdown lint error seen during a CI test:
```
dist/debian/DEBIAN_PACKAGE.md:91:15 MD051/link-fragments Link fragments should be valid [Context: "[start from scratch](#Preparation)"]
```
The link works fine as far as I can tell and I don't know why it is complaining.


DrDaveD:
@edytuk found a better solution to this.

PR #552, opened 2022-07-19
Remove deprecated path configs & restrict paths only on binaries run with extra privileges

### LINKED ISSUE #493: Consider removing deprecated apptainer.conf "path" settings
Consider removing the deprecated apptainer.conf "path" settings in 1.1.0:
```
cryptsetup path
go path
ldconfig path
mksquashfs path
unsquashfs path
nvidia-container-cli path
```

DrDaveD:
- Fixes #493

### DIFF
--- a/internal/pkg/util/bin/bin.go
+++ b/internal/pkg/util/bin/bin.go
@@ -23,28 +23,54 @@ import (
 )
 
 // FindBin returns the path to the named binary, or an error if it is not found.
+// We don't list any default because we want a deliberate decision about whether
+// to use the SuidBinaryPath which is more restrictive when in the suid flow.
 func FindBin(name string) (path string, err error) {
 	switch name {
 	// Basic system executables that we assume are always on PATH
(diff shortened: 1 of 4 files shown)


internal/pkg/util/bin/bin.go
```diff
@@ -23,28 +23,54 @@ import (
...
+		"mkfs.ext3",
+		"mknod",
+		"mount",
+		"rm",
+		"true",
+		"uname":
+		return findOnPath(name, true)
+	// Executables that might be run privileged from the suid flow
+	// We must not search the user's PATH when in the suid flow with these
+	case "cryptsetup":
```
cclerget, line 43: Could be moved with others above

cclerget (approved):
LGTM, just a suggestion but feel free

internal/pkg/util/bin/bin.go
DrDaveD: Yes but I like it separate to emphasize its importance.

PR #560, opened 2022-07-22
Fix loop device creation with loop-control when running inside docker container

### LINKED ISSUE #548: Issue with loop device creation when using /dev/loop-control
With changes introduced by https://github.com/apptainer/apptainer/pull/323, apptainer is requesting to kernel to allocate/create a loop device without the need to call mknod if the loop device is missing in `/dev`.

A side effect was noticed when apptainer is running inside a docker container, apptainer detect the presence of `/dev/loop-control` but the loop device is not created inside the container `/dev`, it is created on host `/dev` as docker is not binding `/dev` but recreate it.

cclerget:

 - Fixes #548


DrDaveD (changes requested):
Looks good, but please add a corresponding CHANGELOG entry

DrDaveD (approved):

PR #561, opened 2022-07-22
Fix shell runscript generation

### LINKED ISSUE #559: Apptainer is generating invalid shell syntax for runscripts
### Version of Apptainer

main branch

### Expected behavior

Being able to run SIF image build with apptainer outside of apptainer.

### Actual behavior

Generated runscript may contain invalid shell syntax (empty if branch), the shell interpreter used by apptainer accept this syntax, but this is not the case with other shell.
The issue was introduced by #438 .

```
apptainer build /tmp/alpine.sif docker://alpine:3.11
apptainer exec /tmp/alpine.sif cat /.singularity.d/runscript
...
	# ENTRYPOINT and CMD - run ENTRYPOINT with CMD as default args
	# override with user provided args
	if [ $# -gt 0 ]; then
		
	else
		set -- '/bin/sh' "$@"

		
	fi
...
```

Invalid branch:
```
bash -c "if [ \$# -gt 0 ]; then else echo "test"; fi"
bash: -c: line 1: syntax error near unexpected token `else'
bash: -c: line 1: `if [ $# -gt 0 ]; then else echo test; fi'

bash -c "if [ \$# -gt 0 ]; then ; else echo "test"; fi"
bash: -c: line 1: syntax error near unexpected token `else'
bash: -c: line 1: `if [ $# -gt 0 ]; then else echo test; fi'
```

Valid branch:
```
bash -c "if [ \$# -gt 0 ]; then echo "test"; else echo "test"; fi"
test
```

source

cclerget:

 - Fixes #559

### DIFF
--- a/internal/pkg/build/sources/conveyorPacker_oci.go
+++ b/internal/pkg/build/sources/conveyorPacker_oci.go
@@ -54,29 +54,37 @@ const ociRunscript = `
 # not evaluate resolved CMD / ENTRYPOINT / ARGS through the shell, and
 # does not modify expected quoting behavior of args.
 if [ -n "$SINGULARITY_NO_EVAL" ]; then
-	# ENTRYPOINT only - run entrypoint plus args
-	if [ -z "$OCI_CMD" ] && [ -n "$OCI_ENTRYPOINT" ]; then
-		{{.PrependEntrypoint}}
-		exec "$@"
-	fi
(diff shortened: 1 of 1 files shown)


DrDaveD (approved):
Since this fixes a bug that hasn't been released, no CHANGELOG entry is needed.

PR #575, opened 2022-07-29
Add confURL & include to Arch packer

vejnar:
Signed-off-by: vejnar <charles.vejnar@gmail.com>

Building an Arch-based container uses (i) the default Pacman configuration provided by Archlinux and (ii) install the `base` metapackage. This PR allows both of these options to be configured by user by adding the following parameters: 
* `confurl`. I added this new parameter name as none of the existing ones seemed to be appropriate.
* `include`. For building package for Archlinux, only `base-devel` (not `base`) is required. With this option, Apptainer will be usable to build proper package for Arch. I reused this parameter name as it was used in the same context of package installation.

Both parameters aren't mandatory and defaults are kept unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -146,6 +146,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Debug output is now shown for nested `apptainer` calls, in wrapped
   `unsquashfs` image extraction, and build stages.
 - Added EL9 package builds to CI for GitHub releases.
+- Add confURL & Include parameters to Arch packer for alternate `pacman.conf`
+  URL and alternate installed (meta)package.
 
 ### Bug fixes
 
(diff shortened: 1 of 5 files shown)


DrDaveD (changes requested):
This looks pretty good but will need some documentation.  Please update `examples/Arch/README.md` with info about them.  While in there please remove that erroneous reference to `Apptainer 2.3 (current development branch)`.   Also please add an entry to CHANGELOG.md new features for this.  

After merging we'll also want a PR to the [user-docs appendix](https://github.com/apptainer/apptainer-userdocs/blob/main/appendix.rst#arch-bootstrap-agent) and I'm not sure but possibly this should be mentioned in the [definition_files](https://github.com/apptainer/apptainer-userdocs/blob/main/definition_files.rst) section somewhere.

vejnar:
Thanks for your feedback. I updated README and CHANGELOG.

DrDaveD:
Looking good.  Please also add yourself to CONTRIBUTING.md

vejnar:
Ok. Done adding to CONTRIBUTORS.

DrDaveD (approved):

PR #602, opened 2022-08-09
Allow fuse-overlayfs with all file types, only warn when unexpectedly readonly

DrDaveD:
Apptainer 1.1.0 release candidate 1 checks for compatible filesystem types with the --overlay option.  That's because overlayfs is very restrictive and sends obscure messages to /var/log/messages instead of to users.  fuse-overlayfs, however, is less restrictive and in some cases it only turns the mount into readonly instead of dying.  This PR removes those checks when running in non-setuid mode and instead checks for when the mount is unexpectedly readonly by making sure /usr/bin is writable when run with fakeroot and just prints an INFO message when it is not writable.

I have tested this in a lot of combinations, and the cases where fuse-overlayfs goes readonly are when NFS and Lustre are the upper layer.  It works fine with either of them as the lower layer.  I don't have access to GPFS so I didn't try it, but I expect it is similar.

### DIFF
--- a/internal/pkg/runtime/engine/apptainer/container_linux.go
+++ b/internal/pkg/runtime/engine/apptainer/container_linux.go
@@ -1043,28 +1043,30 @@ func (c *container) addOverlayMount(system *mount.System) error {
 				}
 				system.Points.AddRemount(mount.PreLayerTag, dst, flags)
 
-				if !img.Writable {
-					// check if the sandbox directory is located on a compatible
-					// filesystem usable overlay lower directory
-					if err := fsoverlay.CheckLower(img.Path, 0); err != nil {
-						return err
+				if !overlayImageDriver {
+					// When no overlay image driver available,
+					//  make sure filesystems type are compatible
+					//  with kernel overlayfs
+					if !img.Writable {
+						// check if the sandbox directory is located on a compatible
+						// filesystem usable overlay lower directory
+						if err := fsoverlay.CheckLower(img.Path); err != nil {
+							return err
+						}
+					} else {
+						// check if the sandbox directory is located on a compatible
+						// filesystem usable with overlay upper directory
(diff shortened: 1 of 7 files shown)


internal/pkg/runtime/engine/apptainer/container_linux.go
```diff
@@ -947,6 +950,7 @@ func (c *container) addOverlayMount(system *mount.System) error {
 	nb := 0
 	ov := c.session.Layer.(*overlay.Overlay)
 	hasUpper := false
+	hasWritable := false
```
cclerget, line 953: you can remove `hasWritable` in `addOverlayMount`

internal/pkg/runtime/engine/apptainer/container_linux.go
```diff
@@ -628,6 +628,9 @@ func (c *container) mountGeneric(mnt *mount.Point, tag mount.AuthorizedTag) (err
 		dest = filepath.Join(c.session.FinalPath(), dest)
 	} else {
 		dest = mnt.Destination
+		if mnt.Type == "overlay" && c.session.GetReadonly() {
```
cclerget, line 631: This is detected by kernel overlay, all you need is to parse overlay mount options in the image driver and look if `upperdir` is present, if not you can consider this is a readonly mount

internal/pkg/util/fs/layout/session_linux.go
```diff
@@ -26,7 +26,8 @@ const (
 // Session directory layout manager
 type Session struct {
 	*Manager
-	Layer layer
+	Layer    layer
+	readonly bool
```
cclerget, line 30: This is not necessary

cclerget (changes requested):

cclerget (approved):

PR #612, opened 2022-08-11
Use user's home directory as /root in fakeroot mode

### LINKED ISSUE #583: Suid flow with --fakeroot --writable-tmpfs attempts to access real root home dir
On the apptainer main branch with apptainer-suid installed and rootless fakeroot I see
```
bash: /root/.bashrc: Permission denied
```
This doesn't happen without apptainer-suid or without --writable-tmpfs.

DrDaveD:
This makes the user's home directory be mounted as /root with `--fakeroot` even in the setuid flow.  With `--writable-tmpfs` it was especially noticeable because it showed Permission denied on `/root`, but even without that it was using the original user's home directory path which didn't make a lot of sense.

- Fixes #583

This also patches up the Usage for the `--userns` action option because I happened to notice that as I was testing this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Replaced checks for compatible filesystem types when using fuse-overlayfs
   with an INFO message when an incompatible filesystem type causes it to
   be unwritable by a fakeroot user.
+- Mount the user's home directory at `/root` when using `--fakeroot` in
+  the setuid flow (fixes a regression introduced in 1.1.0-rc.1 which didn't
+  impact non-setuid flow).
 - The `--nvccli` option now works without `--fakeroot`.  In that case the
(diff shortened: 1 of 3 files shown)


cclerget:
@DrDaveD looks like a regression, tested with apptainer 1.0.3 and `--fakeroot` bind user home directory to `/root` inside container, it is originally done in this chunk https://github.com/apptainer/apptainer/blob/990dd90f9df473a9cb9b78847e2df4bd8a931529/cmd/internal/cli/actions_linux.go#L541

DrDaveD:
Yes I agree it was a regression I introduced in this [commit](https://github.com/apptainer/apptainer/commit/088b9c57b3a6544d62681fa1263d5e93ebc98286) in #526.

DrDaveD:
e2e tests to prevent it happening again will be added in #618

ikaneshiro (approved):

PR #638, opened 2022-08-19
adding support for --sparse when creating ext3, from sylabs 864

edytuk:
This pulls in sylabs PR

- sylabs/singularity#864
 which fixed
- sylabs/singularity#610

> We basically want to add support for a sparse overlay, and we can do that by using truncate -s (size) instead of dd. In the case the command is not found, an error is issued and the user can decide to install or not use it. I think it would be overkill to hard code information about all OS versions / support but if anyone has ideas for how to reasonably do this we can return a more specific error message. I am not running the e2e tests locally but will see how they do in the CI! :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ The Singularity Project has been
 and re-branded as Apptainer.
 For older changes see the [archived Singularity change log](https://github.com/apptainer/singularity/blob/release-3.8/CHANGELOG.md).
 
+## Changes Since Last Release
+
+### New features / functionalities
+
+- Add `--sparse` flag to `overlay create` command to allow generation of a
+  sparse ext3 overlay image.
+
 ## v1.1.0-rc.2 - \[2022-08-16\]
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ The Singularity Project has been
 and re-branded as Apptainer.
 For older changes see the [archived Singularity change log](https://github.com/apptainer/singularity/blob/release-3.8/CHANGELOG.md).
 
+## Changes Since Last Release
+
+### New features / functionalities
```
DrDaveD, line 10: I'd like to include this in 1.1.0-rc.3 and for non-major releases we don't distinguish between the type of change.  So please delete this line.  When you redo the commit, change the title to refer to the apptainer PR.

For the final 1.1.0 release I will merge the changes in rc.2 and rc.3 into one full release announcement.

DrDaveD (changes requested):

DrDaveD (approved):

PR #640, opened 2022-08-19
feat: `bind-paths` options for `--no-mount` flag, from sylabs 885

edytuk:
This pulls in sylabs PR

- sylabs/singularity#885
 which fixed
- sylabs/singularity#883

> Add `bind-paths` option for the `--no-mount` flag. This will disable mounting all of the `bind path` entries set in `singularity.conf`.
> 
> Previously, `bind path` entries could only be disabled one-by-one by specifying their absolute path.


DrDaveD (changes requested):
Please put this whole changelog entry just below the current `## Changes Since Last Release` line at the top, and below that add `## Changes Since Last Release Candidate`.

DrDaveD (approved):

PR #641, opened 2022-08-19
feat: enable cgroups for non-root instances, from sylabs 896

edytuk:
This pulls in sylabs PR

- sylabs/singularity#896
 which fixed
- sylabs/singularity#816

> **feat: enable cgroups for non-root instances**
> 
> Our existing method for joining an instance cgroup with an action on an `instance://` does not work for non-root instances, without a usernamepace - so we blocked non-root instances with cgroups. The existing code attempts to move its parent ID (the starter main process), into the instance cgroup. This will receive an operation not permitted error.
> 
> We can work around by having the CLI process directly join the cgroup, so that everything is spawned within the instance cgroup.
> 
> Also modify e2e cgroups tests to verify this functionality, and correct unintentional test skips in rootless flow.
> 
> **e2e: use cpu controller instead of cpuset for cgroups test**
> 
> The cpuset controller may not be delegated by default on many distributions.
> 
> **fix: handle invalid wd before shell interp.New**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 - The `--no-mount` flag now accepts the value `bind-paths` to disable mounting of
   all `bind path` entries in `apptainer.conf.
+- Instances started by a non-root user can use `--apply-cgroups` to apply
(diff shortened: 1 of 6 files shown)


DrDaveD:
Please resolve merge conflict

DrDaveD (approved):

PR #674, opened 2022-09-02
In fakeroot setuid hybrid mode, have image drivers enter container namespace

### LINKED ISSUE #651: /etc/subuid-based fakeroot with --writable-tmpfs in suid flow has many failures
Lots of different failures happen with various operating system SIF images when attempting to install openssh in the suid flow and --writable-tmpfs, even after increasing the `sessiondir max size`.  For instance on a centos7 SIF it says:
```
error: unpacking of archive failed on file /usr/bin/fipscheck;630678a2: cpio: open
...
error: unpacking of archive failed on file /etc/prelink.conf.d/fipscheck.conf;630678a2: cpio: chmod
```
It works with either `--userns`, a sandbox, or with `--overlay` instead of `--writable-tmpfs`.  What's the difference with `--writable-tmpfs`?

DrDaveD:
This fixes write problems seen when using the setuid-root fakeroot "hybrid" mode with overlay.  It passes the container process id to the Start() function of the image driver so the driver can use nsenter to join the container's user namespace before invoking helper fuse processes.

- Fixes #651

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,9 @@ For older changes see the [archived Singularity change log](https://github.com/a
   sparse ext3 overlay image.
 - Support for a custom hashbang in the `%test` section of an Apptainer recipe
   (akin to the runscript and start sections).
-- Skip trying to user kernel overlayfs when using writable overlay and the
+- When using fakeroot in setuid mode, have the image drivers first enter the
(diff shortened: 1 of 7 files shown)


cclerget (approved):
Need to fix `internal/pkg/image/driver/imagedriver.go:129: File is not gofumpt-ed (gofumpt)`

PR #689, opened 2022-09-07
Make an e2e test for new ECL whitelist[strict] logic

JasonYangShadow:
Signed-off-by: jason yang <jasonyangshadow@gmail.com>

## Description of the Pull Request (PR):

Make an e2e test for new ECL whitelist[strict] logic

### This fixes or addresses the following GitHub issues:

 - Fixes #578 

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/apptainer/apptainer/blob/main/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTORS.md)

PR #705, opened 2022-09-21
Document the fact that sessiondir max size also affects --writable-tmpfs

### LINKED ISSUE #636: `--writable-tmpfs` in suid mode gets too small of space
The default filesystem size used by `--writable-tmpfs` in suid mode is set by the `sessiondir max size` apptainer.conf configuration and defaults to only 16M.  That's not of much use, and the comment in that setting in apptainer.conf doesn't say it is related or applies to only suid mode.

DrDaveD:
- Fixes #636


DrDaveD:
I believe the e2e tests failures are due to changes from the recently merged #689.  They are variations of the error
```
while validating ECL configuration: lstat /tmp/ecl-2922172027: no such file or directory
```
I don't know what to do about it other than backing out the commit from #689.

cclerget (approved):

PR #706, opened 2022-09-21
Fix misleading fakeroot error

### LINKED ISSUE #692: fakeroot build shows misleading error if subuid contains less than 65536 entries
### Version of Apptainer

`apptainer version 1.1.0~rc.3-1.el7`

### Expected behavior

I'd like to get a meaningful error message

### Actual behavior
```
INFO:    User not listed in /etc/subuid, trying root-mapped namespace
```

Create an entry for your user account in /etc/subuid with a value of less than 65536 entries.
then run this command under the user:
```
singularity build --fakeroot --sandbox testbuild Apptainer
```

```sh
$ cat /etc/os-release
NAME="Scientific Linux"
VERSION="7.9 (Nitrogen)"
ID="scientific"
ID_LIKE="rhel centos fedora"
VERSION_ID="7.9"
PRETTY_NAME="Scientific Linux 7.9 (Nitrogen)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:scientificlinux:scientificlinux:7.9:GA"
HOME_URL="http://www.scientificlinux.org//"
BUG_REPORT_URL="mailto:scientific-linux-devel@listserv.fnal.gov"

REDHAT_BUGZILLA_PRODUCT="Scientific Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.9
REDHAT_SUPPORT_PRODUCT="Scientific Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.9"

```

from EPEL-testing as rpm

DrDaveD:
- Fixes #692


internal/pkg/fakeroot/fakeroot.go
```diff
@@ -388,10 +388,10 @@ func IsUIDMapped(uid uint32) bool {
 		sylog.Fatalf("could not retrieve user with UID %d: %s", uid, err)
 	}
 	e, err := config.GetUserEntry(userinfo.Name)
-	if err != nil {
+	if err != nil && strings.Contains(fmt.Sprintf("%v", err), "no mapping entry found") {
```
cclerget, line 391: a cleaner way would be to use wrap errors and use `errors.Is`

cclerget (changes requested):

internal/pkg/fakeroot/fakeroot.go
```diff
@@ -388,10 +388,10 @@ func IsUIDMapped(uid uint32) bool {
...
 		return false
 	}
-	if e.disabled {
+	if e != nil && e.disabled {
```
cclerget, line 394: the function will return true for non existing users

internal/pkg/fakeroot/fakeroot.go
DrDaveD: What do you mean by "wrap errors"?

internal/pkg/fakeroot/fakeroot.go
DrDaveD: I don't think so, because the previous check for "getPwUID()" causes a fatal error if the user does not exist.  Things would be quite messed up if the user does not exist.

internal/pkg/fakeroot/fakeroot.go
```diff
@@ -388,10 +393,10 @@ func IsUIDMapped(uid uint32) bool {
 		sylog.Fatalf("could not retrieve user with UID %d: %s", uid, err)
 	}
 	e, err := config.GetUserEntry(userinfo.Name)
-	if err != nil {
+	if err != nil && errors.Is(err, errNoMappingEntry) {
```
cclerget, line 396: Should it be `if err != nil && !errors.Is(err, errRangeTooLow)` ?

DrDaveD: Either way should work for now, but the question is what would happen if a new error case came up in the future.  It's a guess, but I think it is more likely that a new error case will more likely be one that should be ignored than included.  So I think the current check is the correct one.  This function is called "IsUIDMapped" sp all we want to check is whether or not it is mapped, which is what this error check is looking for.

cclerget: I think it's more explicit and safe to consider all errors and return false except for `errRangeTooLow` to avoid introduction of potential issues in the future

DrDaveD: I still disagree but not strongly enough to object further, so I made the change.

cclerget (approved):

PR #707, opened 2022-09-21
Document the fact that sessiondir max size also affects --writable-tmpfs (release-1.1)

DrDaveD:
This cherry-picks #705 into the release-1.1 branch


DrDaveD:
Created against the wrong branch.

PR #715, opened 2022-09-22
Adding X.509 support for Supercomputing Centers 

fnikolai:
Singularity/Apptainer supports the capability to use PGP keys for signing and verifying containers, thus improving the level of trust for users that need to share containers.

PGP is based on the "Web of Trust" principle, where communicating parties must have at least one trusted party in common, who will sign both keys. PGP relies on asymmetric cryptography (i.e., the creation and usage of public-private key pairs).

X.509 is an alternative security approach, also relying on asymmetric cryptography, but is usually deployed in environments where certification authorities (CAs) vouch for each certificate, meaning the overall system architecture needs a CA. Anyone who wants to participate in the system must have their keys signed by that CA. Certificate creation and validation rely on a chain of trust established among CAs.

However, sharing containers among Supercomputing Center (SC) customers assumes a trusting relationship with the code providers and the SC itself. We advocate that a Supercomputing Center can provide Public Key Infrastructure (PKI) services to its customers, assuming the CA's role in the trusted sharing of pre-validated codes that are packaged as container images. We consider use cases where containers include specialized software packages and libraries, including their configuration parameter settings, for HPC codes expected to operate on sensitive datasets in a wide range of scientific domains.

This PR implements the proposed extensions to the codebase of Singularity/Apptainer so that X.509 certificates, signed by the SC's CA, can be embedded in SIF-formatted container binary image files and then be checked online for validity (incl. checks for revocation).

== Important Note ==
1) the X.509 certificates are supported without removing the existing support for PGP certificates.

2) The implementation of this PR depends on Sif modifications implemented in https://github.com/apptainer/sif/pull/144 -- Thus, the acceptance of this PR should be postponed until the Sif's PR is accepted first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Support for `DOCKER_HOST` parsing when using `docker-daemon://`
 - `DOCKER_USERNAME` and `DOCKER_PASSWORD` supported without `APPTAINER_` prefix.
 
-## Changes Since Last Release Candidate
+## v1.1.0-rc.3 - \[2022-09-06\]
 
 - Imply adding `${prefix}/libexec/apptainer/bin` to the `binary path` in
   `apptainer.conf`, which is used for searching for helper executables.
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -136,7 +136,7 @@ To build a specific version of Apptainer, check out a
 for example:
 
 ```sh
-git checkout v1.1.0-rc.2
+git checkout v1.1.0-rc.3
 ```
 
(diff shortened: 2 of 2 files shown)


DrDaveD:
Thanks for your significant contribution!  There's quite a few CI failures, so please check those out and try to fix them.  I suppose some of them are because of the pending sif changes.

Unfortunately since you are a first-time contributor, one of the maintainers will have to click approve each time you make an update before the tests will run.

DrDaveD:
Oh, at this point it might be a lot easier to create a pull request first in your own github repo and make sure all the CI checks pass there first.

DrDaveD:
Ok apptainer/sif#147 is now merged so this can proceed.

PR #716, opened 2022-09-23
e2e: switch to shared cache by default, from sylabs 975

edytuk:
This pulls in sylabs PR

- sylabs/singularity#975

> The SingularityCmd construct used to execute singularity in the e2e tests has defaulted to a fresh, empty cache directory for each execution.
> 
> This means that we are downloading full container images on every e2e execution of singularity against docker or library, which slows things down and creates uneccesary request load on the Sylabs Library, Docker Hub, etc.
> 
> Switch to using a shared cache by default. Our cache structure has been concurrency safe (given atomic operations on local filesystems) for some time now crossed_fingers ... this will also now test that.
> 
> Pull tests are modified so that they explicitly use an empty cache, to test the entire pull flow, including container image download.
> 
> Any future tests that depend on exercising downloads, the cache specifically, or depend on a specific state / order of cached operations will need to also specify their own cache directory.

### DIFF
--- a/e2e/actions/regressions.go
+++ b/e2e/actions/regressions.go
@@ -306,7 +306,7 @@ func (c actionTests) issue4823(t *testing.T) {
 			expected = e2e.ExpectError(e2e.ContainMatch, "Using image from cache")
 		}
 
-		c.env.ImgCacheDir = cacheDir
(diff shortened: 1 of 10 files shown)


DrDaveD:
replaced by #725

PR #725, opened 2022-09-28
e2e: switch to shared cache by default and re-work docker / pull related e2e tests, from sylabs 975 & 982 & 984 & 986

edytuk:
This pulls in sylabs PRs

- sylabs/singularity#975
- sylabs/singularity#982
- sylabs/singularity#984
- sylabs/singularity#986
 which fixed
- sylabs/singularity#977

> The SingularityCmd construct used to execute singularity in the e2e tests has defaulted to a fresh, empty cache directory for each execution.
> 
> This means that we are downloading full container images on every e2e execution of singularity against docker or library, which slows things down and creates uneccesary request load on the Sylabs Library, Docker Hub, etc.
> 
> Switch to using a shared cache by default. Our cache structure has been concurrency safe (given atomic operations on local filesystems) for some time now crossed_fingers ... this will also now test that.
> 
> Pull tests are modified so that they explicitly use an empty cache, to test the entire pull flow, including container image download.
> 
> Any future tests that depend on exercising downloads, the cache specifically, or depend on a specific state / order of cached operations will need to also specify their own cache directory.

> E2E tests using docker:// sources cannot run in parallel with caching, due to issues with concurrency and the containers/image oci/layout. We want to use caching as many tests rely on the same image, and Docker Hub rate limiting is still an issue where it is not trivial to provide credentials.
> 
> This changeset:
> 
> * Replaces some docker:// sources with equivalent library:// sources, in tests that are not explicitly for verifying docker specific
>       behavior. library:// caching is concurrency safe on local fs with atomic operations (such as CI), allowing a speed-up.
> 
> * Removes some redundant docker:// source tests from env/imgbuild etc. which are already covered by the docker group.
> 
> * Moves all remaining docker:// source tests into the e2e.docker group, so that all tests using Docker Hub are isolated there. This could be used in future to disable them for e2e runs without Docker Hub credentials etc.
> 
> * Runs the e2e.docker tests ordered sequentially, but parallel with respect to other non-docker tests.
> 
> * Replaces a problematic Chdir operation in the non-docker pull tests with specifying the working directory on the exec.Command, so that these pull tests can also be run ordered sequentially, but parallel with respect to other non-pull tests.
> 
> 
> Speed-up reported on CircleCI from these changes is ~20%

> Two e2e tests with docker sources were missed in sylabs/singularity#982 due to a case-sensitive search for bootstrap: docker.
> 
> Move one to the docker e2e group. Modify the other to use a library alpine image.

> One E2E PULL test manipulates the process umask. To avoid causing permissions issues in other tests it must be run alone, in the SEQ section of the E2E suit, via testhelper.NoParallel.
> 
> Convert IMGBUILD def file tests docker -> library

### DIFF
--- a/e2e/actions/actions.go
+++ b/e2e/actions/actions.go
@@ -385,20 +385,6 @@ func (c actionTests) STDPipe(t *testing.T) {
 			input:   "false",
 			exit:    1,
 		},
-		{
-			name:    "TrueDocker",
-			command: "shell",
-			argv:    []string{"docker://busybox"},
-			input:   "true",
-			exit:    0,
-		},
-		{
-			name:    "FalseDocker",
-			command: "shell",
-			argv:    []string{"docker://busybox"},
-			input:   "false",
-			exit:    1,
-		},
 		// TODO(mem): reenable this; disabled while shub is down
 		// {
 		// 	name:    "TrueShub",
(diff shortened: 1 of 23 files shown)


cclerget (approved):

DrDaveD:
@edytuk please rebase.

PR #747, opened 2022-10-06
Add support for x509

fnikolai:
Signed-off-by: Fotis Nikolaidis <nikolaidis.fotis@gmail.com>

This PR follows the discussion on https://github.com/apptainer/sif/pull/149

== Use-Case ==
Singularity/Apptainer supports the capability to use PGP keys for signing and verifying containers, thus improving the level of trust for users that need to share containers.

PGP is based on the "Web of Trust" principle, where communicating parties must have at least one trusted party in common, who will sign both keys. PGP relies on asymmetric cryptography (i.e., the creation and usage of public-private key pairs).

X.509 is an alternative security approach, also relying on asymmetric cryptography, but is usually deployed in environments where certification authorities (CAs) vouch for each certificate, meaning the overall system architecture needs a CA. Anyone who wants to participate in the system must have their keys signed by that CA. Certificate creation and validation rely on a chain of trust established among CAs.

However, sharing containers among Supercomputing Center (SC) customers assumes a trusting relationship with the code providers and the SC itself. We advocate that a Supercomputing Center can provide Public Key Infrastructure (PKI) services to its customers, assuming the CA's role in the trusted sharing of pre-validated codes that are packaged as container images. We consider use cases where containers include specialized software packages and libraries, including their configuration parameter settings, for HPC codes expected to operate on sensitive datasets in a wide range of scientific domains.

This PR implements the proposed extensions to the codebase of Singularity/Apptainer so that X.509 certificates, signed by the SC's CA, can be embedded in SIF-formatted container binary image files and then be checked online for validity (incl. checks for revocation).

Important Note: the X.509 certificates are supported without removing the existing support for PGP certificates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,14 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Support for `DOCKER_HOST` parsing when using `docker-daemon://`
 - `DOCKER_USERNAME` and `DOCKER_PASSWORD` supported without `APPTAINER_` prefix.
 
+- Added support for signing and validating images using X509 certificates. For
+  the signing the flags --pkcs8Key and --x509cert are added. In the respective
+  order, the flags load the private key and the certificate of the authority
+  signing the image (signer). For the validation, --x509cert, --x509RootCA,
+  and --x509IntermediateCerts. In the respective order, the load the signer's
+  certificate, the certificate of the Root CA who signed the signer's certificate,
+  and the certificates of intermediate CAs.
+
 ## v1.1.2 - \[2022-10-06\]
 
 ### Changes since last release
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ For older changes see the [archived Singularity change log](https://github.com/a
   the signing the flags --pkcs8Key and --x509cert are added. In the respective
   order, the flags load the private key and the certificate of the authority
   signing the image (signer). For the validation, --x509cert, --x509RootCA,
-  and --x509IntermediateCerts. In the respective order, the load the signer's
+  and --x509IntermediateCerts. In the respective order, they load the signer's
   certificate, the certificate of the Root CA who signed the signer's certificate,
   and the certificates of intermediate CAs.
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -31,6 +31,7 @@
 - Eng Zer Jun <engzerjun@gmail.com>
 - Eric Müller <mueller@kip.uni-heidelberg.de>
 - Felix Abecassis <fabecassis@nvidia.com>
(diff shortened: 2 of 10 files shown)


DrDaveD (changes requested):
Please add an entry describing the change to CHANGELOG.md and add yourself to CONTRIBUTORS.md

CHANGELOG.md
```diff
@@ -19,6 +19,14 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Support for `DOCKER_HOST` parsing when using `docker-daemon://`
 - `DOCKER_USERNAME` and `DOCKER_PASSWORD` supported without `APPTAINER_` prefix.
 
+- Added support for signing and validating images using X509 certificates. For
+  the signing the flags --pkcs8Key and --x509cert are added. In the respective
+  order, the flags load the private key and the certificate of the authority
+  signing the image (signer). For the validation, --x509cert, --x509RootCA,
+  and --x509IntermediateCerts. In the respective order, the load the signer's
```
DrDaveD, line 26: Change "the load" to either "they load" or "the flags load"

DrDaveD (changes requested):

CHANGELOG.md
fnikolai: done

fnikolai: should be ok now

DrDaveD:
@fnikolai @ikaneshiro There's a related discussion on this in the [SingularityCE mailing list](https://groups.google.com/d/msgid/singularity-ce/c97911b1-97ce-4703-8bef-f61760b5ad09n%40googlegroups.com).  It would be good to know what you think about Adam's idea.

fnikolai:
@DrDaveD Thanks. I 'll follow up on the discussion. 

A few comparison points:
-- Signing Protocol -- (X509 vs DSSE)
1) x509 is RFC-based https://datatracker.ietf.org/doc/html/rfc5280
2) x509 is already well-established -- there is no chance for a slow-moving community like HPC to adopt anything else soon.


-- X509 Implementation -- (this PR vs Sigstore/Cosign)
Perhaps another question is whether to prefer this PR or the code used in Sigstore's cosign[1].

At a top level, the concepts and techniques used in the two implementations have many in common. However, their implementation seems to lack revocation checks. They only check that the image is signed from the given certificate but do not check if the certificate is revoked.
(* I went through it fast, so I may be wrong)

https://github.com/sigstore/cosign/blob/ca0959a1c76cd1cb03a125a836077f0223a57874/pkg/cosign/verify.go

DrDaveD:
@fnikolai This is superceded by #1048, correct?  We should close this one?

GodloveD:
@DrDaveD I think that this is related to #1048, but there are more PRs to add to complete the functionality.

DrDaveD:
Superseded by #1048

PR #808, opened 2022-10-19
remove /rootfs-* like directory when error occurs

### LINKED ISSUE #768: $TMPDIR/rootfs-* not cleaned up after an unsquashfs error
See [#760 comment](760#issuecomment-1274951911).  I'm not sure if there might be other cases of errors where it isn't cleaned up too.

JasonYangShadow:
Signed-off-by: jason yang <jasonyangshadow@gmail.com>

We need to clean up the $TMPDIR/rootfs-*/tmp-rootfs-* when error occurs

 - Fixes #768 

### DIFF
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -28,6 +28,7 @@ import (
 	"github.com/apptainer/apptainer/internal/pkg/image/unpacker"
 	"github.com/apptainer/apptainer/internal/pkg/instance"
 	"github.com/apptainer/apptainer/internal/pkg/plugin"
+	"github.com/apptainer/apptainer/internal/pkg/runtime/engine/apptainer"
 	"github.com/apptainer/apptainer/internal/pkg/runtime/engine/config/oci"
 	"github.com/apptainer/apptainer/internal/pkg/runtime/engine/config/oci/generate"
 	"github.com/apptainer/apptainer/internal/pkg/security"
@@ -839,11 +840,12 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 			sylog.Verbosef("User namespace requested, convert image %s to sandbox", image)
 			sylog.Infof("Converting SIF file to temporary sandbox...")
 			rootfsDir, imageDir, err := convertImage(image, unsquashfsPath, tmpDir)
-			if err != nil {
-				sylog.Fatalf("while extracting %s: %s", image, err)
-			}
 			engineConfig.SetImage(imageDir)
 			engineConfig.SetDeleteTempDir(rootfsDir)
+			if err != nil {
+				sylog.Fatalf("while extracting %s: %s", image, err)
+				defer apptainer.CleanupImageTemp(engineConfig)
+			}
 			generator.SetProcessEnvWithPrefixes(env.ApptainerPrefixes, "CONTAINER", imageDir)
 
 			// if '--disable-cache' flag, then remove original SIF after converting to sandbox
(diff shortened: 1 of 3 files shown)


DrDaveD (changes requested):
We don't want to remove all rootfs-* directories, just the specific one.

I ran a successful `--unsquash` with debugging on and very near the end it shows:
```
VERBOSE [U=3382,P=4289]    CleanupContainer()            Removing image tempDir /scratch/tmp/rootfs-2834231964
```
So you can do something similar to what [CleanupContainer()](https://github.com/apptainer/apptainer/blob/main/internal/pkg/runtime/engine/apptainer/cleanup_linux.go#L58) does.  It actually looks pretty complicated so it's probably worth making a function here that's also used from CleanupContainer().
```

**Comment by JasonYangShadow** [comment]
> We don't want to remove all rootfs-* directories, just the specific one.
> 
> I ran a successful `--unsquash` with debugging on and very near the end it shows:
> 
> ```
> VERBOSE [U=3382,P=4289]    CleanupContainer()            Removing image tempDir /scratch/tmp/rootfs-2834231964
> ```
> 
> So you can do something similar to what [CleanupContainer()](https://github.com/apptainer/apptainer/blob/main/internal/pkg/runtime/engine/apptainer/cleanup_linux.go#L58) does. It actually looks pretty complicated so it's probably worth making a function here that's also used from CleanupContainer().

actually, it won't remove the entire rootfs-* folder, but only remove the sub-folder having the prefix `tmp-rootfs-*` generated here
https://github.com/JasonYangShadow/apptainer/blob/157eca684e499f0bd77d16cdbf4b86b2499f14a5/internal/pkg/image/unpacker/squashfs_apptainer.go#L159
anyway, the cleanup function should be optimized and called appropriately in this case.

File: internal/pkg/runtime/engine/apptainer/cleanup_linux.go [hunk 1]
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/apptainer/apptainer/internal/pkg/util/priv"
 	"github.com/apptainer/apptainer/internal/pkg/util/starter"
 	"github.com/apptainer/apptainer/pkg/build/types"
+	apptainer "github.com/apptainer/apptainer/pkg/runtime/engine/apptainer/config"
```
DrDaveD, line 28: Most other source files in this directory call this `apptainerConfig`.  I see it is inconsistent in container_linux.c.  It should probably be changed there to `apptainerConfig` too.

cmd/internal/cli/actions_linux.go
```diff
@@ -839,11 +840,12 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 			sylog.Verbosef("User namespace requested, convert image %s to sandbox", image)
 			sylog.Infof("Converting SIF file to temporary sandbox...")
 			rootfsDir, imageDir, err := convertImage(image, unsquashfsPath, tmpDir)
+			engineConfig.SetImage(imageDir)
+			engineConfig.SetDeleteTempDir(rootfsDir)
 			if err != nil {
 				sylog.Fatalf("while extracting %s: %s", image, err)
+				defer apptainer.CleanupImageTemp(engineConfig)
```
DrDaveD, line 847: sylog.Fatalf() does not return.  So just invoke this function ahead of the that call.

DrDaveD (changes requested):

DrDaveD:
I'm sorry, Jason, I steered you wrong on this one.  I tried to test this and I see there's actually an attempt to remove the directory inside the `convertImage` function that isn't working.  Also there's more complication in `CleaupImageTemp` than is needed here, because it is not actually running in a separate process like it is over in the engine.  The only thing additional think that needs to be here in addition to the `RemoveAll` is a call to `types.FixPerms`.  I am working on a separate fix.

JasonYangShadow:
> 

yes, as you explained, it looks like that it is more complex than I thought. Thanks for the info and the separate fix.

PR #827, opened 2022-10-25
Change non-writable overlay warning into info

DrDaveD:
It just occurred to me this should have been an info, not a warning.


DrDaveD:
I changed my mind, the current way  is fine.

PR #826, opened 2022-10-25
set --net if --network or --network-args is set

jcphill:
## Description of the Pull Request (PR):

Set the `--net` flag if `--network` or `--network-args` is set rather than silently ignoring them if `--net` was not set.

### This fixes or addresses the following GitHub issues:

 - Fixes https://github.com/apptainer/apptainer/issues/822

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/apptainer/apptainer/blob/main/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTORS.md)

PR #835, opened 2022-10-26
chore: refactor actions_linux.go / execStarter, from sylabs 1003

edytuk:
This pulls in sylabs PR

- sylabs/singularity#1003

> The execStarter function, which creates an EngineConfig defining how to run a container, and then executes the starter binary, is extremely long and complex. It accesses a lot of CLI flags directly, as global variables, and is generally hard to reason about, especially where it modifies those global vars in a way that changes behavior later.
> 
> The code shouldn't really be in the CLI layer. To handle execution of OCI images with an OCI native runtime we need to be able to separate the CLI from starter execution.
> 
> As a first step in refactoring to move this code out, this PR:
> 
> * Renames actions_linux.go to starter_linux.go to reflect the fact this code is all about calling the starter, and is not just for actions... it's used by instances too.
> 
> * Does some extraction of code into functions, to make execStarter itself easier to follow.
> 
> * Re-orders some code (where it is not order sensitive), to group like operations together (e.g. mount handling).
> 
> 
> Note that the starter execution is still tied to the global vars backing CLI flags, and modifies them. This is a first step of many :-)


DrDaveD:
Please rebase on current main, after merge of #826

cmd/internal/cli/starter_linux.go
```diff
@@ -695,6 +873,9 @@ func execStarter(cobraCmd *cobra.Command, image string, args []string, name stri
 		sylog.Infof("Setting --net (required by --network-args)")
 		NetNamespace = true
 	}
+	// unprivileged installation could not use fakeroot
+	// network because it requires a setuid installation
+	// so we fallback to none
```
DrDaveD, line 878: This comment should be left where it was, above the check starting `if !starter.IsSuidInstall()` below.

PR #851, opened 2022-11-02
fix: don't freeze for http(s) pull without content-length, from sylabs 1088

edytuk:
This pulls in sylabs PR

- sylabs/singularity#1088
 which fixed
- sylabs/singularity#1087

> When an http(s) pull doesn't provide a content-length header, we need to ensure the progress bar completes. If we don't do this, the CLI hangs indefinitely.


DrDaveD (approved):

DrDaveD:
Please also submit a release-1.1 PR that cherry-picks this commit.

PR #857, opened 2022-11-07
feat: Allow kernel supported unpriv overlay, from sylabs 1077 & 1081

edytuk:
This pulls in sylabs PRs

- sylabs/singularity#1077
- sylabs/singularity#1081
 which fixed
- sylabs/singularity#818
- sylabs/singularity#1078

> When we are running a container in a user namespace, check whether the kernel supports rootless overlay. If it does, use this instead of underlay.
> 
> At present, the visible effects of this are that:
> 
>  * When running under a user namespace, on a supported system, the overlay session layout is used. The rootfs in the container will appear as an overlay fs mount.
>  * When running under a user namespace, on a supported system, the `--writable-tmpfs` option can be used, with the same semantics as in set-uid mode.
> 
> 
> Note that this changeset does not allow unprivileged persistent overlays (singularity `--overlay` flag). It will be possible to enable unprivileged overlay directory mounts at a later point, and potentially other overlay mounts via FUSE drivers.
> 
> The principle benefit at this stage is to avoid the complex, and numerous, bind mounts used by the underlay layout.
> 
> ToDo: Refactor the underlay / overlay handling to a simpler, more linear flow.
> 
> **Testing note** - CircleCI is using an Ubuntu 22.04 image that supports unprivileged overlay. The non-unpriv overlay flow can be tested on e.g. CentOS 7.

> Kernel unprivileged overlay now means we can permit users to request a persistent overlay from a directory, when running in a non-setuid flow.
> 
> In setuid mode it still remains unsafe, as the mount occurs privileged.

### DIFF
--- a/internal/pkg/util/fs/overlay/overlay_linux.go
+++ b/internal/pkg/util/fs/overlay/overlay_linux.go
@@ -10,8 +10,15 @@
 package overlay
 
 import (
+	"errors"
 	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"syscall"
 
+	"github.com/apptainer/apptainer/internal/pkg/util/bin"
+	"github.com/apptainer/apptainer/pkg/sylog"
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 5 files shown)


internal/pkg/util/fs/overlay/overlay_linux.go
```diff
@@ -133,3 +140,71 @@ func IsIncompatible(err error) bool {
...
+		return fmt.Errorf("while creating temporary directories: %w", err)
+	}
+
+	args := []string{
+		"-t", "overlay",
+		"-o", fmt.Sprintf("lowerdir=%s,upperdir=%s,workdir=%s,userxattr", lowerDir, upperDir, workDir),
+		"none",
+		mountDir,
+	}
+
```
DrDaveD, line 182: @cclerget pointed out to me via Slack that this can be simplified.  For the "-o" option use the constant "lowerdir=/tmp:/etc" and replace "mountDir" with "/tmp".   Then remove all references to "tmpDir".

DrDaveD (changes requested):

DrDaveD (approved):

internal/pkg/util/fs/overlay/overlay_linux.go
DrDaveD: It turns out that the simplified version doesn't work at least on Ubuntu 22.04.  For some reason it gives an error
```
mount: /tmp: wrong fs type, bad option, bad superblock on none, missing codepage or helper program, or other error.
```
Replacing the first "/tmp" with "/mnt" makes it work.

PR #869, opened 2022-11-10
fix: Wait directly in fakeroot MonitorContainer, from sylabs 1116

edytuk:
This pulls in sylabs PR

- sylabs/singularity#1116
 which fixed
- sylabs/singularity#1109

> When the user's container exits, fakeroot cleanup runs `rm` though the fakeroot engine to remove a temporary container directory, extracted from a SIF image, where necessary.
> 
> Under high load / parallelism, on some distros / kernels, execution of fakeroot tempdir cleanup can become stuck. See sylabs/singularity#1109 for discussion of a reproducer.
> 
> The fakeroot engine MonitorContainer code sits in a loop waiting for signals. It requires a SIGCHLD to be received before it will wait4 on the child process, to confirm it has exited.
> 
> Debug logging showed that sometimes a SIGCHLD was not received by this code. It is not clear why, as `signal.Notify` is set early, and the signals channel buffer is not being filled. The behavior varies between distributions / kernels. It is sometimes easy to trigger, and sometimes impossible.
> 
> As a workaround, re-structure the MonitorContainer function so that it performs a blocking `wait4` in a go routine, separate from signal handling. This ensures that receiving a `SIGCHLD` is not necessary to identify that the child process has exited.
> 
> This fix has been verified in the CircleCI ssh environment. A reproducer script that previously will freeze within a few iterations does not freeze at all with this change.

### DIFF
--- a/internal/pkg/runtime/engine/fakeroot/engine_linux.go
+++ b/internal/pkg/runtime/engine/fakeroot/engine_linux.go
@@ -260,26 +260,42 @@ func (e *EngineOperations) StartProcess(masterConnFd int) error {
 // not need them for wait4 and kill syscalls.
 func (e *EngineOperations) MonitorContainer(pid int, signals chan os.Signal) (syscall.WaitStatus, error) {
 	var status syscall.WaitStatus
+	waitStatus := make(chan syscall.WaitStatus, 1)
+	waitError := make(chan error, 1)
+
+	go func() {
+		sylog.Debugf("Waiting for container process %d", pid)
+		_, err := syscall.Wait4(pid, &status, 0, nil)
+		sylog.Debugf("Wait for process %d complete with status %v, error %v", pid, status, err)
+		waitStatus <- status
+		waitError <- err
+	}()
 
 	for {
-		s := <-signals
-		switch s {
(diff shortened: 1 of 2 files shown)


internal/pkg/runtime/engine/fakeroot/engine_linux.go
```diff
@@ -260,26 +260,42 @@ func (e *EngineOperations) StartProcess(masterConnFd int) error {
 // not need them for wait4 and kill syscalls.
 func (e *EngineOperations) MonitorContainer(pid int, signals chan os.Signal) (syscall.WaitStatus, error) {
 	var status syscall.WaitStatus
+	waitStatus := make(chan syscall.WaitStatus, 1)
+	waitError := make(chan error, 1)
+
+	go func() {
+		sylog.Debugf("Waiting for container process %d", pid)
+		_, err := syscall.Wait4(pid, &status, 0, nil)
```
cclerget, line 268: so sending a SIGSTOP/SIGCONT signal to the container process will cause an exit

DrDaveD:
@edytuk please also cherry-pick to release-1.1

PR #904, opened 2022-12-06
Allow unsquashfs to be a relocatable bash script

DrDaveD:
This builds on #883 to allow the `--unsquashfs` option to be used also with relocatable binaries.  It has special treatment because unsquashfs is run as a little isolated container, and so apptainer needs to know how to search it for libraries to bind in to the container.


DrDaveD:
Ok this is updated, ready again for review.

cclerget (approved):

PR #915, opened 2022-12-09
Allow --security uid and gid in non-suid mode

### LINKED ISSUE #816: `--security uid:N` option does not work without suid install
With apptainer-1.1.2 and also singularity-3.8.6:

According to the [documentation](https://apptainer.org/docs/user/main/security_options.html#security) the root user is supposed to be able to pass an option `--security uid:N` when running as real root.  I have found that this only works with a suid install.  Otherwise these messages are printed:
```
ERROR  : Failed to write to UID map: Operation not permitted
ERROR  : Error while waiting event for user namespace mappings: Success
```
I'm guessing that it's because it is trying to set up an unprivileged user namespace instead of a privileged one.  If you just use sudo without a suid install and without a `--security` option, it does set up a privileged namespace, so I'm not sure why this is happening.

### LINKED ISSUE #848: Support unprivileged arbitrary UID/GID mapping
### Version of Apptainer

```
$ apptainer version
1.1.3
```
### Expected behavior

Now that the `--fakeroot` option can be used to map the user's UID on the host to UID 0 in the container **without privilege**, it seems like it should be possible to map host UID/GIDs to container UID/GIDs also without privilege.  @cclerget has suggested that the `--userns` and `--security` options could potentially be used in combination to effect this mapping. 

### Actual behavior

This still requires privilege in Apptainer.  

For example, I've created the user `ruser` with UID `1234` in a container called rstudio-server.sif.  I'd like to run as that user in the container, but the following produces a fatal error instead. 
```
$ apptainer shell -u --security=uid:1234 rstudio-server.sif 
FATAL:   uid security feature requires root privileges
```

```
$ cat /etc/os-release 
NAME="Fedora Linux"
VERSION="36 (Workstation Edition)"
ID=fedora
VERSION_ID=36
VERSION_CODENAME=""
PLATFORM_ID="platform:f36"
PRETTY_NAME="Fedora Linux 36 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:36"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f36/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=36
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=36
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
VARIANT_ID=workstation
```

From source tarball.

DrDaveD:
This enables the `--security uid:N` and `--security gid:N` options to work in non-suid mode.

- Fixes #816
- Fixes #848

### DIFF
--- a/cmd/starter/c/starter.c
+++ b/cmd/starter/c/starter.c
@@ -316,9 +316,11 @@ static void apply_privileges(struct privileges *privileges, struct capabilities 
                 fatalf("Failed to set GID %d: %s\n", targetGID, strerror(errno));
             }
 
-            debugf("Set %d additional group IDs\n", privileges->numGID);
-            if ( setgroups(privileges->numGID, privileges->targetGID) < 0 ) {
-                fatalf("Failed to set additional groups: %s\n", strerror(errno));
+            if ( privileges->numGID > 1 ) {
+                debugf("Set %d additional group IDs\n", privileges->numGID);
+                if ( setgroups(privileges->numGID, privileges->targetGID) < 0 ) {
+                    fatalf("Failed to set additional groups: %s\n", strerror(errno));
+                }
             }
         }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,8 +47,12 @@ For older changes see the [archived Singularity change log](https://github.com/a
   When moving the binaries to a new location, the `/usr` at the top of some
   of the paths needs to be removed.  Relocation is disallowed when the
   `starter-suid` is present, for security reasons.
-- Remove obsolete pacstrap `-d` in Arch packer
-- Adjust warning message for deprecated environment variables usage
+- Remove obsolete pacstrap `-d` in Arch packer.
+- Adjust warning message for deprecated environment variables usage.
+- Enable the `--security uid:N` and `--security gid:N` options to work
+  when run in non-suid mode.  In non-suid mode they work with any user,
+  not just root.  Unlike with root and suid mode, however, only one gid
+  may be set in non-suid mode.
 
 ## v1.1.3 - \[2022-10-25\]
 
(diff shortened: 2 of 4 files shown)


cmd/starter/c/starter.c
```diff
@@ -316,10 +316,12 @@ static void apply_privileges(struct privileges *privileges, struct capabilities
                 fatalf("Failed to set GID %d: %s\n", targetGID, strerror(errno));
             }
 
-            debugf("Set %d additional group IDs\n", privileges->numGID);
-            if ( setgroups(privileges->numGID, privileges->targetGID) < 0 ) {
-                fatalf("Failed to set additional groups: %s\n", strerror(errno));
-            }
+	    if ( privileges->numGID > 1 ) {
+		debugf("Set %d additional group IDs\n", privileges->numGID);
```
cclerget, line 320: bad indentation

cclerget (changes requested):
Except the bad indentation, lgtm

cmd/starter/c/starter.c
DrDaveD: Good catch.  I had vim set up to not expand tabs, for go.  I changed it to expand tabs to spaces for `*.[ch]` files.

GodloveD:
Looks good to me too.  Thanks @DrDaveD !

PR #937, opened 2022-12-17
Set EUID during %environment if not set

### LINKED ISSUE #837: %environment omits, allows overwriting of EUID
### Version of Apptainer

```
apptainer version 1.1.2-1.el9
```

### Expected behavior

`%environment` scriptlet should have an `EUID` variable set during execution, and should not allow it to be set / overridden.

Should output

```
EUID: readonly variable
```

### Actual behavior

`%environment` scriptlet does not have an `EUID` variable set during execution, and allows it to be set / overridden by the scriptlet.

```
%environment
EUID=$(/usr/bin/id -u)
```

```sh
$ cat /etc/os-release
NAME="Rocky Linux"
VERSION="9.0 (Blue Onyx)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="9.0"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Rocky Linux 9.0 (Blue Onyx)"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:rocky:rocky:9::baseos"
HOME_URL="https://rockylinux.org/"
BUG_REPORT_URL="https://bugs.rockylinux.org/"
ROCKY_SUPPORT_PRODUCT="Rocky-Linux-9"
ROCKY_SUPPORT_PRODUCT_VERSION="9.0"
REDHAT_SUPPORT_PRODUCT="Rocky Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.0"
```

EPEL.

anderbubble:
EUID is not defined by mvdan/sh, but some lmod expects it to be set during profile initialization. This sets EUID, when it is missing, before the user-defined %environment section is sourced.

Closes #837

Signed-off-by: Jonathon Anderson <janderson@ciq.co>

### DIFF
--- a/internal/pkg/build/sources/01-base.sh
+++ b/internal/pkg/build/sources/01-base.sh
@@ -22,3 +22,6 @@
 # to reproduce, distribute copies to the public, prepare derivative works, and
 # perform publicly and display publicly, and to permit other to do so.
 #
+if [ -z "$EUID" ]; then
+    EUID="$UID"
+fi


anderbubble:
This PR mostly fixes the reported issue, but only for newly-built containers. Really, the right thing would be to fix the mvdan/sh upstream to include EUID; but, barring that, I couldn't figure out how to set a _shell_ variable (rather than an environment variable) from outside of the interpreter.

Let me know if you think I should submit the change upstream in stead, and then wait to link that in here.

anderbubble:
I don't think the shellcheck warning should matter, since if $UID is undefined, it will just set EUID to the same undefined value.

anderbubble:
I went ahead and submitted the upstream PR. Personally, I think that's a better fix than this.

https://github.com/mvdan/sh/pull/958

DrDaveD:
The shellcheck would need to be made to pass; you can put in 
```
# shellcheck disable=SC3028
```
on the preceding line.  But since the mvdan PR is progressing, it's a moot point.  Please close this when you're ready.

anderbubble:
Closing in favor of traction at https://github.com/mvdan/sh/pull/958

PR #954, opened 2022-12-21
Unprivileged install for openSUSE/SUSE

mslacken:
Adds openSUSE/SUSE distros to `tools/install-unprivileged.sh`.
Installation can be done via KOJI rpms or the rpms from the openBuild service.

### DIFF
--- a/tools/install-unprivileged.sh
+++ b/tools/install-unprivileged.sh
@@ -276,7 +276,7 @@ OSUTILS=""
 EXTRASUTILS="fuse-overlayfs"
 EPELUTILS="fakeroot"
 if [ "$DIST" == el7 ]; then
-	OSUTILS="$OSUTILS lzo squashfs-tools libseccomp fuse-libs"
+	OSUTILS="$OSUTILS lzo libseccomp squashfs-tools fuse-libs"
 	EPELUTILS="$EPELUTILS libzstd fuse2fs fuse3-libs fakeroot-libs"
 elif [ "$DIST" == el8 ]; then
 	OSUTILS="$OSUTILS lzo libseccomp squashfs-tools fuse-libs libzstd e2fsprogs-libs e2fsprogs fuse3-libs"
@@ -290,7 +290,7 @@ elif [ "$DIST" == "suse15" ]; then
 	EXTRASUTILS="$EXTRASUTILS fuse2fs"
 	EPELUTILS=""
 else
-	OSUTILS="$OSUTILS lzo libseccomp squashfs-tools libzstd e2fsprogs-libs e2fsprogs"
+	OSUTILS="$OSUTILS lzo libseccomp squashfs-tools fuse-libs libzstd e2fsprogs-libs e2fsprogs"
 	EXTRASUTILS="$EXTRASUTILS fuse3-libs"
 	EPELUTILS="$EPELUTILS fakeroot-libs"
 fi
--- a/tools/install-unprivileged.sh
+++ b/tools/install-unprivileged.sh
@@ -23,14 +23,12 @@ usage()
 	echo "   e.g. el9 or fc37, default based on the current system."
 	echo "   As a convenience, active debian and ubuntu versions (not counting 18.04)"
 	echo "   get translated into a compatible el version for downloads."
-	echo "   Also openSUSE based distributions are mapped to EL/FC or native openSUSE"
+	echo "   OpenSUSE based distributions are also mapped to el, or native openSUSE"
 	echo "   binaries can be used via the -o switch."
 	echo " arch can be any architecture supported by EPEL or Fedora,"
 	echo "   default based on the current system."
 	echo " version selects a specific apptainer version, default latest release,"
 	echo "   although if it ends in '.rpm' then apptainer will come from there."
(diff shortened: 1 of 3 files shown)


DrDaveD:
Thanks!  In general this looks good.  Please fix the shellcheck ci test.  If you think the message is invalid you can always disable it with a `# shellcheck disable=` comment.  Also please re-do the commits with `git -s` to make the DCO happy.  The rocky9 failure is a known temporary failure that should be fixed soon.

Add to the usage message the fact that current suse distributions are also supported.  For the "-o" option please make it clear that it only applies to opensuse distributions.

mslacken:
Added the requested lines of explanation.

tools/install-unprivileged.sh
```diff
@@ -156,7 +196,7 @@ LASTPKGS=""
 latesturl()
 {
 	typeset URL="$1"
-	if [ "$3" = true ]; then
+	if [ "$3" == true ]; then
```
DrDaveD, line 199: If you make this kind of change please make it consistently.  I see for example below `[ "$4" = true ]`.

tools/install-unprivileged.sh
```diff
@@ -179,11 +219,11 @@ latesturl()
 	fi
 }
 
-APPTAINERURL="https://kojipkgs.fedoraproject.org/packages/apptainer/"
 LOCALAPPTAINER=false
-if [ -z "$VERSION" ]; then
+#if test -z "$VERSION" || ! test -z $OBSURL ; then
+if test -z "$VERSION" || [[ "$KOJI" == "false" ]]; then
```
DrDaveD, line 224: Let's not leave this comment, and don't make the style inconsistent by changing `[ -z` to `test`.

tools/install-unprivileged.sh
```diff
@@ -284,7 +337,12 @@ cd ..
 
 set -e
 
+# remove .build-id files seen on el8 or later
+rm -rf lib/.build-id
+rmdir lib 2>/dev/null || true
```
DrDaveD, line 342: Since this is my change, I think you should instead rebase on the current main and force-push.

DrDaveD (changes requested):
Please address the new failure on ubuntu in the rocky9 CI test.

And that reminds me: please add additional invocations of TEST_TYPE=unpriv in the centos7, rocky8, and rocky9 tests for the different SUSE distributions that have dockerhub containers available.

mslacken:
Checks pass, but I had to remove tumbleweed as rolling distributions show sometimes strange errors.
I will check this more closely next year

DrDaveD (changes requested):
If I test this with the "-o" option (on an opensuse/leap docker container) I get
```
Extracting https://download.opensuse.org/repositories/network:/cluster/openSUSE_Leap_15.4/x86_64/apptainer-1.1.2-lp154.30.17.x86_64.rpm
argument is not an RPM package
cpio: premature end of archive
Error extracting https://download.opensuse.org/repositories/network:/cluster/openSUSE_Leap_15.4/x86_64/apptainer-1.1.2-lp154.30.17.x86_64.rpm
```
I expect that since it isn't 1.1.4 that the resultant install wouldn't work, but that URL gets a 404 Not Found error and that's not right.

DrDaveD:
@mslacken Reminder, this is waiting on you.

mslacken:
Sorry for the late reply, but I was on holidays.
Basically always only *one* apptainer version is available with the open build service as older versions get deleted as soon as a new version is accepted. 
I have updated the package in the devel repo (https://build.opensuse.org/repositories/network:/cluster)  so that only 1.1.4 is now available.
I think this solves the problem.

DrDaveD:
That still doesn't quite do it.  I still get the same message.  I see that https://download.opensuse.org/repositories/network:/cluster/openSUSE_Leap_15.4/x86_64, the one being used, still has apptainer-1.1.2, but that apptainer-1.1.4 is in https://mirrorcache-us.opensuse.org/repositories/network:/cluster/15.4/x86_64/.  Does the URL need to change?  Oh I see it has already been changed in $OSREPOURL but it hasn't been changed in $EPELREPOURL.  Yes, if I make that change, it works.

I want to also make a few style changes, so I'll go ahead and push a fix to the PR branch.

DrDaveD:
Oddly, if you look at https://download.opensuse.org/repositories/network:/cluster there's no "15.4" shown, but yet if you just add it to the URL it gets you to the more up to date repository.  Something is strange there.

DrDaveD:
> Oddly, if you look at https://download.opensuse.org/repositories/network:/cluster there's no "15.4" shown, but yet if you just add it to the URL it gets you to the more up to date repository. Something is strange there.

Actually, it seems that sometimes it redirects me to https://mirrorcache-us.opensuse.org/repositories/network:/cluster and it is only then that I don't see the "15.4".  I don't understand why sometimes I get redirected and sometimes I don't.

DrDaveD (approved):

DrDaveD:
@mslacken This looks good to me now.  I will wait to merge until you tell me what you think about the additional changes I made.

mslacken:
Thanks for your additional work.

PR #967, opened 2022-12-27
Do not relocate if the executable base is unknown

### LINKED ISSUE #958: Regression in 1.1.4: cannot run in htcondor with input file transfer
### Version of Apptainer

```apptainer version 1.1.4-1.el7```

### Expected behavior

It should start without errors

### Actual behavior
```
FATAL:   While initializing: couldn't parse configuration file etc/apptainer/apptainer.conf: open etc/apptainer/apptainer.conf: no such file or directory

```

This is started from a HTCondor job that was working fine for years. A minimal condor submit file:
```
output=folding.$(ClusterId).$(ProcId).out
error=folding.$(ClusterId).$(ProcId).err
should_transfer_files=YES
transfer_input_files=test.sh
executable=/usr/bin/apptainer
queue
```
It is working fine if ```should_transfer_files=NO``` is set. 

The difference between the runs with transfer disabled vs. enabled is the initial job directory.

It is also working fine if apptainer isn't started directly but from a minimal script file test.sh:
```
#!/bin/bash
/usr/bin/apptainer
```

```sh
$ cat /etc/os-release
NAME="Scientific Linux"
VERSION="7.9 (Nitrogen)"
ID="scientific"
ID_LIKE="rhel centos fedora"
VERSION_ID="7.9"
PRETTY_NAME="Scientific Linux 7.9 (Nitrogen)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:scientificlinux:scientificlinux:7.9:GA"
HOME_URL="http://www.scientificlinux.org//"
BUG_REPORT_URL="mailto:scientific-linux-devel@listserv.fnal.gov"

REDHAT_BUGZILLA_PRODUCT="Scientific Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.9
REDHAT_SUPPORT_PRODUCT="Scientific Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.9"

```

EPEL rpm package

DrDaveD:
Starting in version 1.1.4, if apptainer is executed under an unrecognized base executable it mistakenly thinks the prefix is the empty string.  This makes it fall back to the default compiled-in prefix instead.

- Fixes #958


internal/pkg/buildcfg/confgen/gen.go
```diff
@@ -88,12 +87,12 @@ var (
...
 
-func getPrefix() (string, error) {
-	var err error
+func getPrefix() (string) {
 	prefixOnce.Do(func() {
-		var executablePath string
-		executablePath, err = os.Executable()
+		executablePath, err := os.Executable()
 		if err != nil {
+			sylog.Debugf("Error getting executable path, using default: %v", err)
```
cclerget, line 94: a warning would be more appropriate

cclerget (changes requested):

cclerget (approved):

PR #1042, opened 2023-01-30
Update basechecks.chk

simonleary-umass-edu:

`echo "not found!"` assumes that `command -v $go` came up empty. But if `$hstgo_opts` is undefined, `command -v $go` was never run, and no check has actually been conducted.

I think someone with more understanding should come up with a more descriptive error message. I've never used the go compiler, and I'm not experienced with autotools, but I'm confused by a couple things here:
* what's the difference between the 
* why `hstgo_opts` would ever be anything other than `"go"`
* why there should be a list of names that invoke the go compiler
* why a list of names that invoke the go compiler should be called `hstgo_opts`.

### DIFF
--- a/mlocal/checks/basechecks.chk
+++ b/mlocal/checks/basechecks.chk
@@ -79,6 +79,9 @@ if [ $do_go_version_check = 0 ]; then
 	fi
 elif [ "$hstgo" = "" ]; then
 	printf " checking: host Go compiler (at least version $hstgo_version)... "
+	if [ "$hstgo_opts" = "" ]; then
+		echo "\$hstgo_opts is undefined! Cannot check host Go compiler."
+	fi
 	for go in $hstgo_opts; do
 		if $makeit_scriptsdir/check-min-go-version $go $hstgo_version >/dev/null 2>&1; then
 			hstgo=`command -v $go`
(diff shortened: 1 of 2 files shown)


DrDaveD:
What was the higher level problem that you were trying to solve with this?

simonleary-umass-edu:
I was getting
```
 checking: host Go compiler (at least version 1.18)... not found!
```
and after finding [this issue](https://github.com/hpcng/singularity/issues/5099), I was convinced that there was a bug in the configure script. Then I ran it with `set -x` and the output didn't have any indication of it running the for loop. I assumed because it skipped the loop that `hstgo_opts` must be undefined.
```
	for go in $hstgo_opts; do
		if $makeit_scriptsdir/check-min-go-version $go $hstgo_version >/dev/null 2>&1; then
			hstgo=`command -v $go`
			break
		fi
	done
```
In reality, it did go through the for loop, and `check-min-go-version` came back false.

So this `hstgo_opts` undefined scenario is not one that I encountered. But if there was a misconfiguration of `mconfig` this situation would happen and the error would be misleading.

In the end, it turns out that my go just didn't meet the required version. That should have been the first thing that I checked. [I made another PR](https://github.com/apptainer/apptainer/pull/1045) which says "hey, we found go but it's not good enough" rather than "we didn't find go".

PR #1045, opened 2023-01-30
Go check descriptive error message

simonleary-umass-edu:
before:
```
 checking: host Go compiler (at least version 1.18)... not found!
```
after:
```
 checking: host Go compiler (at least version 1.18)... found go 1.17.2 does not meet minimum requirement 1.18!
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,6 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 ## Changes Since Last Release
 
-### Bug fixes
-
-- Fix `GOCACHE` environment variable settings when building debian source
-  package on PPA build environment.
-- Make `PS1` environment variable changeable via `%environment` section on
-  definition file that used to be only changeable via `APPTAINERENV_PS1`
-  outside of container. This makes container's prompt customizable.
-
 ### Changed defaults / behaviours
 
 - When the kernel supports unprivileged overlay mounts in a user
@@ -39,6 +31,14 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Restored the ability for running instances to be tracked when apptainer
   is installed with tools/install-unprivileged.sh.  Instance tracking
   depends on argument 0 of the starter, which was not getting preserved.
+- Fix `GOCACHE` environment variable settings when building debian source
+  package on PPA build environment.
+- Make `PS1` environment variable changeable via `%environment` section on
+  definition file that used to be only changeable via `APPTAINERENV_PS1`
+  outside of container. This makes container's prompt customizable.
+- Fix the passing of nested bind mounts when there are multiple binds
+  separated by commas and some of them have colons separating sources
+  and destinations.
 
 ## v1.1.5 - \[2023-01-10\]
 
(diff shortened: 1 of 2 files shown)


mlocal/checks/basechecks.chk
```diff
@@ -79,6 +79,10 @@ if [ $do_go_version_check = 0 ]; then
 	fi
 elif [ "$hstgo" = "" ]; then
 	printf " checking: host Go compiler (at least version $hstgo_version)... "
+	if [ "$hstgo_opts" = "" ]; then
+		echo "\$hstgo_opts is undefined! Cannot check host Go compiler."
+		exit 1
+	fi
```
DrDaveD, line 85: Rather than adding this, let's delete hstgo_opts from mconfig and just use "go" instead of the for loop and $go below.  It's unnecessarily obscure.

DrDaveD (changes requested):

DrDaveD:
Please squash down to one commit when you are finished, and do the commit with `git -s` to satisfy the DCO check.  See CONTRIBUTING.md if you don't know what that means, and add yourself to CONTRIBUTORS.md if you're not in there yet.

simonleary-umass-edu:
I'm not super savvy with git, I tried to remove the first commit with `rebase` but it's still here with a red X. I can make the proposed changes and do the same `rebase` and there will be 2 red X's. Is that ok? Is there a better way?

DrDaveD:
I usually do `git reset HEAD^1` to remove one commit at a time to bring it back to the beginning.  Then after commit do `git push -f` to force-push.

mlocal/scripts/check-min-go-version
```diff
@@ -27,19 +30,26 @@ MINVERSION="$2"
 MINMAJOR="${MINVERSION/.*}"
 MINMINOR="${MINVERSION#*.}"
 MINPATCH="${MINMINOR#*.}"
+
+insufficient_go_version(){
+    echo "found go $GOVERSION does not meet minimum requirement $MINVERSION!"
```
DrDaveD, line 35: This is always followed by an `exit 1` so I would put that in the function.

mlocal/checks/basechecks.chk
```diff
@@ -62,13 +62,11 @@ fi
...
-			if [ "$gopth" != "" ]; then
-				hstgo=$gopth
-				break
-			fi
-		done
+        gopth=`command -v $go`
+        if [ "$gopth" != "" ]; then
+            hstgo=$gopth
+            break
+        fi
```
DrDaveD, line 69: This has an indentation problem.  This script is using all tabs the way go requires it, so please make sure your editor isn't putting in spaces instead, so it stays consistent.

mlocal/checks/basechecks.chk
```diff
@@ -79,22 +77,19 @@ if [ $do_go_version_check = 0 ]; then
...
-	for go in $hstgo_opts; do
-		if $makeit_scriptsdir/check-min-go-version $go $hstgo_version >/dev/null 2>&1; then
-			hstgo=`command -v $go`
-			break
-		fi
-	done
+    if $makeit_scriptsdir/check-min-go-version go $hstgo_version; then
+        hstgo=`command -v $go`
+        break
+    fi
```
DrDaveD, line 83: Likewise, check the indentation.

mlocal/scripts/check-min-go-version
DrDaveD: This script is using 4 spaces for indent, so the indenting is fine here.

DrDaveD (changes requested):

simonleary-umass-edu:
I think I pulled other people's changes into my commit somehow?

PR #914, opened 2022-12-09
Fix strange variable name APPTAINERENV_SING_USER_DEFINED_PREPEND_PATH in deprecation warning

JasonYangShadow:
Signed-off-by: jason yang <jasonyangshadow@gmail.com>

## Description of the Pull Request (PR):

Adjust warning message for deprecated environment variables usage

### This fixes or addresses the following GitHub issues:

 - Fixes #870 

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/apptainer/apptainer/blob/main/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTORS.md)

PR #1050, opened 2023-02-01
Hide messages about SINGULARITY variables if corresponding APPTAINER …

### LINKED ISSUE #1040: Hide messages about SINGULARITY variables if corresponding APPTAINER variables are defined
### Version of Apptainer

```
apptainer --version
apptainer version 1.1.5-1.el8
```

### Expected behavior

No message about SINGULARITY variables if corresponding APPTAINER variables are defined.

### Actual behavior

I received this message:
```
INFO:    Environment variable SINGULARITYENV_PS1 is set, but APPTAINERENV_PS1 is preferred
```

Run `apptainer` with both variables defined.

```sh
$ cat /etc/os-release
NAME="AlmaLinux"
VERSION="8.7 (Stone Smilodon)"
ID="almalinux"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
PRETTY_NAME="AlmaLinux 8.7 (Stone Smilodon)"
ANSI_COLOR="0;34"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:almalinux:almalinux:8::baseos"
HOME_URL="https://almalinux.org/"
DOCUMENTATION_URL="https://wiki.almalinux.org/"
BUG_REPORT_URL="https://bugs.almalinux.org/"

ALMALINUX_MANTISBT_PROJECT="AlmaLinux-8"
ALMALINUX_MANTISBT_PROJECT_VERSION="8.7"
REDHAT_SUPPORT_PRODUCT="AlmaLinux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.7"
```

RPM

JasonYangShadow:
…variables are defined

Signed-off-by: jason yang <jasonyangshadow@gmail.com>

Hide messages about SINGULARITY variables if corresponding APPTAINER variables are defined

 - Fixes #1040 


internal/pkg/util/env/create.go
```diff
@@ -103,7 +103,16 @@ func warnDeprecatedEnvUsage(hostEnvs []string) {
 				if key != "" {
 					legacyEnv := LegacySingularityEnvPrefix + key
 					newEnv := ApptainerEnvPrefix + key
-					sylog.Infof("Environment variable %v is set, but %v is preferred", legacyEnv, newEnv)
+					skip := false
+					for _, e := range hostEnvs {
+						if strings.HasPrefix(e, newEnv) {
```
DrDaveD, line 108: Why isn't this just a check for equal?  Wouldn't checking the prefix make it get matches for variables names that aren't the same but start with the same value?

CHANGELOG.md
```diff
@@ -39,6 +39,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Fix the passing of nested bind mounts when there are multiple binds
   separated by commas and some of them have colons separating sources
   and destinations.
+- Hide messages about SINGULARITY variables if corresponding APPTAINER
+  variables are defined
```
DrDaveD, line 43: Please add an extra sentence "Fixes a regression introduced in 1.1.4.".

DrDaveD (changes requested):
I think it's unfortunate that this requires a linear search for each matching variable.  However, it is easier to understand than the `precedence` logic that was removed in #914 and I don't suppose that it will use a significant amount of CPU.  On the other hand if someone has a lot of variables it might.  I wonder if @cclerget has an opinion.

DrDaveD:
The test failures are due to a known problem with dockerhub ubuntu images.  It's supposed to get fixed today.

internal/pkg/util/env/create.go
```diff
@@ -100,10 +101,21 @@ func warnDeprecatedEnvUsage(hostEnvs []string) {
 			strs := strings.SplitN(env, "=", 2)
 			if len(strs) == 2 {
 				key := strs[0][len(LegacySingularityEnvPrefix):]
+				value := strs[1]
 				if key != "" {
 					legacyEnv := LegacySingularityEnvPrefix + key
 					newEnv := ApptainerEnvPrefix + key
-					sylog.Infof("Environment variable %v is set, but %v is preferred", legacyEnv, newEnv)
+					skip := false
+					targetValue := fmt.Sprintf("%s=%s", newEnv, value)
```
DrDaveD, line 109: Oh now I see why it was checking the prefix, because it includes also the value.  We used to have a warning if the variables were both set to different values, because that can also be a problem for people.  The warning said the settings were different, but we were using the APPTAINER variable.  Check elsewhere for the exact wording.  So the thing to do is to check prefix, but with an `=` sign appended, and then check to see if the values are the same or different to decide whether to print the warning or info message.

DrDaveD (changes requested):

JasonYangShadow:
> I think it's unfortunate that this requires a linear search for each matching variable. However, it is easier to understand than the `precedence` logic that was removed in #914 and I don't suppose that it will use a significant amount of CPU. On the other hand if someone has a lot of variables it might. I wonder if @cclerget has an opinion.

Let me use a hashset for those host env vars. it'll reduce the each search cost to O(1).

internal/pkg/util/env/create.go
JasonYangShadow: https://github.com/JasonYangShadow/apptainer/blob/332bed37c1809c7464956eda95348a7c75625468/internal/pkg/util/env/env.go#L78

internal/pkg/util/env/create.go
```diff
@@ -95,15 +95,29 @@ func overridesForContainerEnv(g *generate.Generator, hostEnvs []string) envKeyMa
 
 // warning if deprecated keys are set
 func warnDeprecatedEnvUsage(hostEnvs []string) {
+	envMap := make(map[string]string)
+	for _, env := range hostEnvs {
+		strs := strings.SplitN(env, "=", 2)
+		if len(strs) == 2 {
+			envMap[strs[0]+"="] = strs[1]
```
DrDaveD, line 102: This overall looks pretty good to me, except I don't see the added value of including the `+"="` here and below in the lookup.

DrDaveD (changes requested):

internal/pkg/util/env/create.go
JasonYangShadow: thanks, it also looks like that we do not need the "=" as the part of the key

DrDaveD:
@JasonYangShadow Please rebase on the current main for the CHANGELOG conflict.

PR #1071, opened 2023-02-08
instance: start with cgroup if possible, to enable stats, from sylabs 1186

edytuk:
This pulls in sylabs PR

- sylabs/singularity#1186
 which fixed
- sylabs/singularity#891

> Start an instance with a cgroup by default, where the calling user is root, or cgroups v2 unified mode with systemd as manager is available.
> 
> This allows `singularity instance stats` to work by default in these circumstances.
> 
> The PR also includes a fix for a limitation of non-group instances with cgroups that is more broadly exposed by the above change:
> 
> When joining a non-root instance using cgroups, we need to move our instance joining process into the instance cgroup.
> 
> Previously we did this directly. However, it is common for a CLI instance join to be started under a session scope cgroup that is root owned, and not delegated. We cannot move the CLI process into the instance cgroup directly from here, due to the requirements detailed in:
> 
> https://www.kernel.org/doc/html/v4.18/admin-guide/cgroup-v2.html#delegation-containment
> 
> Instead, create a new systemd managed cgroup, and move into this first. The new cgroup will be a sibling of the instance cgroup, in the systemd managed delegated tree. We can then move again, this time into the instance cgroup we need to be in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ For older changes see the [archived Singularity change log](https://github.com/a
   functionality.
 - Overlay is blocked on the `panfs` filesystem, allowing sandbox directories to be
   run from `panfs` without error.
+- Instances are started in a cgroup, by default, when run as root or when
+  unified cgroups v2 with systemd as manager is configured. This allows
+  `apptainer instance stats` to be supported by default when possible.
 
 ### New features / functionalities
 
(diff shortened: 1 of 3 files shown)


cclerget (approved):

DrDaveD:
@edytuk please fix CHANGELOG conflict.

PR #1077, opened 2023-02-09
fix: block overlay for panfs, from sylabs 1219

edytuk:
This pulls in sylabs part of PR

- sylabs/singularity#1219
 which fixed
- sylabs/singularity#1215

> A sandbox container on a panfs (Panasas) filesystem causes an overlay error at runtime. Add panfs to to the list of incompatible filesystems.
> 
> Block panfs as both lower and upper overlay dirs, until support is confirmed with Panasas (same situation as GPFS).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - A new `--reproducible` flag for `./mconfig` will configure Apptainer so that
   its binaries do not contain non-reproducible paths. This disables plugin
   functionality.
+- Overlay is blocked on the `panfs` filesystem, allowing sandbox directories to be
+  run from `panfs` without error.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -34,6 +34,11 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 ### Bug fixes
 
+- Overlay is blocked on the `panfs` filesystem, allowing sandbox directories to be
+  run from `panfs` without error.
+
```
DrDaveD, line 39: I previously asked you to put this here, but I changed my mind and would like you to instead move this message up under the "Changed defaults/behaviors" section.

DrDaveD (changes requested):

DrDaveD (approved):

PR #1092, opened 2023-02-10
Show info message instead of fatal error for name collision between global and individual remote

### LINKED ISSUE #1018: Adding locally a global remote is allowed but causes "remote" commands not to work any more
### Version of Apptainer

This problem happened to a student during a tutorial, using an installation on CVMFS instead of the system installed apptainer. Not sure if I should report this here or the version is obsolete.
`singularity version 3.8.6`

### Expected behavior

A warning during the addition of the remote if it is already present and would cause name collisions

### Actual behavior

The addition was OK, the problem started when using remote commands after that:
```
» singularity remote add --no-login SylabsCloud cloud.sycloud.io
INFO:    Remote "SylabsCloud" added.
» singularity remote use SylabsCloud
FATAL:   name collision while syncing: SylabsCloud
» singularity remote list                                                                                                                                   
FATAL:   name collision while syncing: SylabsCloud
```

1. Add a global remote as root.
2. Add the same remote locally (as regular user)
3. Try to use `singularity remote` commands

```sh
$ cat /etc/os-release
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="http://cern.ch/linux/"
BUG_REPORT_URL="http://cern.ch/linux/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```

It is a CVMFS installation on `lxplus.cern.ch` :
```
$ which singularity
/cvmfs/lhcb.cern.ch/lib/var/lib/LbEnv/2683/stable/linux-64/bin/singularity
```

DrDaveD:
When the same remote name is defined both globally and individually, show an info message instead of causing a fatal error.

- Fixes #1018

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,10 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Increase the timeout on image driver (that is, FUSE) mounts from 2
   seconds to 10 seconds.  Instead, print an INFO message if it takes
   more than 2 seconds.
+- If a `remote` is defined both globally (i.e. system-wide) and
+  individually, change `apptainer remote` commands to print an info message
+  instead of exiting with a fatal error and to give precedence to the
+  individual configuration.
 
 ## v1.1.5 - \[2023-01-10\]
 
--- a/internal/pkg/remote/remote.go
+++ b/internal/pkg/remote/remote.go
@@ -104,13 +104,12 @@ func (c *Config) WriteTo(w io.Writer) (int64, error) {
 
 // SyncFrom updates c with the remotes specified in sys. Typically, this is used
 // to sync a globally-configured remote.Config into a user-specific remote.Config.
-// Currently, SyncFrom will return a name-collision error if there is an EndPoint
-// name which exists in both c & sys, and the EndPoint in c has System == false.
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -73,6 +73,9 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Check for the existence of the runtime executable prefix, to avoid
   issues when running under Slurm's srun. If it doesn't exist, fall
   back to the compile-time prefix.
+- If a `remote` is defined both globally (i.e. system-wide) and
+  individually, change `apptainer remote` commands to print an info message
+  instead of exiting with a fatal error.
```
ikaneshiro, line 78: Lets include that it uses the individuals configuration over the global one in this as well

ikaneshiro (approved):

PR #1142, opened 2023-02-24
SUSE build update

mslacken:
## Simplify SLE builds

Add the possibility to build SUSE containers via suseconnect-container, what 
removed the need to register the container via SUSEConnect.
This needs the possibility to have bind mounts in the post stage of the build,
but this is not exposed in the def file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,8 +47,6 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - In `--rocm` mode, the whole of `/dev/dri` is now bound into the container when
   `--contain` is in use. This makes `/dev/dri/render` devices available,
   required for later ROCm versions.
-- Build via zypper on SLE systems will use repositories of host via
-  suseconnect-container
 
 ### Bug fixes
 
@@ -56,6 +54,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Ensure `DOCKER_HOST` is honored in non-build flows.
 - Make the error message more helpful in another place where a remote is found
   to have no library client.
+- Build via zypper on SLE systems will use repositories of host via
+  suseconnect-container
 
 ## v1.1.6 - \[2023-02-14\]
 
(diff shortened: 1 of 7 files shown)


internal/pkg/build/sources/conveyorPacker_zypper.go
```diff
@@ -469,3 +504,44 @@ func rpmPathCheck() (err error) {
 
 	return nil
 }
+
+/*
+Parse the /etc/os.release file to a a struct, so that SUSE versions
+need not to be set on a SLE syste,
+*/
```
DrDaveD, line 511: Please use `//` style comments to match the rest of the code base.

Also there's a duplicate `a` in the first line.

DrDaveD (changes requested):
In general this looks good to me, thanks.  Minor changes to a comment requested.  Maybe @cclerget can also take a look.

I also prefer to see every related thing in a single commit instead of multiple commits.

Do you think this should go into 1.1.7, or can it wait to 1.2.0?

mslacken:
It would be nice if it can go to the 1.1.7 release. I also had to force push to get it into a single commit.

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - In `--rocm` mode, the whole of `/dev/dri` is now bound into the container when
   `--contain` is in use. This makes `/dev/dri/render` devices available,
   required for later ROCm versions.
+- Build via zypper on SLE systems will use repositories of host via
+  suseconnect-container
```
DrDaveD, line 51: Since this will be ported to 1.1.7, please move the CHANGELOG entry to the Bug fixes section, even if it isn't really a bug fix.  Those are the ones we backport.

DrDaveD (changes requested):
Needs another force-push with a git commit -s for the DCO.

mslacken:
signed and signed off

DrDaveD (approved):

DrDaveD:
I force-pushed after rebasing on main to get the latest CHANGELOG, and added a commit to move the log message to the Bug fixes section.  Please make another PR cherry-picking this to the release-1.1 branch.

mslacken:
@DrDaveD Thanks for reminding, see #1226

PR #1165, opened 2023-03-07
add retry feature to unprivileged install script

### LINKED ISSUE #1164: Actions > ci > rpmbuild-rocky8 sometimes fails to retrieve package URI from mirror
### Version of Apptainer

Failure on GitHub Actions

### Expected behavior

Successfully retrieve package URI from mirror

### Actual behavior

Sometimes fails to retrieve package URI from mirror and that makes Actions > ci > rpmbuild-rocky8 job as failure.

Running Actions > ci > rpmbuild-rocky8

panda1100:

Add retry feature to `install-unprivileged.sh` to prevent
Actions > ci > rpmbuild-rocky8 failure due to fails to retrieve package URI from mirror.

 - Fixes #1164 

### DIFF
--- a/tools/install-unprivileged.sh
+++ b/tools/install-unprivileged.sh
@@ -188,8 +188,11 @@ esac
 # $4 -- if true, try replacing "/updates/" with "/releases/" if nothing found
 # If return value 0, succeeded and stdout contains latest url
 # If return value not zero, failed and stdout contains final directory url
+# If a package is not found, the listing will be silently retried up to 3 times,
+# because sometimes not all mirrors are up to date
 LASTURL=""
 LASTPKGS=""
+RETRY=0
 latesturl()
 {
 	typeset URL="$1"
@@ -201,16 +204,25 @@ latesturl()
 		# optimization: re-use last list if it hasn't changed
 		LASTURL="$URL"
 		LASTPKGS="$(curl -Ls "$URL")"
+	elif [ $RETRY -gt 0 ]; then
+		LASTPKGS="$(curl -Ls "$URL")"
 	fi
 	typeset LATEST="$(echo "$LASTPKGS"|sed -e 's/.*href="//;s/".*//' -e 's/\.mirrorlist//' -e 's/\-32bit//' -e 's@^\.\/@@' |grep "^$2-[0-9].*$ARCH"|tail -1)"
 	if [ -n "$LATEST" ]; then
+		RETRY=0
 		echo "$URL/$LATEST"
 	elif [ "$4" = true ]; then
+		RETRY=0
 		URL="${URL/\/updates\///releases/}"
 		URL="${URL/\/Packages\///os/Packages/}"
 		latesturl "$URL" "$2" false false
 		return $?
+	elif [ $RETRY -lt 3 ]; then
+		RETRY=$((RETRY+1))
+		latesturl "$URL" "$2" false "$4"
+		return $?
 	else
+		RETRY=0
 		echo "$URL"
 		return 1
 	fi


tools/install-unprivileged.sh
```diff
@@ -202,15 +203,24 @@ latesturl()
 		LASTURL="$URL"
 		LASTPKGS="$(curl -Ls "$URL")"
 	fi
+	if [ $RETRY -gt 0 ]; then
+		LASTPKGS="$(curl -Ls "$URL")"
+	fi
```
DrDaveD, line 208: Why try it right away after the previous one just above?  Maybe this should be in elif instead of an if?

tools/install-unprivileged.sh
```diff
@@ -202,15 +203,24 @@ latesturl()
...
 	if [ -n "$LATEST" ]; then
+		RETRY=0
 		echo "$URL/$LATEST"
 	elif [ "$4" = true ]; then
+		RETRY=0
 		URL="${URL/\/updates\///releases/}"
 		URL="${URL/\/Packages\///os/Packages/}"
 		latesturl "$URL" "$2" false false
 		return $?
+	elif [ $RETRY -lt 5 ]; then
```
DrDaveD, line 219: I think retrying 5 times is excessive.  Let's do 3.

tools/install-unprivileged.sh
```diff
@@ -202,15 +203,24 @@ latesturl()
...
+        RETRY=$((RETRY+1))
+        latesturl "$URL" "$2" "$3" "$4"
```
DrDaveD, line 221: These two lines should be indented with an extra tab.

DrDaveD: Add a `return $?` just to make sure future code changes don't let it continue to do something else further down, just like is done in the preceding elif.

DrDaveD (changes requested):
I note that this only retries the step for locating a package, not for downloading it.  But maybe that's OK because we have only observed errors in the locating step.

tools/install-unprivileged.sh
panda1100: This condition is for when `$1` changed otherwise skip to grab html page for scraping  package location.

```
if [ "$URL" != "$LASTURL" ]; then
		# optimization: re-use last list if it hasn't changed
		LASTURL="$URL"
		LASTPKGS="$(curl -Ls "$URL")"
fi
```

when `$1` changed that means always `$RETRY` is 0, when `$1` NOT changed, we need to check retry

```
if [ $RETRY -gt 0 ]; then
		LASTPKGS="$(curl -Ls "$URL")"
fi
```

Yes, this should be elif like the following:

```
if [ "$URL" != "$LASTURL" ]; then
		# optimization: re-use last list if it hasn't changed
		LASTURL="$URL"
		LASTPKGS="$(curl -Ls "$URL")"
elif [ $RETRY -gt 0 ]; then
		LASTPKGS="$(curl -Ls "$URL")"

fi
```

tools/install-unprivileged.sh
```diff
@@ -190,6 +190,7 @@ esac
 # If return value not zero, failed and stdout contains final directory url
 LASTURL=""
```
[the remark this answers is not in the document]
panda1100: Thank you! I added comment.

tools/install-unprivileged.sh
panda1100: Thank you! I added `return $?`.

tools/install-unprivileged.sh
panda1100: Thank you! I changed it to 3.

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 ### Bug fixes
 
+- Add retry feature for locating a package URI on install-unprivileged.sh
```
DrDaveD, line 53: Actually since install-unprivileged.sh is always run from the main branch, it doesn't really make sense to have a CHANGELOG entry for it, which gets associated with a release.  So please remove this  new entry you added.  For the same reason there's no need to backport it to the release-1.1 branch.

DrDaveD (changes requested):

CHANGELOG.md
panda1100: Thank you @DrDaveD! I understand.

PR #1185, opened 2023-03-09
Add gocryptfs support for apptainer

### LINKED ISSUE #484: Support unprivileged encrypted SIF files
One of the last remaining significant features that can't be run without suid and with unprivileged user namespaces is encrypted SIF files.  That could probably be supported with something like [gocryptfs](https://nuetzlich.net/gocryptfs/comparison/).

JasonYangShadow:

support unprivileged encrypted sif creation and execution 

 - Fixes #484 

#### Test
https://apptainer.org/docs/user/main/encryption.html

encryption
```
vagrant@bullseye:~/test$ apptainer build --pem-path rsa_pub.pem test.sif docker://alpine:latest
INFO:    Starting build...
Getting image source signatures
Copying blob 63b65145d645 skipped: already exists  
Copying config 6a2bcc1c7b done  
Writing manifest to image destination
Storing signatures
2023/03/09 11:19:55  info unpack layer: sha256:63b65145d645c1250c391b2d16ebe53b3747c295ca8ba2fcb6b0cf064a4dc21c
INFO:    Creating SIF file...
INFO:    Build complete: test.sif
vagrant@bullseye:~/test$ apptainer sif list test.sif
------------------------------------------------------------------------------
ID   |GROUP   |LINK    |SIF POSITION (start-end)  |TYPE
------------------------------------------------------------------------------
1    |1       |NONE    |32176-32215               |Def.FILE
2    |1       |NONE    |32215-36177               |JSON.Generic
3    |1       |NONE    |36177-36272               |JSON.Generic
4    |1       |NONE    |36864-3338240             |FS (Gocryptfs squashfs/*System/amd64)
5    |1       |4       |3338240-3338640           |Cryptographic Message (PEM/RSA-OAEP)
```

decryption
```
vagrant@bullseye:~/test$ apptainer exec --pem-path rsa_pri.pem test.sif echo "hi"
hi
```


internal/pkg/build/assemblers/sif.go
```diff
@@ -189,38 +190,56 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) error {
...
 		}
 
-		// A dm-crypt device needs to be created with squashfs
-		cryptDev := &crypt.Device{}
+		encOpts = &encryptionOptions{
+			keyInfo:   *b.Opts.EncryptionKeyInfo,
+			plaintext: []byte(g.Pass),
+		}
+	} else {
+		sylog.Debugf("creating squashfs img and might use device-mapper if encryption is requested")
```
DrDaveD, line 208: Same comment about debug message.

internal/pkg/image/driver/gocryptfsdriver.go
```diff
@@ -0,0 +1,464 @@
...
+	cmd, err = g.cmdStart(cmdPrefix, sysAttr, strings.NewReader(fmt.Sprintf("%s\n", g.pass)), nil, g.gocryptfsPath, g.decryptInfo.cipherDir, g.decryptInfo.plainDir)
+	if err != nil {
+		return err
+	}
+
+	go CheckMountInfo(g.decryptInfo.plainDir, ch)
+	err = <-ch
+	if err != nil {
+		return err
+	}
```
DrDaveD, line 236: I don't understand the point of using "go" here if it immediately waits for the response, and likewise below.

internal/pkg/image/driver/gocryptfsdriver.go
```diff
@@ -0,0 +1,464 @@
...
+	squashfusePath, err := bin.FindBin("squashfuse_ll")
+	if err != nil {
+		squashfusePath, err = bin.FindBin("squashfuse")
+		if err != nil {
+			sylog.Debugf("%v mounting not enabled because: %v", binName, err)
+			if desiredFeatures != 0 {
+				sylog.Infof("squashfuse_ll and squashfuse not found, will not be able to %v", purpose)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
DrDaveD: In fact, I was hoping you would be able to completely use the whole squashfs image driver for invoking squashfs.  There's a fair amount of tricky code in there and it's a shame to duplicate it, perhaps tripping over the same problems that have already been solved there.

internal/pkg/image/driver/gocryptfsdriver.go
```diff
@@ -0,0 +1,464 @@
...
+			break
+		}
+		time.Sleep(1 * time.Second)
+		iter++
+		if iter >= 10 {
+			ch <- fmt.Errorf("timeout to check mount info for target path: %s, timeout limit: %d s", path, iter)
+			break
+		}
+	}
+}
```
DrDaveD, line 385: This is similar to code in imagedriver.go, but not as sophisticated.  Sleeping for a second at a time is much too long in my opinion.  Let's make a common function used in both places.

internal/pkg/image/driver/gocryptfsdriver.go
```diff
@@ -0,0 +1,464 @@
...
+			return nil
+		}
+		return fmt.Errorf("could not retrieve the process status for pid: %d, err: %v", pid, err)
+	}
+
+	if wpid != 0 {
+		return fmt.Errorf("process pid: %d exited with status: %v", pid, ws.ExitStatus())
+	}
+	return nil
+}
```
DrDaveD, line 452: This has a lot in common with the image driver's stop() function.  Let's share it.

internal/pkg/image/packer/gocryptfs.go
```diff
@@ -0,0 +1,194 @@
...
+
+type cryptInfo struct {
+	cipherDir, plainDir, pass, confPath, parentDir string
+	pid                                            int
+}
+
+var defaultCleanFnGenerator = func(target string, pid int) func(bool) {
+	return func(kill bool) {
+		// if using syscall.Unmount, it'll show permission denied error
+		cmd := exec.Command("fusermount", "-u", target)
```
DrDaveD, line 33: This is a big red flag to me.  fusermount is a setuid-root program.  We should not have to be using that.  It should be happening at a point where we can use a root-mapped unprivileged user namespace to mount and unmount.

internal/pkg/runtime/engine/apptainer/container_linux.go
```diff
@@ -792,24 +792,29 @@ func (c *container) mountImage(mnt *mount.Point) error {
...
+		Size:       sizelimit,
+		Key:        key,
+		FSOptions:  opts,
+	}
+
+	if mountType == "gocryptfs" && imageDriver != nil && imageDriver.GetDriverName() == "gocryptfsDriver" {
+		return imageDriver.Mount(params, c.rpcOps.Mount)
+	}
+
+	if imageDriver != nil && imageDriver.Features()&image.ImageFeature != 0 && imageDriver.GetDriverName() == "fuseapps" {
```
DrDaveD, line 817: I don't understand how imageDriver can have both gocryptfsDriver and fuseapps as the name at the same time.  I thought they would be all one driver.

DrDaveD (changes requested):
This is my preliminary review.  I'll probably have more later.

DrDaveD:
An overall comment is that I was thinking that instead of adding a separate image driver it would be best to add another type to the existing image driver that did only the gocryptfs layer.  Then the invoking function would call the image driver 3 times, for squashfs, gocryptfs, and squashfs.  Each of the 3 mounts should get added to the `umountPoints` variable in container_linux.go, so they would get automatically unmounted in reverse order in cleanup_linux.go.  I think that would significantly simplify this implementation and make it more robust.  It would be good to see if @cclerget agrees with that approach.

internal/pkg/build/assemblers/sif.go
```diff
@@ -189,38 +190,56 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) error {
 	}
 	sylog.Verbosef("Set SIF container architecture to %s", arch)
 
-	if err := s.Create([]string{b.RootfsPath}, fsPath, flags); err != nil {
-		return fmt.Errorf("while creating squashfs: %v", err)
-	}
-
 	var encOpts *encryptionOptions
+	if b.Opts.Unprivilege {
+		sylog.Debugf("creating squashfs img and will use gocryptfs")
```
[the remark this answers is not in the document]
JasonYangShadow: resolved https://github.com/apptainer/apptainer/pull/1185/commits/42af1a456666c0082475f60be473066a5581e6ca

internal/pkg/build/assemblers/sif.go
JasonYangShadow: resolved https://github.com/apptainer/apptainer/pull/1185/commits/42af1a456666c0082475f60be473066a5581e6ca

internal/pkg/image/driver/gocryptfsdriver.go
```diff
@@ -0,0 +1,464 @@
...
+
+var sysAttr = &syscall.SysProcAttr{
+	AmbientCaps: []uintptr{
+		uintptr(capabilities.Map["CAP_SYS_ADMIN"].Value),
+		// Needed for nsenter
+		//  https://stackoverflow.com/a/69724124/10457761
+		uintptr(capabilities.Map["CAP_SYS_PTRACE"].Value),
+	},
+}
```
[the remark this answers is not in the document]
JasonYangShadow: resolved https://github.com/apptainer/apptainer/pull/1185/commits/c8241c9669e33bc80f22bc9d4c5869ee873cd28b

internal/pkg/runtime/engine/apptainer/container_linux.go
JasonYangShadow: addressed in these commits https://github.com/apptainer/apptainer/pull/1185/commits/ea57976b196239860c32afb4de903f27a0d3051f https://github.com/apptainer/apptainer/pull/1185/commits/c315e9fb461150fbc00b0aa3f94e50569dd90317

internal/pkg/image/packer/gocryptfs.go
JasonYangShadow: here, it looks like we are in the host namespace, and gocryptfs itself will also call fusermount to mount fuse fs. so in either way, fusermount will be called.

DrDaveD: That's probably true, but using fusermount to do the fuse mount (in addition to the fact that it can't be done in a root-restricted environment such as inside an apptainer container) means that cleanup can be a mess, for example if it is killed with `kill -9` (or maybe even a `^C`, you might want to try that) then there will be a leftover mount that's connected to nothing which will get `Transport endpoint is not connected` errors.
 
I think it should be possible to do the original mount in a new root-mapped unprivileged user+mount namespace.  We probably could even make it a pid namespace as well which will make sure that the FUSE process exits when the namespace is shut down.   The [`fakeroot.UnshareRootMapped()`](https://github.com/apptainer/apptainer/blob/release-1.1/internal/pkg/fakeroot/fakefake.go#L29) function may be close to what you need, and I am about to modify it to add a mount namespace so it can bind mount the home directory.  You would need to make sure the same namespace stays in place for the duration of the command, not just execute a subcommand in it, so the mountpoint stays available to the rest of the code.  The way that function is designed to be used is to pass `os.Args` to it like [here](https://github.com/apptainer/apptainer/blob/main/internal/pkg/runtime/launch/launcher_linux.go#L123) so it re-executes the whole program from scratch.  The next time through it detects that things are already set up and then skips the re-exec.

JasonYangShadow: yeah, I tested pressing `^C` when building a sif image, and it'll leave an orphan gocryptfs process.  Yeah, we should put the original mount inside a new namespace.

DrDaveD: #1230 now has a modified `UnshareRootMapped()` function that always adds a mount namespace in addition to a root namespace.  So I think with that when in fakeroot mode (either with /etc/subuid mapping or not) you should be able to start gocryptfs without it invoking fusermount.  Since we're talking about building containers it is always in fakeroot mode, so that may already take care of the trickiest part.

JasonYangShadow: thanks! I've refactored the code and use namespace for the build step.
But one test in the e2e test fails
```
2023-03-27T07:05:50.8786252Z === CONT  TestE2E/PAR/RUN/add_package_with_fakeroot_and_tmpfs
2023-03-27T07:05:50.8787895Z     apptainercmd.go:694: Running command "/usr/local/bin/apptainer exec --fakeroot --userns --writable-tmpfs --ignore-subuid --ignore-fakeroot-command /tmp/stest.3565202831/test-debian.sif sh -c apt-get update && apt-get install -y openssh-client"
2023-03-27T07:05:52.3119905Z     apptainercmd.go:694: 
2023-03-27T07:05:52.3121653Z         "/usr/local/bin/apptainer exec --fakeroot --userns --writable-tmpfs --ignore-subuid --ignore-fakeroot-command /tmp/stest.3565202831/test-debian.sif sh -c apt-get update && apt-get install -y openssh-client" output:
2023-03-27T07:05:52.3122671Z         INFO:    User not listed in /etc/subuid, trying root-mapped namespace
2023-03-27T07:05:52.3303443Z         INFO:    fakeroot command not found, using only root-mapped namespace
2023-03-27T07:05:52.3304354Z         E: setgroups 65534 failed - setgroups (1: Operation not permitted)
2023-03-27T07:05:52.3305102Z         E: setegid 65534 failed - setegid (22: Invalid argument)
2023-03-27T07:05:52.3305746Z         E: seteuid 100 failed - seteuid (22: Invalid argument)
2023-03-27T07:05:52.3306356Z         E: setgroups 0 failed - setgroups (1: Operation not permitted)
2023-03-27T07:05:52.3351223Z         W: chown to _apt:root of directory /var/lib/apt/lists/partial failed - SetupAPTPartialDirectory (22: Invalid argument)
2023-03-27T07:05:52.3352103Z         W: chown to _apt:root of directory /var/lib/apt/lists/auxfiles failed - SetupAPTPartialDirectory (22: Invalid argument)
2023-03-27T07:05:52.3352730Z         E: setgroups 65534 failed - setgroups (1: Operation not permitted)
2023-03-27T07:05:52.3353226Z         E: setegid 65534 failed - setegid (22: Invalid argument)
2023-03-27T07:05:52.3353702Z         E: seteuid 100 failed - seteuid (22: Invalid argument)
2023-03-27T07:05:52.3354224Z         E: setgroups 0 failed - setgroups (1: Operation not permitted)
2023-03-27T07:05:52.3354946Z         E: Method gave invalid 400 URI Failure message: Failed to setgroups - setgroups (1: Operation not permitted)
2023-03-27T07:05:52.3355398Z         E: Method http has died unexpectedly!
2023-03-27T07:05:52.3355863Z         E: Sub-process http returned an error code (112)
2023-03-27T07:05:52.3356580Z         FATAL:   --fakeroot requires either being in /etc/subuid, unprivileged user namespaces, or the fakeroot command
2023-03-27T07:05:52.3357206Z         Reading package lists...
2023-03-27T07:05:52.3357432Z         
2023-03-27T07:05:52.3357881Z     apptainercmd.go:694: got 255 as exit code and was expecting 100: exit status 255
```
not sure if it is related with the gocryptfs changes.

DrDaveD: That test is actually expected to fail with similar messages.  The difference is that it used to fail with a 100 code, and now it is failing with a FATAL message
```
FATAL:   --fakeroot requires either being in /etc/subuid, unprivileged user namespaces, or the fakeroot command
```
which returns a 255 error code.  It's very odd to get a fatal error at that point, after running the command.  I can reproduce interactively with this branch and
```
apptainer exec --fakeroot --userns --writable-tmpfs --ignore-subuid --ignore-fakeroot-command docker://ubuntu:20.04 apt-get update
```

JasonYangShadow: > That test is actually expected to fail with similar messages. The difference is that it used to fail with a 100 code, and now it is failing with a FATAL message
> 
> ```
> FATAL:   --fakeroot requires either being in /etc/subuid, unprivileged user namespaces, or the fakeroot command
> ```
> 
> which returns a 255 error code. It's very odd to get a fatal error at that point, after running the command. I can reproduce interactively with this branch and
> 
> ```
> apptainer exec --fakeroot --userns --writable-tmpfs --ignore-subuid --ignore-fakeroot-command docker://ubuntu:20.04 apt-get update
> ```

thanks for the help! it is introduced by my recent commit
https://github.com/apptainer/apptainer/pull/1185/commits/d1a9600aa333cc64012c933c62f7b4c372d93760
where one error is returned rather than just printed.  Never mind, I've reverted the change.

JasonYangShadow: > Actually, I was hoping you wouldn't need to add an additional call to `UnshareRootMapped()` at all. After the inclusion of #1230 (which is now merged, rebase on the main branch please) I think that you should be able to do a fuse mount anytime that fakeroot is enabled, either via /etc/subuid or via `UnshareRootMapped()`. So if you start gocryptfs after the point that fakeroot is processed, either way, then I think it should just work without invoking fusermount.

Getting a bit of confused here, if we are creating sif based on oci image or existing plain image, it won't run in `--fakeroot` mode? only if we are creating sif using definition file, then it'll run in `--fakeroot` mode automatically.  So we need to initialize gocryptfs fuse mount in two different places?  In terms of suid mode, we currently need to add two options `"--ignore-subuid", "--userns"` because imagedriver is not initialized in suid mode.

JasonYangShadow: I've refactored the code and also make fakeroot mode always enabled when building images.
https://github.com/apptainer/apptainer/blob/2c05f5dd25f2efe9b3051b923b6d07b38a4982ea/cmd/internal/cli/build.go#L351

sorry, my fault, we just need use `--userns`,  no need to use `--ignore-subuid `.

internal/pkg/image/packer/gocryptfs.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	cryptInfo.pid = realPid
+	// trap SIGINT/SIGTERM signals
+	go func() {
+		ch := make(chan os.Signal, 1)
+		signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
+		<-ch
+		sylog.Warningf("Received SIGINT/SIGTERM signal")
+		driver.KillProcess(cmd.Process.Pid, true, "gocryptfs")
+	}()
```
DrDaveD, line 116: I take it from this code that it wasn't possible to directly use the image driver here.  What prevents that?

internal/pkg/image/driver/imagedriver.go
```diff
@@ -307,20 +320,24 @@ func (d *fuseappsDriver) Mount(params *image.MountParams, mfunc image.MountFunc)
...
-		if err != nil {
-			return fmt.Errorf("unable to get wait status on %v: %v: %v", f.binName, err, filterMsg())
-		}
-		if wpid != 0 {
-			msg := filterMsg()
-			if strings.Contains(msg, "fusermount") {
-				sylog.Infof("A fusermount error indicates that the kernel is too old")
-				if params.Filesystem == "squashfs" {
-					sylog.Infof("The --unsquash option may work around it")
+		// gocryptfs has a separated notifypid
```
DrDaveD, line 323: Why does it need a notifypid?

Using "separate" in this comment would be better than "separated".

internal/pkg/image/driver/imagedriver.go
```diff
@@ -459,5 +438,56 @@ func (d *fuseappsDriver) Stop(target string) error {
 	if err = d.overlayFeature.stop(target, false); err != nil {
 		return err
 	}
+	if err = d.gocryptfsFeature.stop(target, false); err != nil {
+		return err
+	}
 	return nil
 }
+
+func KillProcess(pid int, kill bool, name string) error {
```
DrDaveD, line 447: A better name for this would be `StopFuseProcess`.

DrDaveD (changes requested):
During testing with a passphrase, both with the `--passphrase` option an with setting `APPTAINER_ENCRYPTION_PASSPHRASE`, I was able to build OK but when trying to run the encrypted sif file I always got this error:
```
FATAL:   container creation failed: mount hook function failure: mount /proc/self/fd/3->/var/apptainer/mnt/session/rootfs error: while mounting image /proc/self/fd/3: gocryptfs failed to mount /tmp/gocryptfs-2803871562/plain in 10s: failed to unlock master key: cipher: message authentication failed
Password incorrect.
```
It works for you?  I don't see an e2e test using a passphrase.

cmd/internal/cli/build.go
```diff
@@ -347,8 +347,8 @@ func preRun(cmd *cobra.Command, args []string) {
 	if buildArgs.fakeroot {
 		fakerootExec(isDeffile)
 	} else {
-		if os.Getuid() != 0 && isDeffile {
-			sylog.Verbosef("Implying --fakeroot because building from Deffile file unprivileged")
+		if os.Getuid() != 0 {
+			sylog.Verbosef("Implying --fakeroot because it is unprivileged build")
```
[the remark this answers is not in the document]
JasonYangShadow: resolved in this commit https://github.com/apptainer/apptainer/pull/1185/commits/9c7ebbd0175ce54089dd7e259bc6b6ef98584c1a

internal/pkg/build/assemblers/sif.go
```diff
@@ -189,38 +190,56 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) error {
...
-		if err != nil {
-			return fmt.Errorf("unable to obtain encryption key: %+v", err)
+		if err := g.Create([]string{b.RootfsPath}, fsPath, flags, b.TmpDir); err != nil {
+			return fmt.Errorf("while empolying gocryptfs to create squash image, err: %v", err)
 		}
 
-		// A dm-crypt device needs to be created with squashfs
-		cryptDev := &crypt.Device{}
+		encOpts = &encryptionOptions{
+			keyInfo:   *b.Opts.EncryptionKeyInfo,
```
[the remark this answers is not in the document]
JasonYangShadow: resolved in this commit https://github.com/apptainer/apptainer/pull/1185/commits/9c7ebbd0175ce54089dd7e259bc6b6ef98584c1a

internal/pkg/build/build.go
```diff
@@ -122,6 +122,9 @@ func newBuild(defs []types.Definition, conf Config) (*Build, error) {
 		var s stage
 		if conf.Opts.EncryptionKeyInfo != nil {
 			s.b, err = types.NewEncryptedBundle(parentPath, conf.Opts.TmpDir, conf.Opts.EncryptionKeyInfo)
+			if conf.Opts.Unprivilege {
+				s.b.Opts.Unprivilege = true
+			}
```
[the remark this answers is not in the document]
JasonYangShadow: resolved in this commit https://github.com/apptainer/apptainer/pull/1185/commits/9c7ebbd0175ce54089dd7e259bc6b6ef98584c1a

internal/pkg/image/driver/imagedriver.go
JasonYangShadow: resolved in this commit https://github.com/apptainer/apptainer/pull/1185/commits/9c7ebbd0175ce54089dd7e259bc6b6ef98584c1a

internal/pkg/image/driver/imagedriver.go
JasonYangShadow: gocryptfs is spawned in a separate process
resolved in this commit https://github.com/apptainer/apptainer/pull/1185/commits/9c7ebbd0175ce54089dd7e259bc6b6ef98584c1a

JasonYangShadow:
> During testing with a passphrase, both with the `--passphrase` option an with setting `APPTAINER_ENCRYPTION_PASSPHRASE`, I was able to build OK but when trying to run the encrypted sif file I always got this error:
> 
> ```
> FATAL:   container creation failed: mount hook function failure: mount /proc/self/fd/3->/var/apptainer/mnt/session/rootfs error: while mounting image /proc/self/fd/3: gocryptfs failed to mount /tmp/gocryptfs-2803871562/plain in 10s: failed to unlock master key: cipher: message authentication failed
> Password incorrect.
> ```
> 
> It works for you? I don't see an e2e test using a passphrase.

Sorry, my fault, passphrase part is not well handled. I've added the code and test cases.
https://github.com/apptainer/apptainer/pull/1185/commits/9219611cb3b7b6a06b66ccd7af71faa94f5bc8ef

internal/pkg/image/packer/gocryptfs.go
DrDaveD: InitImageDrivers does not require engine configuration, it only requires an apptainerconf.File which is a representation of apptainer.conf.  It's not hard to get.

It doesn't look to me like anything is needed from a session.  This internal image driver doesn't even appear to use anything from the DriverParams structure; you could probably just pass nil.  It does use image.MountParams but it shouldn't be difficult to create one of those structures to pass in for the parameters it needs.

JasonYangShadow:
> > During testing with a passphrase, both with the `--passphrase` option an with setting `APPTAINER_ENCRYPTION_PASSPHRASE`, I was able to build OK but when trying to run the encrypted sif file I always got this error:
> > ...
> > It works for you? I don't see an e2e test using a passphrase.
> 
> Now I can use either the `--passphrase` option on both of them or set `APPTAINER_ENCRYPTION_PASSPHRASE` on both of them but I can't mix the two types using the same passphrase. I think people would expect that to work.

I've also added one test case using both  `--passphrase` and `APPTAINER_ENCRYPTION_PASSPHRASE`.

internal/pkg/runtime/engine/apptainer/cleanup_linux.go
```diff
@@ -173,6 +178,10 @@ func umount() (err error) {
 		}
 	}
 
+	if len(gocryptfsTmp) > 0 {
+		dir := filepath.Dir(gocryptfsTmp[0])
+		defer os.RemoveAll(dir)
+	}
```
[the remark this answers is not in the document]
JasonYangShadow: It is caused by the race condition, gocryptfs require some time to unmount and then we can safely delete the temp folder. I'v added some code to wait for a certain time.

JasonYangShadow:
> I'm not sure where it's coming from, but if I have a suid installation and build with `--userns` such that it makes a gocryptfs-encrypted sif but then accidentally run it without `--userns` I get an unhelpful error message:
> 
> ```
> FATAL:   container creation failed: mount hook function failure: mount /proc/self/fd/3->/var/apptainer/mnt/session/rootfs error: while mounting image /proc/self/fd/3: gocryptfs filesystem seems not enabled and/or supported by your kernel
> ```

The log is from here 
https://github.com/JasonYangShadow/apptainer/blob/3c2f4cd9638664cae2fad60cdd73706619fecef9/internal/pkg/runtime/engine/apptainer/container_linux.go#L881
The root cause is that when we are under suid mode (without `--userns` option added), the image driver is not initialized. I'v added additional checks for this case
https://github.com/JasonYangShadow/apptainer/blob/3c2f4cd9638664cae2fad60cdd73706619fecef9/internal/pkg/runtime/engine/apptainer/container_linux.go#L821

internal/pkg/image/driver/imagedriver.go
JasonYangShadow: unlike squashfs, when we start gocryptfs via `exec.Command`, the gocryptfs process cmdline is like this
```
vagrant   879765       1  0 Mar30 ?        00:00:00 /usr/local/bin/gocryptfs -fg -notifypid=879759 /tmp/bundle-temp-680496664/gocryptfs-3985556534/cipher /tmp/bundle-temp-680496664/gocryptfs-3985556534/plain
```
where `-notifypid=879759`, the `879759` is the value of `cmd.Process.Pid`, and gocryptfs itself is running with another pid (879765). So we can not directly use the `cmd.Process.Pid` but have to search for its real pid via `-notifypid=xxx`

DrDaveD: Can't you just invoke gocryptfs with `--fg`?  That corresponds with the `-f` option that is used for the other 3 driver types, to run in the foreground.  I tried that and I don't see it using `--notifypid` then.

DrDaveD (changes requested):

internal/pkg/build/assemblers/sif.go
```diff
@@ -189,38 +190,59 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) error {
...
-
-		// TODO (schebro): Fix #3876
-		// Detach the following code from the squashfs creation. SIF can be
-		// created first and encrypted after. This gives the flexibility to
-		// encrypt an existing SIF
-		loopPath, err := cryptDev.EncryptFilesystem(fsPath, plaintext)
-		if err != nil {
-			return fmt.Errorf("unable to encrypt filesystem at %s: %+v", fsPath, err)
+		if err := g.Create([]string{b.RootfsPath}, fsPath, flags, b.TmpDir); err != nil {
+			return fmt.Errorf("while empolying gocryptfs to create squash image, err: %v", err)
```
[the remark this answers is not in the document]
JasonYangShadow: thanks! resolved

internal/pkg/image/driver/imagedriver.go
JasonYangShadow: yeah, we can set `-fg` to gocryptfs to explicitly ask it to stay foreground. I have refactored the code to implement it by putting the code inside a separate goroutine.

cclerget (approved):
LGTM

DrDaveD (changes requested):

internal/pkg/build/assemblers/sif.go
```diff
@@ -189,38 +190,59 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) error {
...
 	var encOpts *encryptionOptions
-
-	if b.Opts.EncryptionKeyInfo != nil {
-		plaintext, err := cryptkey.NewPlaintextKey(*b.Opts.EncryptionKeyInfo)
-		if err != nil {
-			return fmt.Errorf("unable to obtain encryption key: %+v", err)
+	if b.Opts.Unprivilege {
+		sylog.Debugf("Creating squashfs image and will use gocryptfs")
+		if b.Opts.EncryptionKeyInfo == nil {
+			return fmt.Errorf("unprivileged encryption build requires valid encryption key or passphrase to work properly")
```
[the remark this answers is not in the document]
JasonYangShadow: ` Surprisingly, privileged encryption without any environment variable or passphrase succeeds, although I don't know if it could ever be unencrypted.`
how did you test this?
I've tried different combinations
```
vagrant@ubuntu:~/test$ apptainer build --encrypt --force 1.sif busybox_arm64.sif 
INFO:    Starting build...
INFO:    Verifying bootstrap image busybox_arm64.sif
WARNING: failed to get key material: 404 Not Found
WARNING: integrity: signature object 5 not valid: openpgp: signature made by unknown entity
WARNING: Bootstrap image could not be verified, but build will continue.
INFO:    Creating SIF file...
ERROR:   no encryption key environment variable or --passphrase provided
FATAL:   While performing build: no encryption key environment variable or --passphrase provided
```

```
vagrant@ubuntu:~/test$ apptainer build --pem-path 1.sif busybox_arm64.sif 
Error for command "build": accepts 2 arg(s), received 1

Usage:
  apptainer [global options...] build [local options...] <IMAGE PATH> <BUILD SPEC>

Run 'apptainer --help' for more detailed usage information.
```

```
vagrant@ubuntu:~/test$ apptainer build --passphrase 1.sif busybox_arm64.sif 
Build target '1.sif' already exists and will be deleted during the build process. Do you want to continue? [N/y] y
Enter encryption passphrase: 
FATAL:   Cannot encrypt container with empty passphrase
```

cmd/internal/cli/build.go
```diff
@@ -347,9 +348,14 @@ func preRun(cmd *cobra.Command, args []string) {
...
 		fakerootExec(isDeffile)
 	} else {
-		if os.Getuid() != 0 && isDeffile {
-			sylog.Verbosef("Implying --fakeroot because building from Deffile file unprivileged")
-			fakerootExec(isDeffile)
+		if os.Getuid() != 0 {
+			if isDeffile {
+				sylog.Verbosef("Implying --fakeroot because building from Deffile file unprivileged")
+				fakerootExec(true)
+			} else if buildArgs.encrypt && (buildcfg.APPTAINER_SUID_INSTALL == 0 || buildArgs.userns) {
```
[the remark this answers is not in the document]
JasonYangShadow: thanks! resolved in this commit
https://github.com/apptainer/apptainer/pull/1185/commits/523354c8a4daec2223e38bc1de95114df7f38ef3

internal/pkg/image/packer/gocryptfs.go
```diff
@@ -0,0 +1,254 @@
...
+
+		// All entries have been checked and the mount point is already unmounted.
+		if !mount {
+			return nil
+		}
+
+	loop:
+		waitRound++
+		// maximum wait time is 5s
+		if waitRound > 200 {
```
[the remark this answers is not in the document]
JasonYangShadow: thanks! resolved in this commit
https://github.com/apptainer/apptainer/pull/1185/commits/523354c8a4daec2223e38bc1de95114df7f38ef3

internal/pkg/image/driver/imagedriver.go
```diff
@@ -404,46 +419,7 @@ func (f *fuseappsFeature) stop(target string, kill bool) error {
...
-				continue
-			}
-			sleepTime := 10 * time.Millisecond
-			time.Sleep(sleepTime)
-			totTime += sleepTime
-		}
-		// This is unexpected, because the kill signal at half
-		//  of maxTime should kill quickly
-		return fmt.Errorf("took more than %v to stop %v pid %v", maxTime, f.binName, process.Pid)
+		return stopFuseProcess(instance.cmd.Process.Pid, kill, f.binName)
```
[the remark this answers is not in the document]
JasonYangShadow: thanks! resolved in this commit
https://github.com/apptainer/apptainer/pull/1185/commits/523354c8a4daec2223e38bc1de95114df7f38ef3

internal/pkg/image/driver/imagedriver.go
```diff
@@ -31,6 +31,15 @@ import (
...
 
+var sysAttr = &syscall.SysProcAttr{
+	AmbientCaps: []uintptr{
+		uintptr(capabilities.Map["CAP_SYS_ADMIN"].Value),
+		// Needed for nsenter
+		//  https://stackoverflow.com/a/69724124/10457761
+		uintptr(capabilities.Map["CAP_SYS_PTRACE"].Value),
+	},
+}
+
```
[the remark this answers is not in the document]
JasonYangShadow: thanks! resolved in this commit
https://github.com/apptainer/apptainer/pull/1185/commits/523354c8a4daec2223e38bc1de95114df7f38ef3

internal/pkg/runtime/engine/apptainer/container_linux.go
```diff
@@ -792,27 +792,35 @@ func (c *container) mountImage(mnt *mount.Point) error {
...
-			Key:        key,
-			FSOptions:  opts,
+		if mountType == "gocryptfs" {
+			return gocryptfsMount(params, c.rpcOps.Mount)
 		}
 		return imageDriver.Mount(params, c.rpcOps.Mount)
 	}
 
+	if imageDriver == nil && mountType == "gocryptfs" {
+		return fmt.Errorf("gocryptfs requires user namespace enabled, please add `--userns` option")
```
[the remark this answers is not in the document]
JasonYangShadow: thanks! resolved in this commit
https://github.com/apptainer/apptainer/pull/1185/commits/523354c8a4daec2223e38bc1de95114df7f38ef3

cmd/internal/cli/build.go
```diff
@@ -347,9 +348,14 @@ func preRun(cmd *cobra.Command, args []string) {
 	if buildArgs.fakeroot {
 		fakerootExec(isDeffile)
 	} else {
-		if os.Getuid() != 0 && isDeffile {
-			sylog.Verbosef("Implying --fakeroot because building from Deffile file unprivileged")
-			fakerootExec(isDeffile)
+		if os.Getuid() != 0 {
+			if isDeffile {
+				sylog.Verbosef("Implying --fakeroot because building from Deffile file unprivileged")
```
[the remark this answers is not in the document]
JasonYangShadow: thanks! resolved in this commit
https://github.com/apptainer/apptainer/pull/1185/commits/523354c8a4daec2223e38bc1de95114df7f38ef3

CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ For older changes see the [archived Singularity change log](https://github.com/a
   flag.
 - Support for online verification checks of x509 certificates using OCSP protocol.
   (introduced flag: `verify --ocsp-verify`)
+- Support for unprivileged encryption using gocryptfs.
```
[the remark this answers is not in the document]
JasonYangShadow: thanks! resolved in this commit
https://github.com/apptainer/apptainer/pull/1185/commits/523354c8a4daec2223e38bc1de95114df7f38ef3

internal/pkg/build/assemblers/sif.go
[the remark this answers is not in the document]
JasonYangShadow: I've added sylog error in this commit
https://github.com/apptainer/apptainer/pull/1185/commits/523354c8a4daec2223e38bc1de95114df7f38ef3

cmd/internal/cli/build.go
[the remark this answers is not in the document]
JasonYangShadow: also stderr will be outputed if error occurs
https://github.com/apptainer/apptainer/pull/1185/commits/6985daa0859d6f817698ee1511d1e4485c4134c9

JasonYangShadow:
why the e2e test frequently failed with `failed to find loop device`
```
2023-04-11T11:59:25.1139008Z === CONT  TestE2E/PAR/LEGACY/run_legacy/True
2023-04-11T11:59:25.1139400Z     actions.go:80: 
2023-04-11T11:59:25.1140535Z         "/usr/local/bin/apptainer exec --bind /tmp/stest.4177996430/test.sif:/test.img /tmp/stest.4177996430/test-singularity.sif singularity run /test.img true" output:
2023-04-11T11:59:25.1141933Z         FATAL:   container creation failed: mount /proc/self/fd/3->/var/lib/singularity/mnt/session/rootfs error: while mounting image /proc/self/fd/3: failed to find loop device: could not attach image file to loop device: no loop devices available
2023-04-11T11:59:25.1142441Z         
2023-04-11T11:59:25.1142860Z     actions.go:80: got 255 as exit code and was expecting 0: exit status 255
2023-04-11T11:59:25.1143869Z         waiting for command "/usr/local/bin/apptainer exec --bind /tmp/stest.4177996430/test.sif:/test.img /tmp/stest.4177996430/test-singularity.sif singularity run /test.img true"
2023-04-11T11:59:25.1144818Z         github.com/apptainer/apptainer/e2e/internal/e2e.TestEnv.RunApptainer.func1
2023-04-11T11:59:25.1145427Z         	github.com/apptainer/apptainer/e2e/internal/e2e/apptainercmd.go:669
2023-04-11T11:59:25.1145959Z         github.com/apptainer/apptainer/e2e/internal/e2e.Privileged.func1
2023-04-11T11:59:25.1146504Z         	github.com/apptainer/apptainer/e2e/internal/e2e/priv.go:88
2023-04-11T11:59:25.1146823Z         testing.tRunner
2023-04-11T11:59:25.1147167Z         	testing/testing.go:1446
2023-04-11T11:59:25.1147449Z         runtime.goexit
2023-04-11T11:59:25.1147739Z         	runtime/asm_amd64.s:1594
```

DrDaveD:
> > why the e2e test frequently failed with `failed to find loop device`
> 
> That's a [known problem](sylabs/singularity#1499). I expect the fix to be imported today.

As it turns out their fix is not relevant to apptainer because we were already using /dev/loop-control, and the e2e tests run on Ubuntu 22 where the problem they found was exclusively only on a recent Ubuntu 20.  I'm going to create a new issue and ask Cedric to look at it.

internal/pkg/build/assemblers/sif.go
```diff
@@ -189,38 +190,61 @@ func (a *SIFAssembler) Assemble(b *types.Bundle, path string) error {
...
-
-		fsPath = loopPath
 
 		encOpts = &encryptionOptions{
 			keyInfo:   *b.Opts.EncryptionKeyInfo,
-			plaintext: plaintext,
+			plaintext: []byte(g.Pass),
+		}
+	} else {
+		sylog.Debugf("Creating squashfs img and might use device-mapper if encryption is requested")
```
DrDaveD, line 213: I think this would be better as two debug messages, one here saying "Creating squashfs image" and another one a little below when it is using device-mapper encryption just saying "Using device-mapper encryption".

internal/pkg/image/driver/imagedriver.go
```diff
@@ -245,17 +252,22 @@ func (d *fuseappsDriver) Mount(params *image.MountParams, mfunc image.MountFunc)
...
-	cmd.SysProcAttr = &syscall.SysProcAttr{
-		AmbientCaps: []uintptr{
-			uintptr(capabilities.Map["CAP_SYS_ADMIN"].Value),
-			// Needed for nsenter
-			//  https://stackoverflow.com/a/69724124/10457761
-			uintptr(capabilities.Map["CAP_SYS_PTRACE"].Value),
-		},
+
+	// when using gocryptfs for build step, we should not use SysProcAttr
+	if !(params.Filesystem == "gocryptfs" && params.SysProcOmit) {
```
DrDaveD, line 257: This could only check the params.SysProcOmit flag, it doesn't also have to check for gocryptfs. 

I like the name DontElevatePrivs better.

internal/pkg/image/driver/imagedriver.go
```diff
@@ -307,10 +323,11 @@ func (d *fuseappsDriver) Mount(params *image.MountParams, mfunc image.MountFunc)
 		time.Sleep(sleepTime)
 		totTime += sleepTime
 		var ws syscall.WaitStatus
-		wpid, err := syscall.Wait4(process.Pid, &ws, syscall.WNOHANG, nil)
-		if err != nil {
+		wpid, err := syscall.Wait4(cmd.Process.Pid, &ws, syscall.WNOHANG, nil)
```
DrDaveD, line 326: Why did this change?

internal/pkg/image/packer/gocryptfs.go
```diff
@@ -0,0 +1,207 @@
...
+	case cryptkey.Passphrase:
+		cryptInfo.pass = g.keyInfo.Material
+	default:
+		err = errors.New("cryptkey type is unknown")
+		return
+	}
+	cryptInfo.confPath = filepath.Join(cipherDir, "gocryptfs.conf")
+
+	errCh := make(chan error, 1)
+	go func() {
```
DrDaveD, line 106: I'm very happy to see this using the image driver now.  I wonder if this go routine here is still needed, since the image driver Mount already forks a process?

DrDaveD (changes requested):

internal/pkg/build/assemblers/sif.go
JasonYangShadow: resolved in this commit
https://github.com/apptainer/apptainer/pull/1185/commits/17c45e4d5d0eab05651c9877307fcb7d58282232

internal/pkg/image/driver/imagedriver.go
JasonYangShadow: resolved in this commit
https://github.com/apptainer/apptainer/pull/1185/commits/17c45e4d5d0eab05651c9877307fcb7d58282232

internal/pkg/image/packer/gocryptfs.go
JasonYangShadow: resolved in this commit
https://github.com/apptainer/apptainer/pull/1185/commits/17c45e4d5d0eab05651c9877307fcb7d58282232

internal/pkg/build/assemblers/sif.go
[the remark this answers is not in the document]
DrDaveD: Oh, no, a sylog error here does not provide an earlier notification, it just ends up giving two duplicate error messages in a row:
```
ERROR:   no encryption key environment variable or --passphrase provided
FATAL:   While performing build: no encryption key environment variable or --passphrase provided
```
So please remove that extra sylog.Errorf().  I was talking about adding a check much earlier, at the place where it would normally prompt for the passphrase, so it would die before doing any building.

DrDaveD (changes requested):
This is getting really close.  In addition to the earlier check for missing passphrase, please consolidate the commits.  That can be one commit or a small number each covering different logical pieces.  I just don't think it's good to merge commits that include changes that weren't approved separately.

DrDaveD:
Rebase on current main to avoid loop device errors.

JasonYangShadow:
> This is getting really close. In addition to the earlier check for missing passphrase, please consolidate the commits. That can be one commit or a small number each covering different logical pieces. I just don't think it's good to merge commits that include changes that weren't approved separately.

I've rebased and merged into https://github.com/apptainer/apptainer/pull/1185/commits/65d984ff5c020ee875c8e7908fa1fcee7b92cba7

cmd/internal/cli/build_linux.go
```diff
@@ -220,16 +223,22 @@ func runBuild(cmd *cobra.Command, args []string) {
...
 		}
 
 		k, err := getEncryptionMaterial(cmd)
 		if err != nil {
 			sylog.Fatalf("While handling encryption material: %v", err)
 		}
 		keyInfo = k
+
+		if keyInfo == nil && unprivilege {
+			sylog.Errorf("Please provide valid encryption info, please add `--pem-path` or `--passphrase`")
```
DrDaveD, line 239: I think it would be good to mention environment variables too, and not say "Please" twice.

I suggest
```
Missing encryption info, please add `--passphrase` or `--pem-path` or corresponding environment variable
```

DrDaveD (changes requested):
When you rebase, use the "rebase" command instead of "merge", because that does not add an extra commit.

cmd/internal/cli/build_linux.go
JasonYangShadow: refected in this commit
https://github.com/apptainer/apptainer/pull/1185/commits/b2c0de740fb01d972ba614547cdf26c176d49cc8

DrDaveD (approved):

DrDaveD:
Congratulations!

Please work on corresponding changes to apptainer-userdocs when you have a chance.

JasonYangShadow:
> Congratulations!
> 
> Please work on corresponding changes to apptainer-userdocs when you have a chance.

Sincerely thanks for your patience and great comments. This is my first big PR to apptainer, I've learned a lot from our conversations. And I would like to work on other big features in the future to contribute to the community. ❤️

PR #1226, opened 2023-03-22
bci like container if no registration is given

mslacken:
Signed-off-by: Christian Goll <cgoll@suse.de>

## Simplify SLE builds

Add the possibility to build SUSE containers via suseconnect-container, what
removed the need to register the container via SUSEConnect.
This needs the possibility to have bind mounts in the post stage of the build,
but this is not exposed in the def file.

PR for the 1.1-release branch, cherry picked from main


DrDaveD (changes requested):
This picks only the second commit with the log entry, not the actual code changes.  Also it is submitted against main, not release-1.1.

mslacken:
thanks for pointing this out
I have made a new pr #1229

PR #1273, opened 2023-04-12
remove unwanted colors to stderr (release-1.1)

JasonYangShadow:
## Description of the Pull Request (PR):

cherry pick https://github.com/apptainer/apptainer/pull/1268/commits/d8ee12fd706ec6d809eeee9b643bcc087b719944

### This fixes or addresses the following GitHub issues:

 - Fixes #1259 

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/apptainer/apptainer/blob/main/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTORS.md)

PR #1278, opened 2023-04-12
Add xino=on for writable overlay mount points

### LINKED ISSUE #1267: `tar` breaks in overlay when used in memory-constrained environment
### Version of Apptainer

```
$ apptainer --version
apptainer version 1.1.7
```

### Expected behavior

Directory inode numbers should remain constant in apptainer overlays after dropping virtual memory cache.

### Actual behavior

Directory inodes change after dropping virtual memory cache. This leads to such errors when extracting tar archives in apptainer overlay when relatively low ammount of memory is available in the environment:
```
tar: ./deps/0/bin: Directory renamed before its status could be extracted
```
I think that this exact error was already reported (and pretty well described) in ubuntu [here](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1728489). However, the fix that keeps the inode numbers constant mentioned there only applies when `lowerdir`, `upperdir` and `workdir` are all on the same mount point (as hinted [here](https://lwn.net/Articles/720794/)). I suspect that the problem is caused by this as running apptainer with `-d` flag indicates that `lowerdir` is on a different mount point than `workdir` and `upperdir`.

```
DEBUG   [U=0,P=2642035]    mountImage()                  Mounting loop device /dev/loop0 to /usr/local/var/apptainer/mnt/session/rootfs of type squashfs
DEBUG   [U=0,P=2642035]    mountGeneric()                Mounting /home/gplaczek/git/apptainer/sin to /usr/local/var/apptainer/mnt/session/overlay-images/0
```

The overlay is mounted like this:
```
$ mount | grep overlay
overlay on / type overlay (rw,nodev,relatime,lowerdir=/usr/local/var/apptainer/mnt/session/overlay-lowerdir:/usr/local/var/apptainer/mnt/session/rootfs,upperdir=/usr/local/var/apptainer/mnt/session/overlay-images/0/upper,workdir=/usr/local/var/apptainer/mnt/session/overlay-images/0/work)
```

If the output from both `stat` commands below is different, the issue was properly reproduced

```
$ mkdir overlay
$ sudo apptainer shell --overlay overlay docker://debian:bullseye
Apptainer> cd /
Apptainer> mkdir dir
Apptainer> stat -c%i dir
Apptainer> echo 2 > /proc/sys/vm/drop_caches
Apptainer> stat -c%i dir
```

To reproduce the issue with `tar` command:

```
$ mkdir overlay
$ sudo apptainer shell --overlay overlay docker://debian:bullseye
Apptainer> cd /
Apptainer> apt update && apt install wget
Apptainer> wget https://github.com/torvalds/linux/archive/refs/tags/v6.3-rc5.tar.gz # this or any relatively big tar
Apptainer> tar -xf v6.3-rc5.tar.gz & for i in {0..5}; do echo 2 > /proc/sys/vm/drop_caches; sleep 1; done
```

Running the commands above is expected to produce errors similar to these:

```
tar: linux-6.3-rc5/scripts/dtc/include-prefixes: Directory renamed before its status could be extracted
tar: linux-6.3-rc5/scripts/dtc: Directory renamed before its status could be extracted
tar: linux-6.3-rc5/scripts: Directory renamed before its status could be extracted
tar: linux-6.3-rc5/include/dt-bindings/input: Directory renamed before its status could be extracted
tar: linux-6.3-rc5/include/dt-bindings: Directory renamed before its status could be extracted
tar: linux-6.3-rc5/include: Directory renamed before its status could be extracted
tar: linux-6.3-rc5/arch/arm64/boot/dts/arm: Directory renamed before its status could be extracted
tar: linux-6.3-rc5/arch/arm64/boot/dts: Directory renamed before its status could be extracted
tar: linux-6.3-rc5/arch/arm64/boot: Directory renamed before its status could be extracted
tar: linux-6.3-rc5/arch/arm64: Directory renamed before its status could be extracted
tar: linux-6.3-rc5/arch: Directory renamed before its status could be extracted
tar: linux-6.3-rc5: Directory renamed before its status could be extracted
tar: Exiting with failure status due to previous errors
```

These errors will occasionally happen when user has relarively low ammount of available memory and cache drops will happen naturally.

```sh
$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

I compiled and installed Apptainer manually following instructions from [INSTALL.md](https://github.com/apptainer/apptainer/blob/main/INSTALL.md)

If you're interested in fixing this issue, I would be eager to contribute and implement it myself if you provide me with some hints on how to approach the problem of keeping all overlays on the same mount point in your app.

cclerget:

 - Fixes #1267

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - `sessiondir maxsize` in `apptainer.conf` now defaults to 64 MiB for new
   installations. This is an increase from 16 MiB in prior versions.
 - Show standard output of yum bootstrap if log level is verbose or higher.
+- Add `xino=on` mount option for writable kernel overlay mount points to fix
+  inode numbers consistency after kernel cache flush (not applicable to fuse-overlayfs).
 
 ### New features / functionalities
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,6 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - `sessiondir maxsize` in `apptainer.conf` now defaults to 64 MiB for new
   installations. This is an increase from 16 MiB in prior versions.
 - Show standard output of yum bootstrap if log level is verbose or higher.
-- Add `xino=on` mount option for writable kernel overlay mount points to fix
-  inode numbers consistency after kernel cache flush (not applicable to fuse-overlayfs).
 
 ### New features / functionalities
 
@@ -75,11 +73,14 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Fix joining cgroup of instance started as root, with cgroups v1,
   non-default cgroupfs manager, and no device rules.
 - Ensure `DOCKER_HOST` is honored in non-build flows.
-- Fix loop device 'no such device or address' spurious errors when using shared
-  loop devices.
 
 ### Bug fixes
 
+- Fix loop device 'no such device or address' spurious errors when using shared
+  loop devices.
+- Add `xino=on` mount option for writable kernel overlay mount points to fix
+  inode numbers consistency after kernel cache flush (not applicable to
+  fuse-overlayfs).
 - Remove unwanted colors to STDERR.
 
 ## v1.1.7 - \[2023-03-28\]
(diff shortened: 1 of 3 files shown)


DrDaveD (approved):

DrDaveD:
Please cherry-pick this also to release-1.1, after #1273 is merged.

PR #1295, opened 2023-04-23
Add arch aware support for apptainer cache

### LINKED ISSUE #886: apptainer cache is not architecture aware
### Version of Apptainer

```
$ apptainer --version
apptainer version 1.1.3-1.el9
```

### Expected behavior

```
/usr/bin/apptainer run docker://alpine:latest hostname
```

ran on both aarch64 and on x86_64 should be posible with the same cache directory.

### Actual behavior

A cached on `x86_64`  image fails on `aarch64` on.

1. Login into an aarch64 node.

```
 /usr/bin/apptainer run docker://alpine:latest hostname
lxplus9s05.example.org
```

2. Login into an x86_64 node.

```
/usr/bin/apptainer run docker://alpine:latest hostname
FATAL:   could not open image /afs/example.ch/user/s/someone/.apptainer/cache/oci-tmp/8914eb54f968791faf6a8638949e480fef81e697984fba772b3976835194c6d4: the image's architecture (arm64) could not run on the host's (amd64)
```

If I purge the cache  all is good and a `x86_64` item is cached.

Need `aarch64` and `x86_64` sharing a `/home` filesystem.

```
CentOS 9
```

EPEL.

### Summary:
So while the `docker://alpine:latest` is multiarch once it is cache it is not.

### Workarond

Use the `--disablecache` option.

JasonYangShadow:

Add arch aware support for apptainer cache

 - Fixes #886 

### DIFF
--- a/cmd/internal/cli/pull.go
+++ b/cmd/internal/cli/pull.go
@@ -16,6 +16,7 @@ import (
 	"runtime"
 
 	"github.com/apptainer/apptainer/docs"
+	build_oci "github.com/apptainer/apptainer/internal/pkg/build/oci"
 	"github.com/apptainer/apptainer/internal/pkg/cache"
 	"github.com/apptainer/apptainer/internal/pkg/client/library"
 	"github.com/apptainer/apptainer/internal/pkg/client/net"
@@ -54,6 +55,8 @@ var (
 	// pullArch is the architecture for which containers will be pulled from the
 	// SCS library.
 	pullArch string
+	// pullArchVariant is the architecture variant, e.g., arm32v5, arm32v6, arm32v7, v5,v6,v7 are variants
+	pullArchVariant string
 )
 
 // --arch
@@ -64,6 +67,16 @@ var pullArchFlag = cmdline.Flag{
 	Name:         "arch",
 	Usage:        "architecture to pull from library",
 	EnvKeys:      []string{"PULL_ARCH"},
+}
+
+// --arch
+var pullArchVariantFlag = cmdline.Flag{
+	ID:           "pullArchVariantFlag",
+	Value:        &pullArchVariant,
+	DefaultValue: "",
+	Name:         "arch-variant",
+	Usage:        "architecture variant to pull from library",
+	EnvKeys:      []string{"PULL_ARCH_VARIANT"},
 }
 
 // --library
@@ -152,6 +165,7 @@ func init() {
 		cmdManager.RegisterFlagForCmd(&pullAllowUnsignedFlag, PullCmd)
 		cmdManager.RegisterFlagForCmd(&pullAllowUnauthenticatedFlag, PullCmd)
 		cmdManager.RegisterFlagForCmd(&pullArchFlag, PullCmd)
+		cmdManager.RegisterFlagForCmd(&pullArchVariantFlag, PullCmd)
 	})
 }
 
@@ -268,12 +282,18 @@ func pullRun(cmd *cobra.Command, args []string) {
 			sylog.Fatalf("While creating Docker credentials: %v", err)
 		}
 
+		arch, err := build_oci.ConvertArch(pullArch, pullArchVariant)
+		if err != nil {
+			sylog.Fatalf("While processing the arch and arch variant: %v", err)
+			return
+		}
 		pullOpts := oci.PullOptions{
 			TmpDir:     tmpDir,
 			OciAuth:    ociAuth,
 			DockerHost: dockerHost,
 			NoHTTPS:    noHTTPS,
 			NoCleanUp:  buildArgs.noCleanUp,
+			Pullarch:   arch,
 		}
 
 		_, err = oci.PullToFile(ctx, imgCache, pullTo, pullFrom, pullOpts)
--- a/internal/pkg/build/oci/oci.go
+++ b/internal/pkg/build/oci/oci.go
@@ -15,6 +15,8 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"io"
+	"os"
+	"runtime"
 	"strings"
 
 	"github.com/apptainer/apptainer/internal/pkg/cache"
@@ -33,6 +35,46 @@ type ImageReference struct {
 	types.ImageReference
 }
 
+type GoArch struct {
+	Arch string
+	Var  string
+}
+
+var ArchMap = map[string]GoArch{
+	"amd64": {
+		Arch: "amd64",
+		Var:  "",
+	},
+	"arm32v5": {
+		Arch: "arm",
+		Var:  "v5",
+	},
+	"arm32v6": {
+		Arch: "arm",
+		Var:  "v6",
+	},
+	"arm32v7": {
+		Arch: "arm",
+		Var:  "v7",
+	},
+	"arm64v8": {
+		Arch: "arm64",
+		Var:  "v8",
+	},
+	"386": {
+		Arch: "386",
+		Var:  "",
+	},
+	"ppc64le": {
+		Arch: "ppc64le",
+		Var:  "",
+	},
+	"s390x": {
+		Arch: "s390x",
+		Var:  "",
+	},
+}
+
 // ConvertReference converts a source reference into a cache.ImageReference to cache its blobs
 func ConvertReference(ctx context.Context, imgCache *cache.Handle, src types.ImageReference, sys *types.SystemContext) (types.ImageReference, error) {
 	if imgCache == nil {
@@ -42,6 +84,13 @@ func ConvertReference(ctx context.Context, imgCache *cache.Handle, src types.Ima
 	// Our cache dir is an OCI directory. We are using this as a 'blob pool'
 	// storing all incoming containers under unique tags, which are a hash of
 	// their source URI.
+	if sys == nil {
+		var err error
+		sys, err = defaultSysCtx()
+		if err != nil {
+			return nil, fmt.Errorf("unable to create default system context: %v", err)
+		}
+	}
 	cacheTag, err := getRefDigest(ctx, src, sys)
 	if err != nil {
 		return nil, err
@@ -88,7 +137,7 @@ func (t *ImageReference) newImageSource(ctx context.Context, sys *types.SystemCo
 // ParseImageName parses a uri (e.g. docker://ubuntu) into it's transport:reference
 // combination and then returns the proper reference
 func ParseImageName(ctx context.Context, imgCache *cache.Handle, uri string, sys *types.SystemContext) (types.ImageReference, error) {
-	ref, err := parseURI(uri)
+	ref, _, err := parseURI(uri)
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse image name %v: %v", uri, err)
 	}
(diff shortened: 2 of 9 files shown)


internal/pkg/cache/cache.go
```diff
@@ -162,9 +165,32 @@ func (h *Handle) GetEntry(cacheType string, hash string) (e *Entry, err error) {
...
+		}
+	}
+
+	for _, arch := range archs {
+		dir := path.Join(parentDir, arch, cacheType)
+		err = h.cleanCache(dir, cacheType, dryRun, days)
+		if err != nil {
+			return
+		}
+	}
```
DrDaveD, line 188: Is it necessary (or helpful) to separately delete each subdirectory, instead of just deleting the whole directory?

internal/pkg/cache/cache.go
```diff
@@ -268,7 +294,11 @@ func New(cfg Config) (h *Handle, err error) {
 	}
 
 	// Initialize the root directory of the cache
-	rootDir := path.Join(parentDir, SubDirName)
+	// https://github.com/apptainer/apptainer/issues/886
```
DrDaveD, line 297: I don't think it's helpful to refer to the issue here.  I think it's pretty clear that it is adding the architecture subdirectory if defined.

DrDaveD (changes requested):

dtrudg:
I don't think the approach of just setting a cache rootdir with the `GOARCH` in effect will be sufficient. Consider:

* Pulling a library image of a different architecture explicitly, e.g. `singularity pull --arch arm64 library://alpine` on an amd64 machine. 
* Pulling a single-architecture docker image explictly `singularity pull docker://arm64v8/ubuntu` (this is always arm64 as there is no manifest list for different architectures).

JasonYangShadow:
@DrDaveD @dtrudg After checking the code, found that there are some issues in current implementation. 
1. We can not actually pull arm64 on amd64 machine. `--arch` is useless here. The only way we can pull a different arch image is `docker://arch/alpine:latest`
2. Even we can pull different images https://github.com/JasonYangShadow/apptainer/blob/99629731baa5fd73b5ae8bb7be478a863778c921/internal/pkg/build/oci/oci.go#L134, it looks like this line and https://github.com/JasonYangShadow/apptainer/blob/99629731baa5fd73b5ae8bb7be478a863778c921/internal/pkg/build/oci/oci.go#L159 won't change. 

So I make the `--arch` useful in current `pull` command and by default the `runtime.GOARCH` will be used for pulling the remote image. 
As the digest is only used for cache (in our impl), so I change it rather than use the real one from manifest.

dtrudg:
> @DrDaveD @dtrudg After checking the code, found that there are some issues in current implementation.
> 
>  1. We can not actually pull arm64 on amd64 machine. `--arch` is useless here. The only way we can pull a different arch image is `docker://arch/alpine:latest`

Yes, my reference to `--arch` was only for pulls from `library://` sources, not `docker://` ones.

>  2. Even we can pull different images https://github.com/JasonYangShadow/apptainer/blob/99629731baa5fd73b5ae8bb7be478a863778c921/internal/pkg/build/oci/oci.go#L134, it looks like this line and https://github.com/JasonYangShadow/apptainer/blob/99629731baa5fd73b5ae8bb7be478a863778c921/internal/pkg/build/oci/oci.go#L159 won't change.
> 
> 
> So I make the `--arch` useful in current `pull` command and by default the `runtime.GOARCH` will be used for pulling the remote image. As the digest is only used for cache (in our impl), so I change it rather than use the real one from manifest.

Be very careful with the digest. In Docker registries, a multi-arch repository has a manifest list. This is something that links to multiple manifests, one for each architecture. The manifest list isn't different per architecture, only the manifest it refers onto, for a specific architecture, is.

What you have done might work, but you need to look very closely at what exactly is being retrieved and hashed in various circumstances, to be sure the digest used is appropriate for correct caching behaviour. There are many combinations to consider, e.g.

* Pulling from a multi-arch Docker registry repository (which provides a manifest list)  without `--arch` 
* Pulling from a multi-arch Docker registry repository (which provides a manifest list)  with a different  `--arch`
* Pulling from a single-arch Docker registry repository (which provides a single manifest) without `--arch` on a system where GOARCH matches the container arch.
* Pulling from a single-arch Docker registry repository (which provides a single manifest) without `--arch` on a system where GOARCH does not match the container arch.
* Pulling from a single-arch Docker registry repository (which provides a single manifest) using `--arch` which matches the container arch.
* Pulling from a single-arch Docker registry repository (which provides a single manifest) using `--arch` which does not match the container arch.

Probably this should be well enforced by tests, as the code is using the containers/image library, which relatively frequently bumps major version.

Good luck :-) 

I'll unsubscribe from this issue now, as how you do it is really a decision for the Apptainer developers.

DrDaveD:
@JasonYangShadow is this ready for re-review?

JasonYangShadow:
> @JasonYangShadow is this ready for re-review?

yes, please help re-review

internal/pkg/build/oci/oci.go
```diff
@@ -33,6 +33,16 @@ type ImageReference struct {
 	types.ImageReference
 }
 
+var archMap = map[string]string{
+	"amd64":   "amd64",
+	"armv6":   "arm",
+	"armv7":   "arm",
```
DrDaveD, line 39: Why is it OK for those two to share the same cache?

internal/pkg/build/oci/oci.go
```diff
@@ -88,38 +98,64 @@ func (t *ImageReference) newImageSource(ctx context.Context, sys *types.SystemCo
...
+		return nil, arch, fmt.Errorf("%s not a registered transport", split[0])
+	}
+
+	uriTmp := split[1]
+	uriTmp = strings.TrimPrefix(uriTmp, "//")
+	archTmp := strings.SplitN(uriTmp, "/", 2)
+
+	if len(archTmp) == 2 {
+		if val, ok := archMap[archTmp[0]]; ok {
+			// replace with golang runtime.GOARCH
```
DrDaveD, line 129: I don't understand this comment.

DrDaveD: Oh, I think you mean "translate into golang's corresponding runtime.GOARCH value"? Is that right?

internal/pkg/build/oci/oci.go
```diff
@@ -88,38 +98,64 @@ func (t *ImageReference) newImageSource(ctx context.Context, sys *types.SystemCo
...
+			arch = val
+			// we need to convert the docker://arch/image:tag to docker://image:tag to make sure digest is the same
+			convertURI := strings.Replace(split[1], fmt.Sprintf("%s/", archTmp[0]), "", 1)
```
DrDaveD, line 132: What does this do?  How can the tag be the same for every architecture?

DrDaveD (changes requested):

internal/pkg/build/oci/oci.go
JasonYangShadow: https://go.dev/doc/install/source#environment
https://docs.docker.com/build/building/multi-platform/
we are acutally using goarch value here, but to map it to docker image archs, we need a map.

internal/pkg/build/oci/oci.go
JasonYangShadow: yes, it's mapping the docker image <-> golang arch

DrDaveD (approved):
This looks good to me but I would also like @cclerget to review it.

JasonYangShadow:
> This looks good to me but I would also like @cclerget to review it.

yes, that's the reason why I also add @cclerget into the review list. The changes will bring impact to frequent image pull action. Let's discuss and review it slowly.

JasonYangShadow:
@cclerget updated, please check when you have time. I set `arch=''` by default, because in this case
`apptainer pull --arch amd64 --force test.sif docker://alpine:latest` on arm32v6 machine
the user should expect to pull `amd64` image type, otherwise different value in `--arch` will be useless.

internal/pkg/build/oci/oci.go
```diff
@@ -88,38 +98,51 @@ func (t *ImageReference) newImageSource(ctx context.Context, sys *types.SystemCo
...
 
-	return transport.ParseReference(split[1])
+	imgRef, err := transport.ParseReference(split[1])
+	return imgRef, arch, err
 }
 
 // ImageDigest obtains the digest of a uri's manifest
-func ImageDigest(ctx context.Context, uri string, sys *types.SystemContext) (string, error) {
-	ref, err := parseURI(uri)
+func ImageDigest(ctx context.Context, uri string, sys *types.SystemContext, pullArch string) (digest string, err error) {
```
cclerget, line 129: I think you can remove `pullArch` here and use `ArchitectureChoice` in `types.SystemContext`

internal/pkg/build/oci/oci.go
```diff
@@ -88,38 +98,51 @@ func (t *ImageReference) newImageSource(ctx context.Context, sys *types.SystemCo
...
+	ref, arch, err := parseURI(uri)
 	if err != nil {
 		return "", fmt.Errorf("unable to parse image name %v: %v", uri, err)
 	}
 
-	return getRefDigest(ctx, ref, sys)
+	digest, err = getRefDigest(ctx, ref, sys)
```
cclerget, line 135: `getRefDigest` should follow the new hash computation logic (digest+arch), currently it looks like only pull will use this new digest, build will keep using old digest leading to cache miss

cclerget (changes requested):
Globally, using `sys.ArchitectureChoice` seems more natural and the new digest computation should be done both in `getRefDigest` and `getDockerRefDigest` instead of `ImageDigest` by retrieving the arch from `sys.ArchitectureChoice` or use current `GOARCH` if empty, to ensure that both pull/build operations generate/get the same cache entry

internal/pkg/build/oci/oci.go
JasonYangShadow: thanks! I changed to use `ArchitectureChoice` in the latest commit
https://github.com/apptainer/apptainer/pull/1295/commits/470ee6eaea3daaebd8135d5f3044926d0c5ebece

internal/pkg/build/oci/oci.go
JasonYangShadow: thanks! I made changes in the latest commit
https://github.com/apptainer/apptainer/pull/1295/commits/470ee6eaea3daaebd8135d5f3044926d0c5ebece

JasonYangShadow:
> Globally, using `sys.ArchitectureChoice` seems more natural and the new digest computation should be done both in `getRefDigest` and `getDockerRefDigest` instead of `ImageDigest` by retrieving the arch from `sys.ArchitectureChoice` or use current `GOARCH` if empty, to ensure that both pull/build operations generate/get the same cache entry

I've also added another test case for build case.

internal/pkg/build/oci/oci.go
cclerget: To continue on this discussion, we should probably add the arch variants into consideration when computing the new digest, between armv6/armv7 here seems not correct and armv5 is missing, looks like the correct values are:
- `arm32v5`
- `arm32v6`
- `arm32v7`

Mapped to Go runtime variables:
- `arm32v5` :
  - `GOARCH = arm`
  - `GOARM = 5`
- `arm32v6` :
  - `GOARCH = arm`
  - `GOARM = 6`
- `arm32v7` :
  - `GOARCH = arm`
  - `GOARM = 7`

`types.SystemContext` has `VariantChoice` and for arm cases we would end up with:
- `arm32v5`:
  - `ArchitectureChoice = arm`
  - `VariantChoice = v5`
- `arm32v6`
  - `ArchitectureChoice = arm`
  - `VariantChoice = v6`
- `arm32v7`
  - `ArchitectureChoice = arm`
  - `VariantChoice = v7`
- `arm64v8` 
  - `ArchitectureChoice = arm64`
  - `VariantChoice = v8`

cclerget (changes requested):

cmd/internal/cli/pull.go
```diff
@@ -60,9 +59,9 @@ var (
 var pullArchFlag = cmdline.Flag{
 	ID:           "pullArchFlag",
 	Value:        &pullArch,
-	DefaultValue: runtime.GOARCH,
+	DefaultValue: "",
 	Name:         "arch",
-	Usage:        "architecture to pull from library",
+	Usage:        "architecture to pull, should be one of ('amd64', 'arm32v5', 'arm32v6', 'arm32v7', 'arm64v8', '386', 'ppc64le', 's390x')",
```
cclerget, line 64: This is shared with sylabs library, I think they are following naming based on GOARCH, so this usage message could be confusing depending on pulling from `docker://` or `library://`, maybe put the list for both in command long description here https://github.com/apptainer/apptainer/blob/e21533c26da455921b365b864f55dd78fd37bfc0/docs/content.go#L617

internal/pkg/build/oci/oci.go
```diff
@@ -150,18 +223,120 @@ func getRefDigest(ctx context.Context, ref types.ImageReference, sys *types.Syst
...
+		} else {
+			tmpArch = fmt.Sprintf("%sv%s", arch, archVariant)
+		}
+		// verification
+		if _, ok := ArchMap[tmpArch]; !ok {
+			return "", fmt.Errorf("arch: %s is not valid, supported archs are: %v, supported variants are [8], please remove --arch-variant option", tmpArch, supportedArchs)
+		}
+		return tmpArch, nil
+	case "arm":
+		if archVariant == "" {
```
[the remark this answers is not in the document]
JasonYangShadow: added in this commit
f3f3960501ed4c7f3558af77649b1b428c2213ed

cclerget (approved):

PR #1209, opened 2023-03-16
squashfuse_ll: do not use -o if optsStr is empty

posch:
## Description of the Pull Request (PR):

When apptainer invokes squashfuse_ll, it always includes the "-o" option. But if optsStr is empty, squashfuse_ll fails with:

> unknown option `'

This makes apptainer fail when used with squashfs images:

> $ apptainer shell container.sqfs 
> FATAL:   container creation failed: mount hook function failure: mount /proc/self/fd/3->/.../apptainer/mnt/session/rootfs error: while mounting image /proc/self/fd/3: squashfuse_ll exited with status 1: fuse: unknown option `

This patch removes '-o' if optsStr is empty. The squashfs image now works as expected:

> $ apptainer shell container.sqfs 
> Apptainer> 

### This fixes or addresses the following GitHub issues:

Maybe fixes #1037

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/apptainer/apptainer/blob/main/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTORS.md)

PR #1312, opened 2023-04-28
Is a warning message sufficient when staging /etc/resolv.conf fails?

posch:
I'm trying to use "apptainer build" to create containers inside the sandbox environment of the Nix package manager.

So far I've successfully been using mksquashfs to create the images for apptainer (see also #1209). But I think it may have advantages to use "apptainer build" instead to create the container from the file tree. It seems to work well, except for /etc/resolv.conf that is missing in the sandbox environment. This makes the Full() function fail and abort the build.

The attached patch makes this a non-fatal error. It's the only change needed to make the build succeed.

An alternative approach could be to make createStageFile use etc/resolv.conf from the sandbox tree or somewhere else if present. Not sure if that's better.

### DIFF
--- a/internal/pkg/build/build.go
+++ b/internal/pkg/build/build.go
@@ -411,7 +411,7 @@ func (b *Build) Full(ctx context.Context) error {
 		// create stage file for /etc/resolv.conf and /etc/hosts
 		sessionResolv, err := createStageFile("/etc/resolv.conf", stage.b, "Name resolution could fail")
 		if err != nil {
-			return err
+			sylog.Warningf("Failed to create stage file /etc/resolv.conf. Name resolution could fail!")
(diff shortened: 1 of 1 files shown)


posch:
> There is already a new `--no-mount` option in the main branch, getting ready for release in 1.2. It would make sense to extend that to `--no-mount /etc/resolv.conf`, if that doesn't already work.

Thanks for the feedback, but I'm not sure how to use --no-mount here.

"apptainer help build" doesn't list --no-mount. If I understand the correct correctly, --no-mount is only available for exec, shell, run, et al. Did you mean that --no-mount could be implemented for the "build" command, too?

DrDaveD:
Related to #1460.

DrDaveD:
Closing stale pull request.

PR #1357, opened 2023-05-25
fix overlay module not loaded under the root user leads to misleading error message

### LINKED ISSUE #1325: overlay module not loaded leads to misleading  error messages
### Version of Apptainer

```apptainer version 1.1.2-lp154.2.1 and apptainer version 1.1.7-1.el8```

### Expected behavior
After installing apptainer, the overlay kernel module should also be loaded.
In case the overlay module is not loaded, corresponding error messages should indicate this. 

### Actual behavior

Both SUSE 15.4 and Rocky 8 do not load the overlay kernel module after apptainer installation.
(zyper install apptainer or dnf install appatiner). Even a reboot will not load that module. 
as a consequence "--writable and writable-tmpfs" will not work.   

``` 
apptainer instance start --writable-tmpfs test.sif httpd
FATAL:   --writable-tmpfs requires overlay kernel support: your kernel doesn't support it
FATAL:   failed to start instance: while running /usr/lib/apptainer/bin/starter: exit status 255
```  
Create overlay inside sif file:
``` 
apptainer overlay create --size 1024 test.sif 
``` 
successful.

e.g. 
```
apptainer exec --writable test.sif /bin/sh
FATAL:   no SIF writable overlay partition found in /root/test.sif 
```
again create overlay:
```
apptainer overlay create --size 1024 open-on-demand.sif 
FATAL:   a writable overlay partition already exists in test.sif (ID: 5), delete it first with "apptainer sif del 5 test.sif"
 ```

simple 
```
modprobe overlay
```
will fix this. 

Rocky 8 or Suse minimal install.
Install apptainer using either dnf or zypper.

```sh
$ cat /etc/os-release

```
NAME="openSUSE Leap"
VERSION="15.4"
ID="opensuse-leap"
ID_LIKE="suse opensuse"
VERSION_ID="15.4"
PRETTY_NAME="openSUSE Leap 15.4"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:leap:15.4"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
DOCUMENTATION_URL="https://en.opensuse.org/Portal:Leap"
LOGO="distributor-logo-Leap"

NAME="Rocky Linux"
VERSION="8.7 (Green Obsidian)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Rocky Linux 8.7 (Green Obsidian)"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:rocky:rocky:8:GA"
HOME_URL="https://rockylinux.org/"
BUG_REPORT_URL="https://bugs.rockylinux.org/"
ROCKY_SUPPORT_PRODUCT="Rocky-Linux-8"
ROCKY_SUPPORT_PRODUCT_VERSION="8.7"
REDHAT_SUPPORT_PRODUCT="Rocky Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.7"

rpm source.
Installed Packages
Name         : apptainer
Version      : 1.1.7
Release      : 1.el8
Architecture : x86_64
Size         : 101 M
Source       : apptainer-1.1.7-1.el8.src.rpm
Repository   : @System
From repo    : epel
Summary      : Application and environment virtualization formerly known as Singularity
URL          : https://apptainer.org
License      : BSD and LBNL BSD and ASL 2.0
Description  : Apptainer provides functionality to make portable
             : containers that can be used across host environments

Information for package apptainer:
----------------------------------
Repository     : Main Update Repository
Name           : apptainer
Version        : 1.1.2-lp154.2.1
Arch           : x86_64
Vendor         : openSUSE
Installed Size : 101.3 MiB
Installed      : Yes
Status         : up-to-date
Source package : apptainer-1.1.2-lp154.2.1.src
Upstream URL   : https://apptainer.org
Summary        : Application and environment virtualization
Description    : 
    Singularity provides functionality to make portable
    containers that can be used across host environments.

JasonYangShadow:

fix overlay module not loaded under the root user leads to misleading error message

 - Fixes #1325 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
   binding fakeroot into container during apptainer startup for --fakeroot
   with fakeroot command.
 - Fix memory usage calculation during apptainer compilation on RaspberryPi.
+- Fix misleading error when an overlay is requested by the root user while the
+  overlay kernel module is not loaded.
 
 ## v1.1.8 - \[2023-04-25\]
 
--- a/internal/pkg/runtime/launch/launcher_linux.go
+++ b/internal/pkg/runtime/launch/launcher_linux.go
@@ -389,8 +389,14 @@ func (l *Launcher) Exec(ctx context.Context, image string, args []string, instan
 	}
 
 	loadOverlay := false
-	if !l.cfg.Namespaces.User && buildcfg.APPTAINER_SUID_INSTALL == 1 {
-		loadOverlay = true
+	if !l.cfg.Namespaces.User && (buildcfg.APPTAINER_SUID_INSTALL == 1 || os.Getuid() == 0) {
+		has, err := proc.HasFilesystem("overlay")
+		if err != nil {
+			return fmt.Errorf("while checking whether overlay filesystem is loaded: %w", err)
+		}
+		if !has {
+			loadOverlay = true
+		}
 	}
 
 	cfg := &config.Common{


JasonYangShadow:
Test

```
root@ubuntu:/home/vagrant/test# modprobe -r overlay
root@ubuntu:/home/vagrant/test# cat /proc/filesystems | grep overlay
root@ubuntu:/home/vagrant/test# apptainer exec --writable ubuntu.sif /bin/sh
Apptainer> exit
root@ubuntu:/home/vagrant/test# 
```

JasonYangShadow:
code here will load overlay kernel driver by force by default
https://github.com/JasonYangShadow/apptainer/blob/e23e92f6bc159b0bd2871462fcc6cda2568136b0/cmd/starter/c/starter.c#L1254
so we just need to skip the original strict check.

internal/pkg/runtime/engine/apptainer/prepare_linux.go
```diff
@@ -1050,6 +1050,12 @@ func (e *EngineOperations) setSessionLayer(img *image.Image) error {
 		return nil
 	}
 
+	if e.EngineConfig.File.EnableOverlay == "try" && os.Getuid() == 0 && os.Geteuid() == 0 && (writableImage && hasSIFOverlay) {
+		sylog.Debugf("Trying overlay because we are the root user")
+		e.EngineConfig.SetSessionLayer(apptainerConfig.OverlayLayer)
+		return nil
+	}
```
DrDaveD, line 1057: That's changing the meaning of the `enable overlay = try` option, and I don't like that very much.  That was only intended to apply to enabling overlay for bind mounts, it doesn't have anything to do with other overlay features.  That's what the description in apptainer.conf says.  On the other hand, further down in the function it is disallowing overlay of other kinds if that value is set to `no`.  In any case, you're not checking for `yes` so that's not right.

I also don't think that there's any need to explicitly check `os.Geteuid() == 0` when `os.Getuid() == 0` is already checked.

Also I don't think that `(writableImage && hasSIFOverlay)` is the right combination of conditions because there may be other conditions where we have the problem, and this is probably not the best place to insert the check either ... it seems to me that the better place is down lower in place of the check for the "overlay" entry in /proc/filesystems.   I suggest changing the `if has..` line and the following line to be something like this:
```
hasoverlay, _ := proc.HasFilesystem("overlay");
if hasoverlay {
        sylog.Debugf("Overlay seems supported and allowed by kernel")
} else if os.Getuid() == 0 {
        // it's probably missing because a modprobe equivalent is needed, which is done later
        sylog.Debugf("Overlay module not loaded but proceeding because we are the root user")
        hasoverlay = true
}
if hasoverlay {
```
It could be argued that this also isn't the best because the root user is restricted by the configuration and we often allow bypassing it, but that's no different than the behavior that was there, and the default settings allow it.

CHANGELOG.md
```diff
@@ -90,6 +90,7 @@ For older changes see the [archived Singularity change log](https://github.com/a
   binding fakeroot into container during apptainer startup for --fakeroot
   with fakeroot command.
 - Fix memory usage calculation during apptainer compilation on RaspberryPi.
+- Fix overlay module not loaded under the root user leads to misleading error message.
```
DrDaveD, line 93: Improved wording:
```
- Fix misleading error when an overlay is requested by the root user while the overlay kernel module is not loaded.
```

DrDaveD (changes requested):

JasonYangShadow:
debug_log
```
DEBUG   [U=0,P=147239]     Init()                        sif image format detected
DEBUG   [U=0,P=147239]     setSessionLayer()             Overlay module not loaded but proceeding because we are the root user
DEBUG   [U=0,P=147239]     setSessionLayer()             Not attempting to use overlay or underlay: writable flag requested
DEBUG   [U=0,P=147239]     PrepareConfig()               image driver is fuseapps
```

```
root@ubuntu:/home/vagrant/test# modprobe -r overlay
root@ubuntu:/home/vagrant/test# cat /proc/filesystems | grep overlay
root@ubuntu:/home/vagrant/test# apptainer exec --writable ubuntu.sif /bin/sh
Apptainer> exit
root@ubuntu:/home/vagrant/test# 
```

internal/pkg/runtime/engine/apptainer/prepare_linux.go
JasonYangShadow: Thanks for the great comment! I totally agree that the code should be changed at the place where overlay file system is checked. 👍🏽

CHANGELOG.md
JasonYangShadow: thanks for the comment!

DrDaveD (changes requested):

internal/pkg/runtime/engine/apptainer/prepare_linux.go
```diff
@@ -1130,6 +1130,9 @@ func (e *EngineOperations) setSessionLayer(img *image.Image) error {
 			}
 			sylog.Debugf("Could not use overlay, disabled by configuration ('enable overlay = no')")
 		}
+	} else if os.Getuid() == 0 {
+		sylog.Debugf("Overlay module not loaded but proceeding because we are the root user")
+		e.EngineConfig.SetSessionLayer(apptainerConfig.OverlayLayer)
```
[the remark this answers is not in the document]
JasonYangShadow: I added return in this commit
https://github.com/apptainer/apptainer/pull/1357/commits/6f54d3dbe5331351fe7590a1c226831287664334

cclerget (approved):

Issue #665, opened 2022-08-29
squashfuse Performance


hollowec:
### Version of Apptainer

```
$ apptainer --version
apptainer version 1.1.0~rc.2-1.el7
```

### Expected behavior

When using SIF images with unprivileged Apptainer, execution time should be similar to unprivileged Singularity.

### Actual behavior

Apptainer's move to squashfuse for unprivileged (user namespace) mounts of SIF images has significantly increased the execution time of some containers, compared to automatically unpacking SIF images to a temporary sandbox as unprivileged Singularity did.  I believe this is primarily a concern for containers running multiple processes/threads, as it seems there is a single squashfuse process to handle all of the parallel I/O requests and decompression.

### Steps to reproduce this behavior
`apptainer run -i -c -e -B /tmp/atlasgen:/results -B /tmp docker://gitlab-registry.cern.ch/hep-benchmarks/hep-workloads/atlas-gen-bmk:v2.1 -W --threads 1 --events 200 `
This is an ATLAS event generation benchmark container that will run a process per logical core on the host.  Execution times on a system with 2x AMD EPYC 7351 CPUs (64 logical cores total):  
Singularity with user namespaces (unpack to sandbox)
        Execution time: ~24 min

Apptainer with setuid (squashfs privileged mount)
        Execution time: ~25 min

Apptainer with user namespaces (squashfuse mount)
        Execution time: **~2 hours 50 minutes**

During execution, I see the squashfuse process using 100% of a single CPU core during most of the run.

Ideally the default behavior would be to revert to automatically unpacking SIF images when used unprivileged.

### What OS/distro are you running

Scientific Linux 7

### How did you install Apptainer

RPM from EPEL testing repo.

DrDaveD:
Thanks so much for this report and the details on your benchmark!

Indeed I was able to reproduce the issue and run many of my own measurements using your benchmark.  I have access to 16-core nodes that have both local disk and lustre.  They have dual 2.6 GHz Intel E5-2650v2 CPUs and 128G of RAM.  Because I had 16 cores I ran my tests with 32 events instead of 200, and I always pre-converted the container to the format being tested.  I don't have access to a setuid-root installation on the machine so couldn't measure using kernel squashfs (hopefully I can get a sysadmin to cooperate for a test later).  I included testing the image from cvmfs at
```
/cvmfs/unpacked.cern.ch/gitlab-registry.cern.ch/hep-benchmarks/hep-workloads/atlas-sim-bmk:v2.1
[...]

PR #1359, opened 2023-05-25
Package gocryptfs with apptainer

### LINKED ISSUE #1285: Package gocryptfs
Include gocryptfs in the packaging and INSTALL.md, as a followup to #1185.

DrDaveD:
This PR adds gocryptfs to apptainer rpm and debian packages and to the installation instructions. It also adds a test of encryption and decryption to the ci check for unprivileged installations, which uses the rpm it builds as input.

- Fixes #1285

Unfortunately I found that gocryptfs, unlike fuse packages based on fuse-libs, [requires](https://github.com/rfjakob/gocryptfs/pull/749) using the `fusermount` command even when run in a root-mapped unprivileged user namespace.  At least it does not depend on the setuid-root functionality of fusermount, but it does require the fuse (or fuse3 package on Debian) package to be installed.

I also re-ran the benchmark in #665 on a gocryptfs-encrypted SIF file and got almost identical results, an average of 6:28 on two runs.  That was using apptainer installed by install-unprivileged.sh with an rpm as input that was built with this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,7 +59,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
   flag.
 - Support for online verification checks of x509 certificates using OCSP protocol.
   (introduced flag: `verify --ocsp-verify`)
-- Support for unprivileged encryption of SIF files using gocryptfs.
+- Support for unprivileged encryption of SIF files using gocryptfs.  The
+  gocryptfs command is included in rpm and debian packaging.
 - The `instance start` command now accepts an optional `--app <name>` argument which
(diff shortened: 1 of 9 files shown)


cclerget (approved):

DrDaveD:
For the record: I see I had not recorded the measurements for the option of using only gocryptfs on top of squashfuse_ll, instead of squashfuse_ll + gocryptfs + squashfuse_ll that the Apptainer 1.2.0 implementation ended up with.  I just ran the same benchmark mentioned in the description above again with only the two layers, and an average of two runs was 10:06, over 50% slower.

Issue #1336, opened 2023-05-12
Inspecting container definition file drops characters from build file


dcat52:
### Version of Apptainer

```bash
davis@host:~/tmp$ apptainer --version
apptainer version 1.1.5
davis@host:~/tmp$ sudo apptainer --version
apptainer version 1.1.5
davis@host:~/tmp$ sudo singularity --version
singularity version 3.6.2
```

### Expected behavior

```
$ sudo singularity inspect -d tmp.sif
Bootstrap: docker
From: ubuntu:16.04

%post
  # pre-reqs
  apt-get update -q -y
  apt-get install -q -y \
    build-essential \
    bzip2 \
    cmake \
    cmake-curses-gui \
    dnsutils \
    emacs-nox \
    git net-tools \
    python
```

### Actual behavior

Notice the dropped `\` from the end of the lines in the `%post`. Also the loss of the letter `n` from `python` in the `%post`.

```bash
$ sudo apptainer inspect -d tmp.sif
INFO:    /usr/local/etc/singularity/ exists; cleanup by system administrator is not complete (see https://apptainer.org/docs/admin/latest/singularity_migration.html)
Bootstrap: docker
From: ubuntu:16.04

%post
  # pre-reqs
  apt-get update -q -y
  apt-get install -q -y \
    build-essential \
    bzip2 
    cmake 
    cmake-curses-gui 
    dnsutils \
    emacs-nox \
    git net-tools \
    pytho
 ```

### Steps to reproduce this behavior

Using the following original definition file, build a container: `sudo apptainer build tmp.sif tmp.def` then inspect it `sudo apptainer inspect -d tmp.sif`
```bash
$ cat tmp.def 
Bootstrap: docker
From: ubuntu:16.04

%post
  # pre-reqs
  apt-get update -q -y
  apt-get install -q -y \
    build-essential \
    bzip2 \
    cmake \
    cmake-curses-gui \
    dnsutils \
    emacs-nox \
    git net-tools \
    python
 ```

### What OS/distro are you running

```sh
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.2 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.2 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
[...]

PR #1375, opened 2023-06-01
more robust way of setting IFS to newline, from sylabs 1708

edytuk:
This pulls in sylabs PR

- sylabs/singularity#1708
 which fixed
- sylabs/singularity#1704
and fixes
- apptainer/apptainer#1336

> As part of the function of `singularity inspect`, an ad-hoc shell script is set up, by concatenating programmatically-generated snippets of shell code. One of these snippets, created in cmd/internal/cli/inspect.go:227, includes a line whose function is to set IFS (the Input Field Separator) _back_ to newline in the scope of the `cat_file()` function, despite being set to ":" (colon) in the global scope (cmd/internal/cli/inspect.go:256).
> 
> However, the way this was done until now uses escaping of the newline, as follows: `local IFS=$'\n'` This escaping apparently doesn't work on certain platforms, resulting in the relevant input being split using the character "n" as the relevant delimiter, rather than using newlines.
> 
> I have verified that it fails on Debian 11 "bullseye" on amd64, and the user who reported sylabs/singularity#1704 encountered the same thing on Ubuntu 22.04 "jammy" on amd64. There is a newer syntax for this - `local IFS=$'...'` - but there is a worry this won't work on older systems; see discussion in [this](https://stackoverflow.com/questions/16831429/when-setting-ifs-to-split-on-newlines-why-is-it-necessary-to-include-a-backspac) StackOverflow post. In the same post, however, it is mentioned that there is a rather simple and possibly more robust way to achieve the same effect, namely, to use a literal newline between double-quotes:
> 
> ```
>       <clipped>
>       local IFS="
> "
>       <clipped>
> ```
> 
> This PR replaces the previous implementation, using `$'\n'`, with the one above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,6 +114,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Remove warning about unknown `xino=on` option from fuse-overlayfs,
   introduced in 1.1.8.
 - Ignore extraneous warning from fuse-overlayfs about a readonly `/proc`.
+- Fix dropped "n" characters on some platforms in definition file stored as part
+  of SIF metadata.
 
 ## v1.1.8 - \[2023-04-25\]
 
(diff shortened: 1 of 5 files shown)


DrDaveD:
I rebased this on main to fix one e2e faliure but now this seems to be failing the newly added e2e regression test.  Perhaps it is a case matching issue?  @edytuk please check.

PR #1381, opened 2023-06-04
Instance run

### LINKED ISSUE #959: better support for startscript.sh (feature request)
### Version of Apptainer

```
$ apptainer --version
apptainer version 1.1.3
```

### Expected behavior

I'd like to run `apptainer instance start docker://some/container` or similar and get the entrypoint to run as the startscript.

### Actual behavior

afaik, there is no good way to get the container to recognize the docker entrypoint as the startscript.  Instead, apptainer shoves all of the entrypoint stuff into the runscript by default and if you run the container as an instance it does not execute the entrypoint.  

I've tried a few different workarounds, but so far nothing is ideal.  One idea is to force the container to execute the runscript when the instance is started.  I tried this by creating a wrapper script (on the host system) like so:

```
echo "/bin/sh /.singularity.d/runscript" >startscript

chmod 750 startscript
```
Then bind mounting it into the metadata directory at runtime like so:

```
export APPTAINER_BINDPATH=startscript:/.singularity.d/startscript
```
This seems to work, but it is very brittle and there are potential side effects for running other containers. 

The second thing I tried was to rebuild the container with a def file of the following format.

```
Bootstrap: docker                                                               
From: some/container                                                
                                                                                   
%post                                                                           
    cp /.singularity.d/runscript /.singularity.d/startscript
```

This also works, but it produces an intermediate build artifact that may be unwanted.  

### Solution?

I think some discussion is needed about the correct way to approach this.  Maybe the user could pass a flag to `instance start` command to let it know it should treat the runscript as the startscript.  Or maybe Apptainer should intelligently copy the docker CMD and entrypoint stuff to the startscript instead of the runscript if the user is executing the container as an instance.  This second option would probably be a sensible default, but it would also be a breaking change if folks are currently using instance start on OCI containers.  

What say the developers?

GodloveD:

This adds a new command `run` to the `instance` group.  This command does exactly the same thing as `start` except it executes the runscript instead of the startscript.  This should enable more compatibility with OCI containers.   

 - Fixes #959


GodloveD:
Deprecated by #1396

GodloveD:
Both deprecated by #1401

PR #1396, opened 2023-06-06
Instance run 

### LINKED ISSUE #959: better support for startscript.sh (feature request)

GodloveD:

This adds a new command `run` to the `instance` group.  This command does exactly the same thing as `start` except it executes the runscript instead of the startscript.  This should enable more compatibility with OCI containers.   

 - Fixes #959 
 - Deprecates PR #1381 (which was quite messy)


cmd/internal/cli/instance_actions_linux.go
```diff
@@ -37,35 +37,56 @@ var instanceStartPidFileFlag = cmdline.Flag{
...
+	name := args[1]
+	cmdName := cmd.Name()
+	script := "start"
+	killCont := ""
+
+	if cmdName == "run" {
+		script = "run"
+		killCont = "/bin/sh -c kill -CONT 1; "
+	}
+	a := append([]string{killCont + "/.singularity.d/actions/" + script}, args[2:]...)
```
DrDaveD, line 52: I'm surprised to not see quotes around the full parameter to `-c`.  This works?

DrDaveD (changes requested):

cmd/internal/cli/instance_actions_linux.go
GodloveD: hmm.  I guess it probably works because it is just a single command.  But quoting would be better.   I'll fix.

GodloveD: Actually I think that there is some quoting going on at another stage in the go code because if I add the quotes it treats them as literals and I get errors like:
```
ERROR:   Unknown action run"
```
or 
```
ERROR:   Unknown action run'
```
It looks like the PID 1 is getting cleaned up when I stop the container:
```
$ apptainer instance list
INSTANCE NAME    PID        IP    IMAGE
acow             1834314          /home/user/.apptainer/cache/oci-tmp/4830a4faf1280161fc1ead7a241ac4dc0831f77a3c0cf26466d94c35c53ea84d

$ apptainer shell instance://acow

Apptainer> ps -ef
UID          PID    PPID  C STIME TTY          TIME CMD
user           1       0  0 07:26 ?        00:00:00 appinit
user          19       0  0 07:26 pts/0    00:00:00 /bin/bash --norc
user          26      19  0 07:26 pts/0    00:00:00 ps -ef
Apptainer> exit
exit

$ ps -ef | grep appinit
user 1834314 1834313  0 07:26 ?        00:00:00 appinit
user 1834444   11341  0 07:26 pts/0    00:00:00 grep --color appinit

$ apptainer instance stop --all
INFO:    Stopping acow instance of /home/user/.apptainer/cache/oci-tmp/4830a4faf1280161fc1ead7a241ac4dc0831f77a3c0cf26466d94c35c53ea84d (PID=1834314)

$ ps -ef | grep appinit
user 1834502   11341  0 07:27 pts/0    00:00:00 grep --color appinit
```
It also cleans up the `appinit` process if I kill the container by PID or if I kill the parent `starter` process, so I think it is all good?

internal/pkg/util/fs/files/action_script.sh
```diff
@@ -222,6 +222,21 @@ start)
...
+            exec "/scif/apps/singularity/scif/test" "$@"
+        fi
+        sylog error "No runscript for contained app: ${SINGULARITY_APPNAME:-}"
+        exit 1
+    elif test -x "/.singularity.d/runscript"; then
+        exec "/.singularity.d/runscript" "$@"
+    fi
+
+    sylog info "No instance run runscript found in container"
+    exit 0 ;;
```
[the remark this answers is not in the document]
GodloveD: Whoops, no.  I'll fix that up.

GodloveD:
deprecated by #1401

PR #1440, opened 2023-06-15
Improved discussion of LICENSE_DEPENDENCIES.md in CONTRIBUTING.md, from sylabs 1393

edytuk:
This pulls in sylabs PR

- sylabs/singularity#1393
 which fixed
- sylabs/singularity#1392

> Addresses two problems in `CONTRIBUTING.md`, specifically in the discussion of `LICENSE_DEPENDENCIES.md`:
> 
> * Replaced use of the phrase "Any new dependency" with the hopefully clearer "Any new or updated dependency". (The former could be misinterpreted to mean "only wholly new dependencies" excluding, e.g., version bumps, and that would be incorrect.)
> * Fixed typo in the name of the update script: its extension is `.sh` but was given as `.md`

### DIFF
--- a/e2e/cache/cache.go
+++ b/e2e/cache/cache.go
@@ -293,11 +293,11 @@ func (c cacheTests) testMultipleArch(t *testing.T) {
 	)
 
 	shamap := map[string]string{
-		"arm64":      "028a3acf50e5ffcd0a9ffc03bce653cc8e4c02d2806c948a48d7bd78136b7372",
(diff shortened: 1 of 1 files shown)


DrDaveD:
I had to fix not one, not two, but three (!) ci check failures here on things unrelated to the change.  They were all due to changes in docker containers that the ci checks depended on.

DrDaveD (approved):

PR #1494, opened 2023-06-28
Fix 'enable overlay = driver' to always force using image driver

### LINKED ISSUE #1464: "enable overlay = driver" option does not force the use of fuse-overlayfs
At least in version 1.1.9, setting "enable overlay = driver" in /etc/apptainer/apptainer.conf does not force the use of fuse-overlayfs as I expected.

It would also be good to update the configuration file comments ahead of "image driver" to mention that setting it will override the internal image driver used for unprivileged mounting.

DrDaveD:
This makes the `enable overlay = driver` apptainer.conf option work like it was commented, to always use the image driver for overlayfs when running unprivileged, even if overlay is available.  It also updates the `image driver` comments in apptainer.conf to more accurately reflect what the builtin image driver is doing.

- Fixes #1464


internal/pkg/runtime/engine/apptainer/container_linux.go
```diff
@@ -665,34 +665,38 @@ mount:
...
-		//  Unfortunately this also affects read-only fuse2fs
-		//  overlays even though no problems had been seen with
-		//  those using the kernel overlayfs.
-
-		for _, ldir := range strings.Split(lowerdirs, ":") {
-			err = fsoverlay.CheckFuse(ldir)
-			if err != nil {
-				break
+		if c.engine.EngineConfig.File.EnableOverlay == "driver" {
+			err = fmt.Errorf("overlay image driver selected by configuration")
```
cclerget, line 669: can you put a comment like `// set the error to switch to the overlay image driver below during the mount error check`  as this is not an obvious code path

cclerget (changes requested):

PR #1503, opened 2023-06-30
fix rel. path --workdir with --scratch, add native e2e tests, from sylabs 1694

edytuk:
This pulls in sylabs PR

- sylabs/singularity#1694
 which fixed
- sylabs/singularity#1693

> Fixes a bug where `--scratch` failed when used in conjunction with `--workdir` if the latter was given a relative (rather than absolute) path.
> 
> Also changes the behavior of `--workdir` code so that an error (rather than a warning) is generated if specified workdir cannot be converted to an absolute path.
> 
> Lastly, adds e2e tests for this behavior (`--scratch` in conjunction with relative `--workdir`), in both native and OCI modes.


cclerget (approved):

DrDaveD:
Please cherry-pick to release-1.2

PR #1532, opened 2023-07-05
Include Sylabs changes to build templates (release-1.2)

edytuk:
This cherry-picks #1525 to the release-1.2 branch


DrDaveD:
We'll just leave this for the main branch because it depends pretty heavily on #1510 which is only in the main branch.

PR #1548, opened 2023-07-10
replace build-args Transformer with full-text replacement, from sylabs 1831

edytuk:
This pulls in sylabs PR

- sylabs/singularity#1831

> In sylabs/singularity#1738, the `transform.Transformer` object used to replace build-args (e.g. `"{{ argname }}"`) to enable templating in definition files (currently in internal/pkg/build/build.go) could encounter _partial input_ scenarios - where the bytes it has read contain the beginning of a build-arg, but it gets clipped due to the block size (e.g. `"{{ argname }"`, `"{{ argna"`, `"{"`, etc.).
> 
> The `transform.Transformer` [API](https://pkg.go.dev/golang.org/x/text/transform#Transformer) is not well-suited to facilitate a workaround for this type of problem. That's because `transform.Transformer` is fundamentally built for byte-by-byte transformations, rather than transformations with significantly richer state.
> 
> To correct this problem, this PR replaces the previous implementation of build-args templating - which used `transform.NewReader()` to wrap the underlying definition-file Reader in a custom `transform.Transformer` - with an implementation that simply reads in the full text of the definition file, wraps the modified text in a new Reader, and passes that to the code that builds `types.Definition` objects.


JasonYangShadow (approved):
LGTM!

PR #1561, opened 2023-07-17
Revamp passwd utility function, from sylabs 1597 and 1822

edytuk:
This pulls in sylabs PRs

- sylabs/singularity#1597
- sylabs/singularity#1822
-  which fixed
- sylabs/singularity#1812

> _NOTE: This PR is a continuation of sylabs/singularity#1531. See additional discussion there._
> 
> The way the Passwd() utility-function (internal/pkg/util/fs/files/passwd.go:22) worked until now, it was only able to append a new line to what it got as input (from the base layer).
> 
> This meant that, for example, to override the homedir of a regular user, we would simply add a modified line (with the custom homedir) to the in-container /etc/passwd; but to override the homedir of root (when running singularity itself as root, or when running with `--fakeroot`), we couldn't do that – since there was already a "root" line in the base layer's /etc/passwd – and we resorted to an entirely different method, namely, overriding the `$HOME` environment variable upon container initialization.
> 
> In addition to being non-uniform, the approach involving environment variables can also give rise to an internally inconsistent state, where, when we're root inside the container, our $HOME is set to one thing (a custom homedir) but the /etc/passwd file still has `/root` as the homedir for the root user.
> 
> A more uniform approach, which also avoids this inconsistent state described above, is possible: change the Passwd() utility-function so that when "adding" a line to /etc/passwd it first checks if a line with the corresponding uid already exists, and if it does, replaces that specific line instead of appending a new line to the file.
> ### Performance considerations
> 
> The implementation in this PR involves calling `ParsePasswdLine()` from the [github.com/stat0s2p/etcpwdparse](https://pkg.go.dev/github.com/stat0s2p/etcpwdparse) package on every line in the /etc/passwd file (internal/pkg/util/fs/files/passwd.go:52).
> 
> If this turns out to be problematic from a performance perspective, we can replace it with strings.split or something similar to just pull out the first field of each /etc/passwd line (=the username), and using only that to judge sameness-of-users for the purposes of deciding whether to replace a line or append a new one.
> ### Related PRs
> 
> Bug sylabs/singularity#1529 was a failure to override root's homedir when running in OCI mode as root/fakeroot. It was fixed in PR sylabs/singularity#1530 using the environment-variables method described above, in order to maintain uniformity with the native-runtime mode, and in order to not have the fix depend on the redoing of Passwd() described here.

> If all of the following conditions hold —
> 
> * The user inside the container is root
> * The root filesystem is writable
> * The root filesystem is not a tmpfs
> 
> — then writes to `/etc/passwd` or `/etc/group` are possible, and would be expected to persist from build to action stage (run/exec/shell) or across distinct actions (e.g. from one run to the next).
> 
> Under this exact combination of conditions, then, we inhibit bind-mounting modified versions of `/etc/passwd` and `/etc/group` from tmpfs storage into the container.


DrDaveD:
Looks good to me, but I would like @cclerget to also review.

cclerget (approved):

PR #1569, opened 2023-07-20
Fix the issue that APPTAINER_MESSAGELEVEL does not have effect

### LINKED ISSUE #1556: APPTAINER_MESSAGELEVEL doesn't have much effect
At least on the main branch, there's an APPPTAINER_MESSAGELEVEL environment variable referenced in pkg/sylog/sylog.go but setting the variable does not have much effect. Reproduce by setting `APPTAINER_MESSAGELEVEL=5` which should put it into debug mode.  A couple of debug messages come out, and then it stops.

JasonYangShadow:

This `APPTAINER_MESSAGELEVEL` is only used when sylog is initialized, so we need to do some calculation between log level envs and message level env. 
https://github.com/JasonYangShadow/apptainer/blob/issue%2F1556/cmd/internal/cli/apptainer.go#L275
https://github.com/JasonYangShadow/apptainer/blob/issue%2F1556/cmd/internal/cli/apptainer.go#L291

 - Fixes #1556 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,12 @@ For older changes see the [archived Singularity change log](https://github.com/a
   pkg/build/types/parser.All(), `.Raw` contains the raw content of a single
   build stage. Otherwise, it is equal to `.FullRaw`.
 
+## Changes since last release
+
+### Bug fixes
+
+- Fix `$APPTAINER_MESSAGELEVEL` to correctly set the logging level.
+
 ## v1.2.0 - \[2023-07-18\]
 
 Changes since v1.1.9
--- a/cmd/internal/cli/apptainer.go
+++ b/cmd/internal/cli/apptainer.go
@@ -15,11 +15,13 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"os/exec"
 	"os/signal"
 	"os/user"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"text/template"
 
@@ -271,18 +273,26 @@ func addCmdInit(cmdInit func(*cmdline.CommandManager)) {
 func setSylogMessageLevel() {
 	var level int
 
+	l, err := strconv.Atoi(env.GetenvLegacy("MESSAGELEVEL", "MESSAGELEVEL"))
+	if err == nil {
+		level = l
+	}
+
 	if debug {
 		level = 5
 		// Propagate debug flag to nested `apptainer` calls.
 		os.Setenv("APPTAINER_DEBUG", "1")
 	} else if verbose {
 		level = 4
+		os.Setenv("APPTAINER_VERBOSE", "1")
 	} else if quiet {
 		level = -1
+		os.Setenv("APPTAINER_QUIET", "1")
 	} else if silent {
 		level = -3
+		os.Setenv("APPTAINER_SILENT", "1")
 	} else {
-		level = 1
+		level = int(math.Max(float64(level), 1))
 	}
 
 	color := true
--- a/pkg/sylog/sylog.go
+++ b/pkg/sylog/sylog.go
@@ -20,8 +20,6 @@ import (
 	"strings"
 )
 
-const messageLevelEnv = "APPTAINER_MESSAGELEVEL"
-
 var messageColors = map[messageLevel]string{
 	FatalLevel: "\x1b[31m",
 	ErrorLevel: "\x1b[31m",
@@ -37,9 +35,9 @@ var (
 var logWriter = (io.Writer)(os.Stderr)
 
 func init() {
-	level, err := strconv.Atoi(os.Getenv(messageLevelEnv))
+	l, err := strconv.Atoi(os.Getenv("APPTAINER_MESSAGELEVEL"))
 	if err == nil {
-		loggerLevel = messageLevel(level)
+		loggerLevel = messageLevel(l)
 	}
 }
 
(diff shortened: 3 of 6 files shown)


cmd/internal/cli/apptainer.go
```diff
@@ -282,7 +288,7 @@ func setSylogMessageLevel() {
 	} else if silent {
 		level = -3
 	} else {
-		level = 1
+		level = sylog.Max(level, 1)
```
DrDaveD, line 291: Use `math.Max()` instead and delete the new function below.

pkg/sylog/sylog.go
```diff
@@ -20,7 +20,7 @@ import (
 	"strings"
 )
 
-const messageLevelEnv = "APPTAINER_MESSAGELEVEL"
+const MessageLevelEnv = "APPTAINER_MESSAGELEVEL"
```
DrDaveD, line 23: Instead of making this global, I think it would be better to have a function to get the message level.

In fact, for almost all environment variables we support either a SINGULARITY_ or APPTAINER_ prefix.  If it's not difficult to use the common function where that's done that would be better, even though I know it wasn't being done before.  For the test case where this variable is set I think it's fine to use the string `APPTAINER_MESSAGELEVEL` rather than a global variable name.

DrDaveD (changes requested):
Don't forget a CHANGELOG entry.  This is a bug fix so it will be backported to release-1.2.  That means there should be a new `Changes since last release` section just above the v1.2.0 release, it shouldn't join all the existing changes in the CHANGELOG main branch which are for 1.3.

JasonYangShadow:
@DrDaveD 
hmm, find something that is a bit of inconsistency. 
here, we are setting both `APPTAINER_MESSAGELEVEL` and `--slient`

Though it is reasonable because until setSylogMessageLevel is called, the message level is changed. So any debug messages outside of that will be printed out. 

```
vagrant@ubuntu:~/test$ apptainer --silent build --force 1.sif docker://alpine:latest
DEBUG   [U=1000,P=63024]   func1()                       Install prefix is /usr/local
DEBUG   [U=1000,P=63024]   func1()                       Using apptainer directory "/home/vagrant/.apptainer"
DEBUG   [U=1000,P=63101]   func1()                       Install prefix is /usr/local
DEBUG   [U=1000,P=63101]   func1()                       Using apptainer directory "/home/vagrant/.apptainer"
DEBUG   [U=1000,P=63101]   setValue()                    Updated flag 'silent' value to: true
vagrant@ubuntu:~/test$ echo $APPTAINER_MESSAGELEVEL
5
```

cmd/internal/cli/apptainer.go
JasonYangShadow: thanks! reflected in this commit
https://github.com/apptainer/apptainer/pull/1569/commits/1a2003474ceaa5144f602b234aa4bd811f3d336b

pkg/sylog/sylog.go
JasonYangShadow: just using pure string instead
https://github.com/apptainer/apptainer/pull/1569/commits/1a2003474ceaa5144f602b234aa4bd811f3d336b

DrDaveD: How about using nternal/pkg/util/env/env.go's GetenvLegacy to get the MESSAGELEVEL?

DrDaveD:
> @DrDaveD hmm, find something that is a bit of inconsistency. here, we are setting both `APPTAINER_MESSAGELEVEL` and `--slient`

I don't think we really need to worry about the case of somebody using inconsistent options.

pkg/sylog/sylog.go
JasonYangShadow: reflected in this commit
https://github.com/apptainer/apptainer/pull/1569/commits/7f8566c352d1a5affa0cfdbef11bc61a0c9130b6
sylog/sylog_dummy could not use GetenvLegacy due to the cycle import.

CHANGELOG.md
```diff
@@ -27,6 +27,12 @@ For older changes see the [archived Singularity change log](https://github.com/a
   pkg/build/types/parser.All(), `.Raw` contains the raw content of a single
   build stage. Otherwise, it is equal to `.FullRaw`.
 
+## Changes since last release
+
+### Bug fixes
+
+- Fix the issue that `APPTAINER_MESSAGELEVEL` does not have effect.
```
DrDaveD, line 34: Please change to
```
- Fix `$APPTAINER_MESSAGELEVEL` to correctly set the logging level.
```

DrDaveD (changes requested):

pkg/sylog/sylog_dummy.go
```diff
@@ -14,13 +14,23 @@ package sylog
 import (
 	"io"
 	"os"
+	"strconv"
 )
 
+const messageLevelEnv = "APPTAINER_MESSAGELEVEL"
```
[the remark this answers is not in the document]
JasonYangShadow: thanks for the comment! fixed in this commit
https://github.com/apptainer/apptainer/pull/1569/commits/04ffbdbffa3c88229578e5befccb5b62be3198d0

CHANGELOG.md
JasonYangShadow: thanks for the comment! fixed in this commit
https://github.com/apptainer/apptainer/pull/1569/commits/04ffbdbffa3c88229578e5befccb5b62be3198d0

DrDaveD (approved):

PR #1587, opened 2023-07-27
add progress bar support for apptainer push/pull for oras protocol

### LINKED ISSUE #437: Feature request: progress indicator during `apptainer push`
Apptainer should support an optional progress indicator during push, with percentage complete as well as bytes transferred / total bytes to transfer.

JasonYangShadow:

add progress bar support for `apptainer push/pull` for oras protocol

 - Fixes #437 

### Apptainer Push
push

### Apptainer Pull
pull

### DIFF
--- a/internal/pkg/client/oras/oras.go
+++ b/internal/pkg/client/oras/oras.go
@@ -86,7 +86,7 @@ func newOrasUploadTransport() http.RoundTripper {
 	}
 }
 
-func getResolver(ctx context.Context, ociAuth *ocitypes.DockerAuthConfig, noHTTPS, push bool) (remotes.Resolver, error) {
+func getResolver(ctx context.Context, ociAuth *ocitypes.DockerAuthConfig, noHTTPS, push, progressBar bool) (remotes.Resolver, error) {
 	opts := docker.ResolverOptions{Credentials: genCredfn(ociAuth), PlainHTTP: noHTTPS}
 	if ociAuth != nil && (ociAuth.Username != "" || ociAuth.Password != "") {
 		return docker.NewResolver(opts), nil
@@ -109,7 +109,16 @@ func getResolver(ctx context.Context, ociAuth *ocitypes.DockerAuthConfig, noHTTP
 		httpClient.Transport = newOrasUploadTransport()
 	}
 
-	return cli.Resolver(ctx, httpClient, noHTTPS)
+	solver, err := cli.Resolver(ctx, httpClient, noHTTPS)
+	if err != nil {
+		return solver, err
+	}
+
+	if progressBar {
+		return &resolver{solver}, nil
+	}
+
+	return solver, nil
 }
 
 // DownloadImage downloads a SIF image specified by an oci reference to a file using the included credentials
@@ -128,7 +137,7 @@ func DownloadImage(ctx context.Context, imagePath, ref string, ociAuth *ocitypes
 		sylog.Infof("No tag or digest found, using default: %s", SifDefaultTag)
 	}
 
-	resolver, err := getResolver(ctx, ociAuth, noHTTPS, false)
+	resolver, err := getResolver(ctx, ociAuth, noHTTPS, false, true)
 	if err != nil {
 		return fmt.Errorf("while getting resolver: %s", err)
 	}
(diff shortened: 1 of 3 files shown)


JasonYangShadow:
I will also add progress bar to `apptainer pull oras://` as well. Please do not merge until the code is applied.

panda1100:
I tested push/pull and both works great!

DrDaveD:
I'm surprised this has a whole new source file just for the oras protocol.  Aren't there progress bars on other protocols that can share code?  Or is this the first one and it should be extended to others?  Why should we treat the oras protocol specially?

panda1100:
@DrDaveD -san,
This is not the first progress bar implementation. `docker://` pull and `library://` push/pull has progress bar. Each has its own implementation. 
The reason we need unique implementation for oras "as well", as my understanding from discussing with @JasonYangShadow and @cclerget about this PR design, we need to access reader/writer buffer to track progress to implement progress bar, to access those buffer we need to control over the reader/writer returned by Push/Fetch method those are under oras library (those are not exposed as callable function?). oras library itself doesn't have progress bar implementation yet. so, we need to use those interfaces directly from Apptainer. This is the reason we decided to implement this way.

panda1100:
Thank you @DrDaveD -san, I see your point. I'll discuss with @JasonYangShadow about it tomorrow morning.

JasonYangShadow:
I will take a look at current progressbar and mpb related implementation and see if we can reuse mpb.

JasonYangShadow:
@panda1100 I have rewritten the function using apptainer existing progressbar, the same we used previously. Please help verify and check if it is okay.

panda1100:
@JasonYangShadow Thank you Jason! I did test on my end and works good for both push and pull 👍

DrDaveD (changes requested):
This is an improvement.  However I still don't like that there's a new function called `getOrasResolver()` along with an old function called `getResolver()` with the same parameters and only for oras, in fact in oras.go.  It also looks like they're mostly identical code too.

I suggest renaming `getOrasResolver` to `getResolver` and adding a flag that makes it optionally behave like the existing `getResolver`, then remove the current `getResolver`.

JasonYangShadow:
> This is an improvement. However I still don't like that there's a new function called `getOrasResolver()` along with an old function called `getResolver()` with the same parameters and only for oras, in fact in oras.go. It also looks like they're mostly identical code too.
> 
> I suggest renaming `getOrasResolver` to `getResolver` and adding a flag that makes it optionally behave like the existing `getResolver`, then remove the current `getResolver`.

Thanks for the comments, next time I will be more serious on the duplicated code.

internal/pkg/client/oras/resolver.go
```diff
@@ -0,0 +1,133 @@
+package oras
```
DrDaveD, line 1: Please include the standard header on this file, and also include a comment to say what the file is for, to add progress bars.

internal/pkg/client/oras/oras.go
```diff
@@ -86,7 +86,7 @@ func newOrasUploadTransport() http.RoundTripper {
 	}
 }
 
-func getResolver(ctx context.Context, ociAuth *ocitypes.DockerAuthConfig, noHTTPS, push bool) (remotes.Resolver, error) {
+func getResolver(ctx context.Context, ociAuth *ocitypes.DockerAuthConfig, noHTTPS, push, orasResolver bool) (remotes.Resolver, error) {
```
DrDaveD, line 89: I don't think the name `orasResolver` is very descriptive.  How about something like `progressBar` or `includeProgressBar`.

DrDaveD (changes requested):

internal/pkg/client/oras/oras.go
JasonYangShadow: renamed to `progressBar` thanks!
https://github.com/apptainer/apptainer/pull/1587/commits/aab871d7c4c217c169908a6e6afddc9f49654c3f

internal/pkg/client/oras/resolver.go
JasonYangShadow: thanks! added
https://github.com/apptainer/apptainer/pull/1587/commits/aab871d7c4c217c169908a6e6afddc9f49654c3f

DrDaveD (approved):
Please also cherry-pick to release-1.2

panda1100:
Thank you DrDave-san and Jason!

PR #1597, opened 2023-07-29
Allow containers using `--nv` and `--rocm` GPU binds simultaneously

### LINKED ISSUE #1596: [Feature Request] Use of `--nv` and `--rocm` simultaneously
I would like to be able to run both ROCm and CUDA binaries in the same container for testing and development. Is there anything fundamentally preventing this from working? I have been successful with the same requirements in a docker container, and there seems to be no overlap in the libraries involved, so I would also be happy to take a look at tackling the implementation with some guidance.

upsj:

For development and testing purposes, it may be necessary to run both ROCm and CUDA binaries in the same container. There is no obvious technical limitation that prevents this from working, so I created this PR to enable it.
Since both `--rocm` and `--nv` search for `libOpenCL.so`, I needed to remove the resulting duplicate from the bind mounts. This means that OpenCL binaries can only work on one backend, but this kind of frankenstein development environment is an obscure use case anyways, so it should not matter.

I tested the behavior locally with an image with both CUDA and ROCm installed, and it seems to work (after the usual [nvidia_uvm kernel module issues](https://apptainer.org/user-docs/3.5/gpu.html#cuda-error-unknown-when-everything-seems-to-be-correctly-configured) have been addressed)

 - Fixes #1596

### DIFF
--- a/pkg/runtime/engine/apptainer/config/config.go
+++ b/pkg/runtime/engine/apptainer/config/config.go
@@ -660,21 +660,19 @@ func (e *EngineConfig) GetTargetGID() []int {
 	return e.JSON.TargetGID
 }
 
-// Appends all entries that are not already present in the slice.
-func AppendSliceNonduplicate(slice []string, newElements []string) []string {
-	newSlice := make([]string, 0, len(slice)+len(newElements))
-	exists := make(map[string]bool)
-	for _, element := range slice {
-		newSlice = append(newSlice, element)
-		exists[element] = true
-	}
-	for _, element := range newElements {
-		if !exists[element] {
-			newSlice = append(newSlice, element)
-			exists[element] = true
+// ConcatenateSliceDeduplicate concatenates two string slices and returns a string slice without duplicated entries.
+func ConcatenateSliceDeduplicate(first []string, second []string) []string {
+	dedup := make(map[string]struct{})
+	for _, slice := range [][]string{first, second} {
+		for _, elem := range slice {
+			dedup[elem] = struct{}{}
 		}
 	}
-	return newSlice
+	slice := make([]string, 0, len(dedup))
+	for elem := range dedup {
+		slice = append(slice, elem)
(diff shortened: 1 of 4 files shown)


DrDaveD:
I wonder if there's a simpler way to do what `AppendSliceNonduplicate` does.  @cclerget do you know?

pkg/runtime/engine/apptainer/config/config.go
```diff
@@ -660,6 +660,23 @@ func (e *EngineConfig) GetTargetGID() []int {
 	return e.JSON.TargetGID
 }
 
+// Appends all entries that are not already present in the slice.
+func AppendSliceNonduplicate(slice []string, newElements []string) []string {
```
cclerget, line 664: @upsj I see mainly two issues with the current code, first issue is that `newSlice` can allocate more memory than necessary in case of duplicate, the second issue is that `slice` is not deduplicated in the first loop with `newSlice = append(newSlice, element)`, this could be rewritten/renamed like this:
```
// ConcatenateSliceDeduplicate concatenates two string slices and returns a string slice without duplicated entries.
func ConcatenateSliceDeduplicate(first []string, second []string) []string {
	dedup := make(map[string]struct{})
	for _, slice := range [][]string{first, second} {
		for _, elem := range slice {
			dedup[elem] = struct{}{}
		}
	}
	slice := make([]string, 0, len(dedup))
	for elem := range dedup {
		slice = append(slice, elem)
	}
	return slice
}
```

cclerget (changes requested):

pkg/runtime/engine/apptainer/config/config.go
upsj: Looks much cleaner, thanks!

DrDaveD:
@upsj Please also submit a PR cherry-picking this to the release-1.2 branch.

PR #1615, opened 2023-08-15
move keyserver-related cmds from remote to their own cmd, from sylabs 1905

DrDaveD:
This pulls in sylabs PR

- sylabs/singularity#1905
    which fixed
- sylabs/singularity#1893

>   Moves the operations relating to keyservers - adding them, listing them, and removing them - from subcommands of the remote command (singularity remote add-keyserver, singularity remote remove-keyserver, etc.) to their own dedicated command, keyserver (singularity keyserver add, singularity keyserver remove, etc.). See the output of singularity help keyserver for more information.
>
> Also changes the format of the keyserver list output (now generated via singularity keyserver list) in accordance with option (a) outlined in sylabs/singularity#1639

### DIFF
--- a/cmd/internal/cli/remote.go
+++ b/cmd/internal/cli/remote.go
@@ -144,8 +144,7 @@ var remoteKeyserverOrderFlag = cmdline.Flag{
 	DefaultValue: uint32(0),
 	Name:         "order",
 	ShortHand:    "o",
-	Usage:        "define the keyserver order",
-	Deprecated:   "use 'keyserver' subcommand instead of 'remote', see the output of 'apptainer help keyserver' for more information",
+	Hidden:       true,
 }
 
 // -i|--insecure (deprecated)
@@ -155,8 +154,7 @@ var remoteKeyserverInsecureFlag = cmdline.Flag{
 	DefaultValue: false,
 	Name:         "insecure",
 	ShortHand:    "i",
-	Usage:        "allow insecure connection to keyserver",
-	Deprecated:   "use 'keyserver' subcommand instead of 'remote', see the output of 'apptainer help keyserver' for more information",
+	Hidden:       true,
 }
 
 // -i|--insecure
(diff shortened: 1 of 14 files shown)


DrDaveD:
Closing & reopening to kick the DCO test

DrDaveD:
Trying again now that it ran once

cmd/internal/cli/remote.go
```diff
@@ -454,56 +448,24 @@ var RemoteStatusCmd = &cobra.Command{
...
-		}
-
-		if cmd.Flag(remoteKeyserverOrderFlag.Name).Changed && remoteKeyserverOrder == 0 {
-			sylog.Fatalf("order must be > 0")
-		}
-
-		if err := apptainer.RemoteAddKeyserver(name, uri, remoteKeyserverOrder, remoteKeyserverInsecure); err != nil {
-			sylog.Fatalf("%s", err)
-		}
+		sylog.Fatalf("command name changed: use 'keyserver add' instead of 'remote add-keyserver'")
```
GodloveD, line 455: If we are going to leave these command here what is the point of removing the functions and making them error.  Seems like a better idea to leave them in a working state and just add a deprecation warning.  At least until the next minor release (when behavior can change).

GodloveD (commented):
Seems like we should probably not change the CLI until the commands have been properly deprecated?

cmd/internal/cli/remote.go
DrDaveD: It was already going to go into the next minor release, not a patch release.  Are you saying it ought to be two minor releases before it gets removed?

GodloveD: I feel like there ought to be at least one release in which there is a deprecation warning before the behavior changes.  If the next release is a minor one then yes imho this should probably be deprecated with a warning for the entire patch release cycle.

DrDaveD: Ok I made that change.

ikaneshiro (approved):

PR #1622, opened 2023-08-20
Remove non-runtime libraries from rocmliblist.conf

### LINKED ISSUE #1621: `--rocm` includes unnecessary libraries
### Version of Apptainer

What version of Apptainer (or Singularity) are you using? `main` branch

### Expected behavior

Containers should provide libraries like rocBLAS, rocFFT etc, the host libraries should not be forwarded for this case.

### Actual behavior

The host libraries get loaded into .singularity.d/libs

When running a binary that tries to use rocBLAS, this leads to the following error:

```
rocBLAS error: Cannot read /.singularity.d/libs/rocblas/library/TensileLibrary.dat: Illegal seek
```

To my knowledge, both rocBLAS and rocFFT rely on additional files, probably for JIT compilation. It should be safest to let the container provide these files to avoid incompatibilities.

Run `apptainer shell --rocm` on a container containing and using `librocblas.so`. The host-provided library will be used instead. Trying to run a binary that uses rocBLAS will fail with the above error

Ubuntu 22.04.2 LTS

from source

upsj:

The GPU integration should provide only the necessary runtime library components to execute kernels.
Additional libraries like rocBLAS etc. should be provided by the container.

 - Fixes #1621

### DIFF
--- a/etc/rocmliblist.conf
+++ b/etc/rocmliblist.conf
@@ -16,9 +16,6 @@ libcomgr.so
 libCXLActivityLogger.so
 libelf.so
 libhc_am.so
-libhipblas.so
-libhip_hcc.so
-libhiprand.so
 libhsakmt.so
 libhsa-runtime64.so
 libmcwamp.so
@@ -30,11 +27,6 @@ libdrm.so
 libdrm_amdgpu.so
 libnuma.so
 libpci.so
-librccl.so
-librocblas.so
-librocfft-device.so
-librocfft.so
-librocrand.so
 librocr_debug_agent64.so
 
 libamdcomgr64.so


DrDaveD:
Thanks for the PR.  I'm not familiar with the rocm GPU integrations so I'll let @GodloveD and/or @ikaneshiro comment on that, but at minimum this needs a CHANGELOG.md item.  Put it in the 1.3.x `Changed defaults / behaviors` section please.

GodloveD:
I'm also not really in a position to determine whether these libraries are better installed into the container or bind-mounted in at runtime, because I don't have access to an AMD system to test.  @upsj how did you determine that these were the libraries that are required? 

I guess a user can always add them back again if they encounter difficulty.  For this reason it might be better to just comment the questionable libraries out for the time being to make it easier for a user to reverse the action if necessary.

GodloveD:
I'd like to take a chance to look at this a bit more closely, but I'll also note that another user in [this thread](https://apptainer.slack.com/archives/C02NGSXR84U/p1693406287165679) found that `libelf.so` and `libnuma.so` were in the rocm liblist and broke their container.  These seem like system libraries to me.

DrDaveD:
Yes and in that case it was a matter of incompatible glibc versions between the host and the container.

tom-papatheodore:
To follow up on the thread mentioned above, after removing `libelf.so` and `libnuma.so`, I still ran into a `GLIBCXX` incompatibility because the `--rocm` flag brought in my host ROCm environment, where the libraries (e.g., `libhsa-runtime64.so`) were built with a newer `libstdc++.so`:

> NOTE: This was using docker://rocm/tensorflow:latest

```
$ apptainer run --rocm rocm_tf.sif

Apptainer> python3
Python 3.9.17 (main, Jun  6 2023, 20:11:04)
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.


>>> import tensorflow as tf
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/dist-packages/tensorflow/__init__.py", line 37, in <module>
    from tensorflow.python.tools import module_util as _module_util
  File "/usr/local/lib/python3.9/dist-packages/tensorflow/python/__init__.py", line 36, in <module>
    from tensorflow.python import pywrap_tensorflow as _pywrap_tensorflow
  File "/usr/local/lib/python3.9/dist-packages/tensorflow/python/pywrap_tensorflow.py", line 26, in <module>
    self_check.preload_check()
  File "/usr/local/lib/python3.9/dist-packages/tensorflow/python/platform/self_check.py", line 63, in preload_check
    from tensorflow.python.platform import _pywrap_cpu_feature_guard
ImportError: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /.singularity.d/libs/libhsa-runtime64.so.1)
```

I was able to simply not use the `--rocm` flag and use the ROCm environment included within the container. That said, I've just started using containers in the last couple of weeks so this could just be user error, but I thought it was worth sharing in case it's helpful when making decisions about this PR.

GodloveD:
I've finally had a chance to look at this a little closer and try to determine which libs are necessary.  

First off, I'm pretty sure that the list of libraries that is currently in this section is just a "tried it and it worked" kind of thing.  Some of the stuff in here really doesn't have anything in particular to do with the GPU and needs to just be installed in container.  

Now, I'm not sure if this is the right approach, but in the example of the NVIDIA libraries there are specific library versions that match the version of the loaded kernel module.  So for instance on my laptop I have the following (oldish) kernel module. 
```
$ modinfo nvidia | grep '^version*'
version:        525.85.05
```
And then if you look in `lib64` you can find a bunch of libraries with that version on the end.  
```
$ ll /lib64/*525\.85\.05
-rwxr-xr-x. 1 root root  11M Jul 14 09:33 /lib64/libcudadebugger.so.525.85.05
-rwxr-xr-x. 1 root root  29M Jul 14 09:33 /lib64/libcuda.so.525.85.05
-rwxr-xr-x. 1 root root 1.3M Jul 14 09:33 /lib64/libEGL_nvidia.so.525.85.05
-rwxr-xr-x. 1 root root  67K Jul 14 09:33 /lib64/libGLESv1_CM_nvidia.so.525.85.05
-rwxr-xr-x. 1 root root 115K Jul 14 09:33 /lib64/libGLESv2_nvidia.so.525.85.05
-rwxr-xr-x. 1 root root 1.2M Jul 14 09:33 /lib64/libGLX_nvidia.so.525.85.05
-rwxr-xr-x. 1 root root 7.5M Jul 14 09:33 /lib64/libnvcuvid.so.525.85.05
-rwxr-xr-x. 1 root root 149K Jul 14 09:33 /lib64/libnvidia-allocator.so.525.85.05
-rwxr-xr-x. 1 root root 253K Jul 14 09:33 /lib64/libnvidia-cfg.so.525.85.05
-rwxr-xr-x. 1 root root  54M Jul 14 09:33 /lib64/libnvidia-compiler.so.525.85.05
-rwxr-xr-x. 1 root root  41M Jul 14 09:33 /lib64/libnvidia-eglcore.so.525.85.05
-rwxr-xr-x. 1 root root 195K Jul 14 09:33 /lib64/libnvidia-encode.so.525.85.05
-rwxr-xr-x. 1 root root 131K Jul 14 09:33 /lib64/libnvidia-fbc.so.525.85.05
-rwxr-xr-x. 1 root root  43M Jul 14 09:33 /lib64/libnvidia-glcore.so.525.85.05
-rwxr-xr-x. 1 root root 633K Jul 14 09:33 /lib64/libnvidia-glsi.so.525.85.05
-rwxr-xr-x. 1 root root  24M Jul 14 09:33 /lib64/libnvidia-glvkspirv.so.525.85.05
-rwxr-xr-x. 1 root root 1.4M Jul 14 09:33 /lib64/libnvidia-gtk2.so.525.85.05
-rwxr-xr-x. 1 root root 1.4M Jul 14 09:33 /lib64/libnvidia-gtk3.so.525.85.05
-rwxr-xr-x. 1 root root 1.8M Jul 14 09:33 /lib64/libnvidia-ml.so.525.85.05
-rwxr-xr-x. 1 root root 3.2M Jul 14 09:33 /lib64/libnvidia-ngx.so.525.85.05
-rwxr-xr-x. 1 root root  88M Jul 14 09:33 /lib64/libnvidia-nvvm.so.525.85.05
-rwxr-xr-x. 1 root root  22M Jul 14 09:33 /lib64/libnvidia-opencl.so.525.85.05
-rwxr-xr-x. 1 root root  66K Jul 14 09:33 /lib64/libnvidia-opticalflow.so.525.85.05
-rwxr-xr-x. 1 root root  20M Jul 14 09:33 /lib64/libnvidia-ptxjitcompiler.so.525.85.05
-rwxr-xr-x. 1 root root 113M Jul 14 09:33 /lib64/libnvidia-rtcore.so.525.85.05
-rwxr-xr-x. 1 root root  19K Jul 14 09:33 /lib64/libnvidia-tls.so.525.85.05
-rwxr-xr-x. 1 root root  19K Jul 14 09:33 /lib64/libnvidia-vulkan-producer.so.525.85.05
-rwxr-xr-x. 1 root root 9.9K Jul 14 09:33 /lib64/libnvidia-wayland-client.so.525.85.05
-rwxr-xr-x. 1 root root 192M Jul 14 09:33 /lib64/libnvoptix.so.525.85.05
```
These libraries are specific to the kernel module and thus should be bind mounted in at runtime.  (I'll note that not all of them appear in our current `nvidialiblist.conf` but that is a different issue.)  

Now I've installed [this version](https://repo.radeon.com/amdgpu-install/23.20/rhel/8.8/amdgpu-install-5.7.50700-1.el8.noarch.rp) of the AMD GPU driver in a container.  I've noted that some of the libraries in the container have the string `50700` at the end.  If I search for all of them I get this (pretty short) list. 

```
libamd_comgr.so.2.5.50700
libamdhip64.so.5.7.50700
libhiprtc-builtins.so.5.7.50700
libhiprtc.so.5.7.50700
libhsa-runtime64.so.1.11.50700
librocm-core.so.1.0.50700
```
I see that different projects have different version numbers here, but I suspect this is a good starting point for libraries that should be bind mounted from the host at runtime (because they are probably specific to the kernel module).  

@upsj (or anyone else with time and interest and an AMD GPU).  I propose that you attempt the following. 
- Nuke your current `rocmliblist.conf` file and replace it with the following entries. 
```
```
libamd_comgr.so
libamdhip64.so
libhiprtc-builtins
libhiprtc.so
libhsa-runtime64.so
librocm-core.so
```
- Try running a test workflow.
- If it looks like something is missing from the old `rocmliblist.conf` try installing it directly into the container.  If the error changes to something that could signify an module/library version incompatibility, add it back to the conf file.  

I'm thinking this method should probably give us a pretty good starting point for identifying the necessary libs.  Maybe we could get somebody from AMD to weigh in on this too if anybody knows the right person(s) to ping.  

```

tom-papatheodore:
I think any attempt to bring in libs from the host environment could run into trouble like I mentioned [above](https://github.com/apptainer/apptainer/pull/1622#issuecomment-1716395937) since the host libraries could be built with a different e.g., `GLIBC` than what's available in the container. See my link above for an example where this happened to me with the `libhsa-runtime64.so.1` lib:

```
ImportError: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /.singularity.d/libs/libhsa-runtime64.so.1)
```

To resolve this, I simply didn't use the `--rocm` flag and instead relied on the ROCm install in my container:

```
$ apptainer run --containall --bind=/home1/${USER},/dev/kfd,/dev/dri tensorflow_latest.sif
```

I used `--containall` to ensure my container environment didn't get polluted with unwanted components of my host environment and bind-mounted the `/dev` directories that ROCm needs to identify the GPUs (as well as my home dir). This allowed me to successfully use `rocm-smi` and `rocminfo` to collect information from the GPUs and use TensorFlow on the GPUs (confirmed by using `rocm-smi`) on a simple example: https://www.tensorflow.org/tutorials/quickstart/beginner

I'm not sure this is a general solution, but it's what currently work for me. AFAICT, any attempts to bring in libs from the outside could potentially run into trouble since they might not be compatible with the container environment. I'd be curious to hear further thoughts on this. Thanks : )

tom-papatheodore:
@GodloveD Thanks. That makes sense.

DrDaveD:
So how shall we proceed with this?

upsj:
I'll evaluate @GodloveD's suggestion next week and come back to you.

DrDaveD:
@upsj we expect to be ready for a 1.3.0 pre-release early next week.  Can you get this completed this week?

upsj:
@DrDaveD seems insufficient so far, I am getting `hipErrorInvalidDevice` trying to access the devices, but I'll need to dig further, I don't have a solid development environment with an AMD GPU right now.

wyphan:
@upsj Try also using the official ROCm Docker containers (preferably converted to `.sif` first) from here: https://hub.docker.com/u/rocm

upsj:
@tom-papatheodore I am now trying to run this in a production environment for continuous integration, and it turns out that we actually have the same issue - our HPC systems are running RHEL, and our development containers are based on Ubuntu. Not sure what NVIDIA does differently, but `--nv` works fine while `--rocm` fails with the aforementioned issue. I'll close this PR, since it seems to miss the point. I am also leaning more and more towards not binding any libraries

tom-papatheodore:
Ok, thanks @upsj. Yes, I've just been mounting `/dev/dri` from the host environment, then using the ROCm and libraries within the container.

upsj:
How do you mount the devices? It looks like `--bind` mounts use `MS_NODEV` by default, so they don't work properly on device files. (rocminfo fails with permission issues for `/dev/kfd`)

wyphan:
AFAIK with Docker/Podman you need to pass in both `/dev/dri` and `/dev/kfd`
along with the correct subuid/subgid mapping (for rootless Docker/Podman)
and proper group permissions in `render` or `video` for the host-side
users. Only then can all GPU functions inside the container work as
expected.

upsj:
I don't think this necessarily applies here, since the user and group should be inherited from the user launching `apptainer`, if I understand the implementation correctly.

PR #1663, opened 2023-09-06
Move OCI/Docker-registry-related cmds from `remote` to their own cmd, from sylabs1908

DrDaveD:
This pulls in
- sylabs/singularity#1908
which addressed
- sylabs/singularity#1894

The original PR description was
> Moves the operations associated with OCI/Docker registries - `login`, `logout`, and `list` - from subcommands of the `remote` command (`singularity remote login [options] <registryURI>`, etc.) to their own dedicated command, `registry` (`singularity registry login [options] <registryURI>`, etc.). See the output of `singularity help registry` for more information.
> 
> Note that this does not **remove** the subcommands `login`, `logout`, and `list` from `singularity remote`, as they have separate functionality when given a remote endpoint (Singularity Cloud Services, Singularity Enterprise installs) as an argument. This former "duality" of functions of `singularity remote {login, logout, list}` is in fact a central part of the motivation for this PR. (See #1639 for discussion.) See the output of `singularity help remote` for more information.

In addition, I made the old commands invoke the new ones with a deprecation warning.


ikaneshiro:
> In addition, I made the old commands invoke the new ones with a deprecation warning.
It looks like this PR just has the original commit, is there an additonal one with these changes that needs to be pushed?

DrDaveD:
> > In addition, I made the old commands invoke the new ones with a deprecation warning.
> 
> It looks like this PR just has the original commit, is there an additonal one with these changes that needs to be pushed?

No but when I redid the commit to fix a small gofumpt problem I accidentally only included the new files and not the modified ones.  Updated now.  The one commit includes all the modifications to the cherry pick.

ikaneshiro (approved):

PR #1735, opened 2023-10-10
add skipping comment line in templating

### LINKED ISSUE #1719: Templating replaces commented out lines in definition file
### Version of Apptainer

`apptainer version 1.2.3`

### Expected behavior

Commented out code to not have templated variables replaced

### Actual behavior

Commented out lines (lines beginning with `#` are also templated. This produces erroneous-like behavior if a commented line contains template arguments that are not declared (this includes commented out args).

```
Bootstrap: docker
From: ubuntu:20.04

%arguments
  my_var="mydefault"
#   not_my_var="commentedoutdefault"

%post
  echo {{ my_var }}

  # echo {{ not_my_var }}
```

Build the container.

```
apptainer build test.sif undefined_arg.def 
FATAL:   While processing the definition file: while updating the definition file with build args, err: while replacing var marshaled definition struct, err: build var not_my_var is not defined through either --build-arg (--build-arg-file) or 'arguments' section
```

```sh
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.3 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

Github source (tag v1.2.3)

JasonYangShadow:

add the ability to handle comment line starting with `# ` in template feature 

 - Fixes #1719

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Adding a new remote endpoint using the `apptainer remote add` command will
   now set the new endpoint as default. This behavior can be suppressed by
   supplying the `--no-default` (or `-n`) flag to `remote add`.
+- Skip parsing build definition file template variables after comments
+  beginning with a hash symbol.
 
 ### New Features & Functionality
 
--- a/internal/pkg/build/args/reader.go
+++ b/internal/pkg/build/args/reader.go
@@ -20,7 +20,10 @@ import (
 	"github.com/samber/lo"
 )
 
-var buildArgsRegexp = regexp.MustCompile(`{{\s*(\w+)\s*}}`)
+var (
+	buildArgsRegexp   = regexp.MustCompile(`{{\s*(\w+)\s*}}`)
+	commentLineRegexp = regexp.MustCompile(`\s*[#][^!]\s*.*`)
+)
 
 // NewReader creates a io.Reader that will provide the contents of a def file
 // with build-args replacements applied. src is an io.Reader from which the
@@ -35,12 +38,27 @@ func NewReader(src io.Reader, buildArgsMap map[string]string, defaultArgsMap map
 		return nil, err
 	}
 
+	// do templating
 	matches := buildArgsRegexp.FindAllSubmatchIndex(srcBytes, -1)
 	mapOfConsumedArgs := make(map[string]bool)
 	var buf bytes.Buffer
 	bufWriter := io.Writer(&buf)
 	i := 0
 	for _, m := range matches {
+		// find the last newline
+		newlineIdx := i
+		for start := m[0]; start >= newlineIdx; start-- {
+			if srcBytes[start] == '\n' {
+				newlineIdx = start
+				break
+			}
+		}
+
+		// check whether current line containing {{ VAR }} is commented line
+		if commentLineRegexp.Match(srcBytes[newlineIdx:m[0]]) {
+			continue
+		}
+
 		bufWriter.Write(srcBytes[i:m[0]])
 		argName := string(srcBytes[m[2]:m[3]])
 		val, ok := buildArgsMap[argName]
--- a/internal/pkg/build/args/args_test.go
+++ b/internal/pkg/build/args/args_test.go
@@ -219,6 +219,37 @@ func TestReader(t *testing.T) {
 			defaultArgsMap: map[string]string{},
 			err:            "is not defined through either --build-arg (--build-arg-file) or 'arguments' section",
 		},
+		{
+			name: "ok case with variables defined in comment lines",
+			input: `
+			%argument
+				OS_VER=1 #  comment line {{ OS_VER }}
+			%post
+			    	# comment
+				#an
+				#!/bin/{{ BASH }}
+				apt install {{ OS_VER }}#comment {{ OS_VER }}
+				#some other comment {{ OS_VER }}
+				#should not be replaced as well 
+			`,
+			output: `
+			%argument
+				OS_VER=1 #  comment line {{ OS_VER }}
+			%post
+			    	# comment
+				#an
+				#!/bin/csh
+				apt install 1#comment {{ OS_VER }}
+				#some other comment {{ OS_VER }}
+				#should not be replaced as well 
+			`,
+			argsMap: map[string]string{
+				"OS_VER": "1",
+				"BASH":   "csh",
+			},
+			defaultArgsMap: map[string]string{},
+			err:            "",
+		},
 	}
 
 	for _, test := range tests {


DrDaveD (changes requested):
This changes behavior so I don't think it should be in a 1.2.x release.  Please add a CHANGELOG entry in the 1.3.x section.

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Adding a new remote endpoint using the `apptainer remote add` command will
   now set the new endpoint as default. This behavior can be suppressed by
   supplying the `--no-default` (or `-n`) flag to `remote add`.
+- Template supports skipping comment lines started with a hash and a space `#␣`.
```
DrDaveD, line 40: This implies that it only covers lines that begin with hash-space, but actually it also applies to comments later on in the line, right?  I think the CHANGELOG entry should be more accurate.

Also, does it really have anything to do with templating?  Doesn't it apply to all definition files, whether or not templating is being used?

Now I'm wondering under what conditions exactly that comments aren't already being applied. At least in sections that are interpreted by a shell (e.g. %post) the shell should be removing them.  Maybe this shouldn't be being applied to the entire definition file, because a shell would also be smart enough to ignore comments that are inside strings, for example.

DrDaveD (changes requested):

CHANGELOG.md
JasonYangShadow: yeah, good point, any places having comment line pattern will be erased. 
I also think that this comment line erase feature should only not be applied to all sections.

JasonYangShadow: good points, thanks for the comments.
I've updated the entry in the CHANGELOG. I also refactored the code. The current plan is to check whether the template `{{ VAR }}` is defined inside a comment line during template parsing. If it is defined inside a comment line `# `, then the replacement will be skipped.

GodloveD:
My $0.02 is that it is kindof weird to only apply this it `# ` (hash space).  What is the motivation for that?  avoiding shebangs?  Can it avoid shebangs in a smarter way?

JasonYangShadow:
> My $0.02 is that it is kindof weird to only apply this it `# ` (hash space). What is the motivation for that? avoiding shebangs? Can it avoid shebangs in a smarter way?

I changed it to the pattern like starting with `#` following by at least one space, which is `#` + multiple spaces. what do you think ?

GodloveD:
So, in bash scripts you can create comments using just hashes with no spaces afterward.  If they are part of the first characters in the file, they are interpreted specially as shebangs (or other stuff by other programs like Slurm).  I think that it should be possible to copy this behavior exactly to minimize surprise for users.  Is there something I'm overlooking?

DrDaveD:
I agree, I think it should be just a hash followed by anything.  Some people might use tabs for white space, and some might not have any whitespace.

JasonYangShadow:
> So, in bash scripts you can create comments using just hashes with no spaces afterward. If they are part of the first characters in the file, they are interpreted specially as shebangs (or other stuff by other programs like Slurm). I think that it should be possible to copy this behavior exactly to minimize surprise for users. Is there something I'm overlooking?

yes, good point, should be # with anything followed, should not be `#!`

DrDaveD:
Now that this is limited to just the `%arguments` section, I think this could go into a 1.2.x release.  Changing tag.

DrDaveD:
> Now that this is limited to just the `%arguments` section, I think this could go into a 1.2.x release. Changing tag.

That was a wrong interpretation -- it isn't the `%arguments` section, but it is limited only to lines in which template variables appear, so in any case I still think that's probably limited enough to go into 1.2.x.

JasonYangShadow:
> > Now that this is limited to just the `%arguments` section, I think this could go into a 1.2.x release. Changing tag.
> 
> That was a wrong interpretation -- it isn't the `%arguments` section, but it is limited only to lines in which template variables appear, so in any case I still think that's probably limited enough to go into 1.2.x.

Yes, exactly correct. It only applies to template variables. It's solving the issue that template variables is mis-replaced in the comment line. The comment line is started with `#` but not `#!` in current impl.

dcat52:
Following up on my related query regarding which lines are being stripped before passing to build time. The current implementation strips all lines not "within" the post sections. so stripping the arguments section would follow the existing behavior (for v1.2.4).

In the example below, lines 8-9,11-14 are stripped. (skipping lines 1-7 and line 10). ```/.post.script: 5: I_ERROR_HERE: not found``` From this, it correctly reports line 5 of the post script has an error (wherever that may be in the definition file)
image

DrDaveD:
@dcat52 Please create a separate issue about the line number problem.

internal/pkg/build/args/reader.go
```diff
@@ -35,12 +38,27 @@ func NewReader(src io.Reader, buildArgsMap map[string]string, defaultArgsMap map
...
+		newlineIdx := i
+		for start := m[0]; start >= newlineIdx; start-- {
+			if srcBytes[start] == '\n' {
+				newlineIdx = start
+				break
+			}
+		}
+
+		// check whether current line containing {{ VAR }} is commented line
+		if commentLineRegexp.Match(srcBytes[newlineIdx:m[0]]) {
```
DrDaveD, line 58: What if there's a comment toward the end of the line after there's a variable to expand?  Won't this skip the whole line?

DrDaveD (changes requested):

internal/pkg/build/args/reader.go
JasonYangShadow: > What if there's a comment toward the end of the line after there's a variable to expand? Won't this skip the whole line?

do you mean this type of comment?

let's assume `OS_VER=1`
```
apt install {{ OS_VER }}#comment {{ OS_VER }}
```

it will be parsed as
```
apt install 1#comment {{ OS_VER }}
```

CHANGELOG.md
```diff
@@ -37,6 +37,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Adding a new remote endpoint using the `apptainer remote add` command will
   now set the new endpoint as default. This behavior can be suppressed by
   supplying the `--no-default` (or `-n`) flag to `remote add`.
+- Skip parsing template variables defined in comment lines started with
+  a hash symbol.
```
DrDaveD, line 41: So actually lines don't have to start with a hash symbol.  I suggest
```
Skip parsing build definition file template variables after comments beginning with a hash symbol.
```

DrDaveD (changes requested):

DrDaveD (approved):

DrDaveD:
Please also submit a PR cherry-picking this to release-1.2.

JasonYangShadow:
@DrDaveD 
This PR can not be cherry picked to release-1.2 because it was developed on top of these PRs
https://github.com/apptainer/apptainer/pull/1776
all files in this PR are originally created in the above PR first and updated in this PR.

DrDaveD:
Ok then we'll leave it with 1.3.0.

PR #1760, opened 2023-10-24
Add sharens mode for apptainer exec/run/shell

### LINKED ISSUE #1583: RFE: Use same namespaces when running in MPI context
Actually with unprivileged setup, each processes spawned by MPI run under its own user namespace, performance impact has been observed as highlighted in the issue https://github.com/apptainer/apptainer/issues/769.

The basic idea is to run all processes in the same user namespace, similarly to what's done with apptainer SUID.
The first process will setup the container (similar to `instance start`) while the other processes will join this container (like `exec instance://...`).

As it could be hard to automatically detect this case when apptainer is launched by `mpirun`, the addition of a CLI flag would be necessary to instruct `apptainer` to coordinate and run in this special mode for unprivileged setup.

JasonYangShadow:

Adding mpi mode for apptainer exec/run to run multiple containers in the same namespace when launched with mpirun

 - Fixes #1583 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
   inferred by `--containall` and `--compat`.
 - Added `--config` option to`keyserver` commands.
 - Honor an optional remoteName argument to the `keyserver list` command.
+- Adding `--sharens` mode for `apptainer exec/run/shell`, which enables to run multiple
+  apptainer instances in the same user namespace.
 
 ### Developer / API
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,7 +64,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Added `--config` option to`keyserver` commands.
 - Honor an optional remoteName argument to the `keyserver list` command.
 - Adding `--sharens` mode for `apptainer exec/run/shell`, which enables to run multiple
-  apptainer instances in the same user namespace.
+  apptainer instances created by the same parent using the same image in the same
+  user namespace.
 
 ### Developer / API
 
--- a/cmd/internal/cli/action_flags.go
+++ b/cmd/internal/cli/action_flags.go
@@ -824,7 +824,7 @@ var actionShareNSFlag = cmdline.Flag{
 	Value:        &shareNS,
 	DefaultValue: false,
 	Name:         "sharens",
-	Usage:        "mode for launching containers in the same namespace",
+	Usage:        "share the namespace and image with other containers launched from the same parent process",
 	EnvKeys:      []string{"SHARENS"},
 	Hidden:       false,
 }
--- a/cmd/internal/cli/action_flags.go
+++ b/cmd/internal/cli/action_flags.go
@@ -90,6 +90,8 @@ var (
 	ignoreUserns      bool
 
 	underlay bool // whether using underlay instead of overlay
+
+	shareNS bool // mode for launching container using shared namespace
 )
 
 // --app
@@ -816,6 +818,17 @@ var actionUnderlayFlag = cmdline.Flag{
 	Hidden:       false,
 }
 
+// --sharens
+var actionShareNSFlag = cmdline.Flag{
+	ID:           "shareNSFlag",
+	Value:        &shareNS,
+	DefaultValue: false,
+	Name:         "sharens",
+	Usage:        "mode for launching containers in the same namespace",
+	EnvKeys:      []string{"SHARENS"},
+	Hidden:       false,
+}
+
 func init() {
 	addCmdInit(func(cmdManager *cmdline.CommandManager) {
 		cmdManager.RegisterCmd(ExecCmd)
@@ -908,5 +921,6 @@ func init() {
 		cmdManager.RegisterFlagForCmd(&actionIgnoreFakerootCommand, actionsInstanceCmd...)
 		cmdManager.RegisterFlagForCmd(&actionIgnoreUsernsFlag, actionsInstanceCmd...)
 		cmdManager.RegisterFlagForCmd(&actionUnderlayFlag, actionsInstanceCmd...)
+		cmdManager.RegisterFlagForCmd(&actionShareNSFlag, actionsCmd...)
 	})
 }
--- a/e2e/instance/instance.go
+++ b/e2e/instance/instance.go
@@ -134,7 +134,7 @@ func (c *ctx) testCreateManyInstances(t *testing.T) {
 			e2e.ExpectExit(0),
 		)
 
-		c.expectInstance(t, instanceName, 1)
+		c.expectInstance(t, instanceName, 1, false)
 	}
 }
 
@@ -383,7 +383,7 @@ func (c *ctx) testInstanceWithConfigDir(t *testing.T) {
 		e2e.ExpectExit(0),
 	)
 
-	c.expectInstance(t, name, 1)
+	c.expectInstance(t, name, 1, false)
 	c.execInstance(t, name, "id")
 	c.stopInstance(t, name)
 
@@ -392,6 +392,17 @@ func (c *ctx) testInstanceWithConfigDir(t *testing.T) {
 			t.Fatalf("failed %v", err)
 		}
 	})(t)
+}
+
+// testShareNSMopde will test --sharens flag
+func (c *ctx) testShareNSMode(t *testing.T) {
+	c.env.RunApptainer(
+		t,
+		e2e.WithProfile(c.profile),
+		e2e.WithCommand("exec"),
+		e2e.WithArgs("--sharens", c.env.ImagePath, "true"),
+		e2e.ExpectExit(0),
+	)
 }
 
 // E2ETests is the main func to trigger the test suite
@@ -424,6 +435,7 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
 				{"GhostInstance", c.testGhostInstance},
 				{"CheckpointInstance", c.testCheckpointInstance},
 				{"InstanceWithConfigDir", c.testInstanceWithConfigDir},
+				{"ShareNSMode", c.testShareNSMode},
 			}
 
 			profiles := []e2e.Profile{
--- a/e2e/instance/instance.go
+++ b/e2e/instance/instance.go
@@ -10,7 +10,9 @@
 package instance
 
 import (
+	"bufio"
 	"bytes"
+	"fmt"
 	"os"
 	"os/user"
 	"path/filepath"
@@ -18,6 +20,7 @@ import (
 	"strings"
 	"syscall"
 	"testing"
+	"time"
 
 	"github.com/apptainer/apptainer/e2e/internal/e2e"
 	"github.com/apptainer/apptainer/e2e/internal/testhelper"
@@ -396,13 +399,54 @@ func (c *ctx) testInstanceWithConfigDir(t *testing.T) {
 
 // testShareNSMopde will test --sharens flag
 func (c *ctx) testShareNSMode(t *testing.T) {
-	c.env.RunApptainer(
-		t,
-		e2e.WithProfile(c.profile),
-		e2e.WithCommand("exec"),
-		e2e.WithArgs("--sharens", c.env.ImagePath, "true"),
-		e2e.ExpectExit(0),
-	)
+	dir, err := os.MkdirTemp(c.env.TestDir, "InstanceShareNS")
+	if err != nil {
+		t.Fatalf("Failed to create temporary directory: %v", err)
+	}
+	defer os.RemoveAll(dir)
+
+	file := fmt.Sprintf("%s/file", dir)
+	f, err := os.Create(file)
+	if err != nil {
+		t.Fatalf("failed to create file, this is unexpected, err: %v", err)
+	}
+	f.Close()
+
+	insNumber := 2
+	for i := 0; i < insNumber; i++ {
+		go c.env.RunApptainer(
+			t,
+			e2e.WithProfile(c.profile),
+			e2e.WithCommand("exec"),
+			e2e.WithArgs("--bind", fmt.Sprintf("%s:/canary/file", file), "--sharens", c.env.ImagePath, "sh", "-c", "echo 0 >> /canary/file; sleep 1"),
+			e2e.ExpectExit(0),
+		)
+	}
+
+	// waiting enough time for the file written
+	time.Sleep(2 * time.Second)
+
+	f, err = os.Open(file)
+	if err != nil {
+		t.Fatalf("failed to open file: %v, this is unexpected", err)
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	scanner.Split(bufio.ScanLines)
+
+	var count int
+	for scanner.Scan() {
+		count++
+	}
+
+	if err := scanner.Err(); err != nil {
+		t.Fatalf("having issue while scanning file: %s, err: %v", file, err)
+	}
+
+	if count != insNumber {
+		t.Fatalf("should have %d lines, but actually is %d", insNumber, count)
+	}
 }
 
 // E2ETests is the main func to trigger the test suite
--- a/internal/pkg/runtime/engine/apptainer/process_linux.go
+++ b/internal/pkg/runtime/engine/apptainer/process_linux.go
@@ -45,6 +45,7 @@ import (
 	apptainercallback "github.com/apptainer/apptainer/pkg/plugin/callback/runtime/engine/apptainer"
 	apptainerConfig "github.com/apptainer/apptainer/pkg/runtime/engine/apptainer/config"
 	"github.com/apptainer/apptainer/pkg/sylog"
+	"github.com/apptainer/apptainer/pkg/util/fs/lock"
 	"github.com/apptainer/apptainer/pkg/util/rlimit"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"golang.org/x/sys/unix"
@@ -63,6 +64,11 @@ const defaultShell = "/bin/sh"
 //
 //nolint:maintidx
 func (e *EngineOperations) StartProcess(masterConnFd int) error {
+	// close the opened fd inside container, preventing leak
+	if fd := e.EngineConfig.GetShareNSFd(); fd != -1 && e.EngineConfig.GetShareNSMode() {
+		sylog.Debugf("Close --sharens fd lock, fd: %d", fd)
+		_ = unix.Close(fd)
+	}
 	// Manage all signals.
 	// Queue them until they're ready to be handled below.
 	// Use a channel size of two here, since we may receive SIGURG, which is
@@ -74,7 +80,7 @@ func (e *EngineOperations) StartProcess(masterConnFd int) error {
 		return err
 	}
 
-	isInstance := e.EngineConfig.GetInstance()
+	isInstance := e.EngineConfig.GetInstance() && !e.EngineConfig.GetShareNSMode()
 	bootInstance := isInstance && e.EngineConfig.GetBootInstance()
 	shimProcess := false
 
@@ -248,6 +254,12 @@ func (e *EngineOperations) StartProcess(masterConnFd int) error {
 
 	syscall.Close(masterConnFd)
 
+	if e.EngineConfig.GetShareNSMode() && !e.EngineConfig.GetInstanceJoin() {
+		// the first process should sleep for a while waiting for other processes execution
+		// should keep the user namespace open. As we do not have any ways to know how many
+		// instances are running at the same time.
+		time.Sleep(100 * time.Millisecond)
+	}
 	for {
 		select {
 		case s := <-signals:
@@ -360,6 +372,9 @@ func (e *EngineOperations) PostStartProcess(ctx context.Context, pid int) error 
 			return err
 		}
 
+		// add sharens flag
+		file.ShareNSMode = e.EngineConfig.GetShareNSMode()
+
 		pw, err := user.CurrentOriginal()
 		if err != nil {
 			return err
@@ -427,6 +442,18 @@ func (e *EngineOperations) PostStartProcess(ctx context.Context, pid int) error 
 
 		err = file.Update()
 
+		if err != nil {
+			return err
+		}
+
+		// release lock if accquired
+		if fd := e.EngineConfig.GetShareNSFd(); fd != -1 && e.EngineConfig.GetShareNSMode() {
+			err = lock.Release(fd)
+			if err != nil {
+				return err
+			}
+		}
+
 		// send SIGUSR1 to the parent process in order to tell it
 		// to detach container process and run as instance.
 		// Sleep a bit in case child would exit
@@ -437,6 +464,7 @@ func (e *EngineOperations) PostStartProcess(ctx context.Context, pid int) error 
 
 		return err
 	}
+
 	return nil
 }
 
--- a/internal/pkg/runtime/engine/apptainer/process_linux.go
+++ b/internal/pkg/runtime/engine/apptainer/process_linux.go
@@ -451,7 +451,7 @@ func (e *EngineOperations) PostStartProcess(ctx context.Context, pid int) error 
 				return err
 			}
 		} else if fd := e.EngineConfig.GetShareNSFd(); fd != -1 {
-			br := lock.NewByteRange(fd, 0, 1)
+			br := lock.NewByteRange(fd, 0, 0)
 			if err := br.Unlock(); err != nil {
 				return err
 			}
--- a/internal/pkg/runtime/engine/apptainer/process_linux.go
+++ b/internal/pkg/runtime/engine/apptainer/process_linux.go
@@ -288,7 +288,7 @@ func (e *EngineOperations) StartProcess(masterConnFd int) error {
 				// permissions to send signals to its childs and EINVAL would
 				// mean to update the Go runtime or the kernel to something more
 				// stable :)
-				if isInstance && cmdPid > 0 {
+				if (isInstance || e.EngineConfig.GetShareNSMode()) && cmdPid > 0 {
 					if err := syscall.Kill(-cmdPid, signal); err == syscall.ESRCH {
 						sylog.Debugf("No child process, exiting ...")
 						os.Exit(128 + int(signal))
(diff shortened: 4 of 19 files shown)


DrDaveD:
I tried running this on centos7 but got no shell prompts:
```
apptainer run --mpi /scratch/tmp/centos7.sif
```
and it just hung.  Then in another window (while the first was still running):
```
apptainer exec --mpi /scratch/tmp/centos7.sif bash
```
and it also hung.

If I run it in `--debug` mode it ends like this:
```
DEBUG   [U=3382,P=358]     Update()                      Storing instance data to /nashome/d/dwd/.apptainer/instances/app/dwdosgdev.fnal.gov/dwd/mpi_instance_25969/mpi_instance_25969.json
DEBUG   [U=3382,P=1]       StartProcess()                Received signal child exited
DEBUG   [U=3382,P=358]     func1()                       Release mpi fd lock, fd: 6
```
My home directory is on NFS, so I wondered if that makes a difference.  To test that I set APPTAINER_CONFIGDIR to point to directory on the local disk, but it still hung.

Why not support this with the `shell` command too?

Also I'm not completely convinced that `--mpi` is the best name for this because it is a general feature that might have use in other contexts.  I'm thinking maybe `--sharedns` or something like that.

JasonYangShadow:
> I tried running this on centos7 but got no shell prompts:
> 
> ```
> apptainer run --mpi /scratch/tmp/centos7.sif
> INFO:    Instance stats will not be available - requires cgroups v2 with systemd as manager.
> ```
> 
> and it just hung. Then in another window (while the first was still running):
> 
> ```
> apptainer exec --mpi /scratch/tmp/centos7.sif bash
> INFO:    Instance stats will not be available - requires cgroups v2 with systemd as manager.
> ```
> 
> and it also hung.
> 
> If I run it in `--debug` mode it ends like this:
> 
> ```
> DEBUG   [U=3382,P=358]     Update()                      Storing instance data to /nashome/d/dwd/.apptainer/instances/app/dwdosgdev.fnal.gov/dwd/mpi_instance_25969/mpi_instance_25969.json
> DEBUG   [U=3382,P=1]       StartProcess()                Received signal child exited
> DEBUG   [U=3382,P=358]     func1()                       Release mpi fd lock, fd: 6
> ```
> 
> My home directory is on NFS, so I wondered if that makes a difference. To test that I set APPTAINER_CONFIGDIR to point to directory on the local disk, but it still hung.
> 
> Why not support this with the `shell` command too?
> 
> Also I'm not completely convinced that `--mpi` is the best name for this because it is a general feature that might have use in other contexts. I'm thinking maybe `--sharedns` or something like that.

The mpi mode works like this:
1. There are multiple apptainer instances starting at the almost the same time (using mpirun -n xxx apptainer exec xxx.sif command)
2. The first process will prepare and create instance related info but *won't* run the instance as daemon (No return and no background daemon process). 
3. Other process will join the previous created instance and do execution. 
4. In this way, all processes will run inside the same namespace. 
5. mpirun should terminate apptainer instance and all processes will exit

@cclerget not sure if I explained the situation clearly, please help clarify. Thanks!

DrDaveD:
Oh doesn't the first process also run the instance?   That is what people would expect to happen.  The instance should stay running until all of the commands have exited including the first one.

JasonYangShadow:
> Oh doesn't the first process also run the instance? That is what people would expect to happen. The instance should stay running until all of the commands have exited including the first one.

yeah, the first process will also run the instance, but all instances won't run as daemons.

panda1100:
@DrDaveD 
> Why not support this with the shell command too?

I would like to understand this one, what kind of use case do you have in your mind?
I can't imagine how you use `apptainer shell --mpi`.

`run` and `exec` will be look like the following:
- `mpirun -np 8 apptainer run --mpi alltoall`
- `mpirun -np 8 apptainer exec --mpi /opt/IMB/IMB-MPI1 alltoall`

cclerget (changes requested):

DrDaveD:
> @DrDaveD
> 
> > Why not support this with the shell command too?
> 
> I would like to understand this one more, [Q1] what kind of use case do you have in your mind? I can't imagine how you use `apptainer shell --mpi`.

That's true, I hadn't really thought of that.  I was just thinking in terms of being consistent across the action commands.  I suppose somebody might want to do one shell interactively and share the namespace with others that are doing exec or run.

panda1100:
@DrDaveD 
To me, that pretty much sounds like `apptainer instance` feature. Maybe I'm missing something though.
>  I suppose somebody might want to do one shell interactively and share the namespace with others that are doing exec or run.

e2e/internal/e2e/apptainercmd.go
```diff
@@ -471,6 +456,57 @@ func ExpectExit(code int, resultOps ...ApptainerCmdResultOp) ApptainerCmdOp {
 	}
 }
 
+func SkipExit(resultOps ...ApptainerCmdResultOp) ApptainerCmdOp {
```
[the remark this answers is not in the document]
JasonYangShadow: yes, this is a hack, and we actually won't need this if the `isInstance` is set to `false`.

e2e/instance/instance.go
```diff
@@ -439,6 +465,10 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
 					}
 				})
 			}
+
+			// run mpi mode using user profile only
+			c.profile = e2e.UserProfile
+			t.Run("MpiMode", c.testMpiMode)
```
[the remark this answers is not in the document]
JasonYangShadow: thanks! this is refactored and there should be no limitation.

internal/pkg/runtime/engine/apptainer/process_linux.go
```diff
@@ -63,6 +64,14 @@ const defaultShell = "/bin/sh"
 //
 //nolint:maintidx
 func (e *EngineOperations) StartProcess(masterConnFd int) error {
+	// Close mpi open fd if exists
+	defer func() {
```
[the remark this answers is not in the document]
JasonYangShadow: oh, thanks for the info! yeah, this is the issue in current impl, the first process has to wait for a signal to exit. 
I have refactored but I think another issue comes out. As the first process won't run `appinit` code any more, it'll release the lock and return. When other processes start their execution, the first process will start cleaning up. (As its execution has completed), this clean up will terminate other processes from running. I think we need some mechanism to sync(wait until) among different processes?

DrDaveD:
> @DrDaveD To me, that pretty much sounds like `apptainer instance` feature. Maybe I'm missing something though.
> 
> > I suppose somebody might want to do one shell interactively and share the namespace with others that are doing exec or run.

It would be slightly different in the number of steps you do, because an instance would just be started as a side effect instead of a separate command, and stopped automatically when the last one exits.

DrDaveD:
Now it's better and it doesn't hang.  But I don't actually see evidence of multiple commands using the `--mpi` option using the same namespaces.  Each one I run shows up as a different instance in `apptainer instance list --all`, and I see different namespaces in /proc/1/ns.  Also why can't it allow something to be run interactively?  /dev/fd/0 is /dev/null.

internal/pkg/runtime/engine/apptainer/process_linux.go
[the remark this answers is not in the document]
JasonYangShadow: thanks for the useful info. set `isInstance` to `false` will stop the hangs exactly. but as the first process can return now, then other processes will face with the issue of 
```
FATAL   [U=1000,P=733509]  StageOne()                    could not open proc directory /proc/733484: no such file or directory
```
as comment here
https://github.com/JasonYangShadow/apptainer/blob/e36960837caa53bbb3969e6f8f5fad27d01f3ca3/internal/pkg/runtime/engine/apptainer/prepare_linux.go#L732

JasonYangShadow:
> Now it's better and it doesn't hang. But I don't actually see evidence of multiple commands using the `--mpi` option using the same namespaces. Each one I run shows up as a different instance in `apptainer instance list --all`, and I see different namespaces in /proc/1/ns. Also why can't it allow something to be run interactively? /dev/fd/0 is /dev/null.

still not working due to some errors mentioned here
https://github.com/apptainer/apptainer/pull/1760#discussion_r1378381280
I am working on that. 
>  Also why can't it allow something to be run interactively
You mean for command `apptainer shell --mpi` ? I guess we can do that easily.

JasonYangShadow:
the reason might be that the first process ends too early which will close the namespace and terminating other running processes. (adding one sleep here can make it work, but this is not the solution, I think we need to add some additional sync code to prevent the first process from quitting too early?)
https://github.com/apptainer/apptainer/pull/1760/commits/4814be3508d5a29c8c2251a1f7ba912bd5487dad

GodloveD:
I'm sorry to come to this late.  Can I request that the `--sharedns` option be changed to `--sharens`?  I've been reading `--sharedns` in an different thread and it took me a while to realize that we did not mean "share dns" which would be something totally different.

DrDaveD:
> I'm sorry to come to this late. Can I request that the `--sharedns` option be changed to `--sharens`? I've been reading `--sharedns` in an different thread and it took me a while to realize that we did not mean "share dns" which would be something totally different.

It hasn't yet been decided to change the option from `--mpi` but I agree that `--sharens` is preferable to `--sharedns` if it does get changed.

DrDaveD:
> > Also why can't it allow something to be run interactively

> You mean for command `apptainer shell --mpi` ? I guess we can do that easily.

Also for `apptainer exec --mpi ... bash`; the /dev/null on stdin was causing it to exit right away.

JasonYangShadow:
> I'm sorry to come to this late. Can I request that the `--sharedns` option be changed to `--sharens`? I've been reading `--sharedns` in an different thread and it took me a while to realize that we did not mean "share dns" which would be something totally different.

yeah, we can change that easily, okay, will rename the flag.

JasonYangShadow:
I have renamed the flag from `--mpi` to `--sharens`.

but I feel like it is really hard to sync among unknown amount of containers. so the fix here
https://github.com/apptainer/apptainer/blob/7310aff4b27167bdc6056d828b6b8b1007823133/internal/pkg/runtime/engine/apptainer/process_linux.go#L257C19-L257C19
is unreliable.

I feel like that this requires further confirmation and discussion with @DrDaveD @cclerget .

DrDaveD:
I don't understand why it's important to wait for others to join.  I assume that under normal circumstances the job will run for a while, and anything that joins while it is running ought to connect to the same instance.

I tried again the current version, and I do not see it joining the same instance.   This is what I get when running these commands in 3 different windows:
```
$ apptainer exec --sharens /scratch/tmp/centos7.sif sleep 120
```
```
$ apptainer exec --sharens /scratch/tmp/centos7.sif sleep 120
```
```
$ apptainer instance list --all
INSTANCE NAME             PID     IP    IMAGE
sharens_instance_21868    7480          /scratch/tmp/centos7.sif
sharens_instance_23846    7541          /scratch/tmp/centos7.sif
```

Also having a `shell --sharens` option is of no value unless stdin is not set to /dev/null, and likewise `exec --sharens ... bash` immediately exits.

DrDaveD:
Oh, I guess it is only supporting joining during that short period of time it is waiting.  That's not what I was expecting.  I was expecting it to allow joining for as long as any of the processes are running.

JasonYangShadow:
> Oh, I guess it is only supporting joining during that short period of time it is waiting. That's not what I was expecting. I was expecting it to allow joining for as long as any of the processes are running.

yes, correctly, it was original designed for the case of mpirun, where multiple apptainer containers will be launched at the same time (only milliseconds delay). then we need to make sure that one of these launched processes creates the instance info while remain ones join. it's not acting like that we can call apptainer at different time, later ones will be executed in the namespace created by the first process.

panda1100:
@DrDaveD @JasonYangShadow 
The original idea is when parent process is the same, children processes shares same namespace.

ex.) `mpirun -np 8 apptainer run application.sif command`

in this case `command` has same parent process `mpirun` (this is not accurate though), `command`s  shares same namesapce.

DrDaveD:
Does it have to be that way?  I think allowing processes to have any amount of delay when starting up would be more robust, more intuitive, and more generally useful.

JasonYangShadow:
> Does it have to be that way? I think allowing processes to have any amount of delay when starting up would be more robust, more intuitive, and more generally useful.

In this case why not using existing instance commands like
```
apptainer instance start xx.sif xxx
apptainer exec instance://xxx command
```

DrDaveD:
> > Does it have to be that way? I think allowing processes to have any amount of delay when starting up would be more robust, more intuitive, and more generally useful.
> 
> In this case why not using existing instance commands like
> 
> ```
> apptainer instance start xx.sif xxx
> apptainer exec instance://xxx command
> ```

Because that can be less convenient because the application needs to keep track of when to start and stop the instance.

JasonYangShadow:
> > > Does it have to be that way? I think allowing processes to have any amount of delay when starting up would be more robust, more intuitive, and more generally useful.
> > 
> > 
> > In this case why not using existing instance commands like
> > ```
> > apptainer instance start xx.sif xxx
> > apptainer exec instance://xxx command
> > ```
> 
> Because that can be less convenient because the application needs to keep track of when to start and stop the instance.

Understood. then the current impl actually does not meet the goal here.
1. The instance actually is not demonized.
2. It has lock during creation targeting at multiple simultaneous launchers. 

But it can be easily changed to above mode if reaching to this agreement among the community.

DrDaveD:
Let's discuss it in the community call tomorrow.

DrDaveD:
> After the call it occurred to me that rather than starting and stopping an instance it might be easier to do the equivalent of `nsenter` and simply join an existing unprivileged user namespace.

Nevermind about that idea.  I just tried it by starting an apptainer container running, then in another window running nsenter, then exiting apptainer.  It didn't work because apptainer shut down the fuse programs, so the root filesystem started getting `Transport endpoint is not connected` errors.

JasonYangShadow:
@DrDaveD @GodloveD @panda1100 cc: @cclerget 
btw, Cedric and I have updated the code, and we made several changes
1. The `--sharens` mode is available in sub-comands `apptainer exec/run`, we think that `shell` mode can be made available by `apptainer exec xxx sh -c `. And there are few scenarios for using `--sharens` for `apptainer shell`
2. We still think that implementing `--sharens` in the way that allows arbitrary processes to join at any time makes the case fall into the instance mode. The key point here is that `--sharens` mode is developed for the case that multiple processes will have the same parent.  

Feel free to leave comments and we will have another discussion on this in the next meeting.

GodloveD:
If it is not too much trouble to add the option to the `shell` command I would advocate for it.  I think it is a good idea to try to keep the action commands in sync as much as possible, and it might be useful for debugging or whatnot.

cclerget (dismissed):

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 - The `remote status` command will now print the username, realname, and email
   of the logged-in user, if available.
+- Adding `--sharens` mode for `apptainer exec/run/shell`, which enables to run multiple
+  apptainer instances in the same user namespace.
```
DrDaveD, line 52: Please extend this to indicate that they all need to be run by the same parent process.  I believe it also shares the image, such that if anybody runs the second command with a different container image, the second one will be ignored.  I suggest explaining that in this release message.

cmd/internal/cli/action_flags.go
```diff
@@ -805,6 +807,17 @@ var actionUnderlayFlag = cmdline.Flag{
 	Hidden:       false,
 }
 
+// --sharens
+var actionShareNSFlag = cmdline.Flag{
+	ID:           "shareNSFlag",
+	Value:        &shareNS,
+	DefaultValue: false,
+	Name:         "sharens",
+	Usage:        "mode for launching containers in the same namespace",
```
DrDaveD, line 816: I would like to see the usage also mention the need for the same parent process.  I suggest
```
share the namespace and image with other containers launched from the same parent process
```

pkg/util/fs/lock/lock.go
```diff
@@ -78,15 +95,25 @@ func (r *ByteRange) flock(lockType int16) error {
 
 // Lock places a write lock for the corresponding byte-range.
 func (r *ByteRange) Lock() error {
-	return r.flock(unix.F_WRLCK)
+	return r.flock(unix.F_WRLCK, setLk)
+}
+
+// Lock places a write lock and wait for the corresponding byte-range.
```
DrDaveD, line 101: Change `wait` to `waits`

pkg/util/fs/lock/lock.go
```diff
@@ -78,15 +95,25 @@ func (r *ByteRange) flock(lockType int16) error {
...
+	return r.flock(unix.F_WRLCK, setLkw)
 }
 
 // RLock places a read lock for the corresponding byte-range.
 func (r *ByteRange) RLock() error {
-	return r.flock(unix.F_RDLCK)
+	return r.flock(unix.F_RDLCK, setLk)
+}
+
+// RLock places a read lock and wait for the corresponding byte-range.
```
DrDaveD, line 111: Change `wait` to `waits`

JasonYangShadow: thanks! updated in the latest commit.

pkg/util/fs/lock/lock.go
JasonYangShadow: thanks! updated in the latest commit.

cmd/internal/cli/action_flags.go
JasonYangShadow: thanks! updated in the latest commit.

CHANGELOG.md
JasonYangShadow: thanks! updated in the latest commit.

JasonYangShadow:
> I tried to test this interatively and got a failure. I used the command:
> 
> ```
> $ apptainer exec --sharens /scratch/tmp/ubuntu_22.04.sif bash -c 'cat /etc/os-release;ls -l /proc/$$/ns;sleep 300' &
> ```
> 
> and the second one I ran got an error:
> 
> ```
> FATAL:   instance sharens_instance_30218 already exists
> ```
> 
> In addition, when I did `apptainer instance stop sharens_instance_30218` I got the messages
> 
> ```
> INFO:    Stopping sharens_instance_30218 instance of /scratch/tmp/centos7.sif (PID=4766)
> INFO:    Killing sharens_instance_30218 instance of /scratch/tmp/centos7.sif (PID=4766) (Timeout)
> ```
> 
> and the apptainer exited but it didn't stop the bash or sleep processes.

yeah, this is an issue. Thanks so much @cclerget, he helped fix this issue.
image

DrDaveD (changes requested):
Removing accidental approval.

JasonYangShadow:
@DrDaveD 
Again, thanks so much @cclerget, there is a new commit here that should fix pid issue along with other issues. 
https://github.com/apptainer/apptainer/pull/1760/commits/db076a76624543574ea998126d833ab6f61e83f0

DrDaveD:
Ok now I see that the `--pid` option works, good, and then doing an instance stop causes all the processes to exit, good.  But I think the `--pid` option should be supplied by default as it is for the instance start command.  We have a `--no-pid` option now so if somebody doesn't want it they could add that.

JasonYangShadow:
> Ok now I see that the `--pid` option works, good, and then doing an instance stop causes all the processes to exit, good. But I think the `--pid` option should be supplied by default as it is for the instance start command. We have a `--no-pid` option now so if somebody doesn't want it they could add that.

I changed it to this
https://github.com/JasonYangShadow/apptainer/blob/2955452ae9b82e8400fa842950cbac53647ee965/internal/pkg/runtime/launch/launcher_linux.go#L340
by default, instance will always use separated pid namespace, only if `--no-pid` is set, then will not use pid namespace.

JasonYangShadow:
> Ok now I see that the `--pid` option works, good, and then doing an instance stop causes all the processes to exit, good. But I think the `--pid` option should be supplied by default as it is for the instance start command. We have a `--no-pid` option now so if somebody doesn't want it they could add that.

@cclerget mentioned once that
> PID namespace with --sharens must be disabled by default, this is incompatible with MPI

cclerget:
@DrDaveD The main use case for sharens is MPI and it's known to be incompatible with PID namespace, so from a user experience point of view it would be less convenient to force the primary consumers of this flag to also add --no-pid to make it works as expected, disabling PID namespace should be the default not the inverse. And the main difference with instance is that sharens doesn't daemonize the apptainer process, non-sharens instances are requiring PID namespace to maintain instances alive in case the container process exits, as apptainer is running its appinit process as PID 1, it maintains the instance until it's stopped by instance stop, but sharens doesn't require PID namespace and doesn't have to for supporting MPI use case, the PID namespace support with sharens has been added recently in order to generalize sharens, the usage of instance is an internal implementation details and sharens consumers shouldn't have to deal with instance stop command, that's why sharens type instances are hidden by default with instance list. @JasonYangShadow  For those reasons I'm requesting to restore previous changes to not have PID namespace by default with sharens and target MPI use case first

cclerget (changes requested):
Please restore no PID namespace by default with sharens

JasonYangShadow:
> Please restore no PID namespace by default with sharens

yeah, restored

cclerget:
@JasonYangShadow could you also change the lock file location from /tmp to /dev/shm please ?

DrDaveD:
I had forgotten about pid namespace causing problems with MPI.  Are we certain that it still causes problems in combination with a shared namespace?  Or did that only happen in combination with separate namespaces?

DrDaveD (approved):

cclerget:
> I had forgotten about pid namespace causing problems with MPI. Are we certain that it still causes problems in combination with a shared namespace? Or did that only happen in combination with separate namespaces?

Since mpirun is usually executed from the Host PID namespace, MPI processes spawned by apptainer must run under the same PID namespace to not break implementations

cclerget (approved):

panda1100:
I'm testing against HPC-X with Oralce Cloud RoCE v2.

- with `--sharens`, I couldn't run Intel MPI Benchamrk example.
- without `--sharens`, I could run Intel MPI Benchamrk example (I got Shared memory error but UCX shm fallback feature works here).

```
[rocky@inst-mvma1-instance-pool-20231220-0559 ~]$ UCX_NET_DEVICES=mlx5_2:1 mpirun -n 32 -N 16 -hostfile ~/hostlist.txt -x UCX_NET_DEVICES=mlx5_2:1 apptainer run --sharens hpcx-benchmark.sif alltoall
Warning: Permanently added 'inst-vsoca-instance-pool-20231220-0559,10.0.1.15' (ECDSA) to the list of known hosts.
INFO:    gocryptfs not found, will not be able to use gocryptfs
FATAL:   first process with --sharens has already exited, could not execute process (pid 492811)
FATAL:   first process with --sharens has already exited, could not execute process (pid 492810)
FATAL:   first process with --sharens has already exited, could not execute process (pid 492820)
FATAL:   first process with --sharens has already exited, could not execute process (pid 492809)
FATAL:   first process with --sharens has already exited, could not execute process (pid 492825)
INFO:    gocryptfs not found, will not be able to use gocryptfs
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
FATAL:   first process with --sharens has already exited, could not execute process (pid 492843)
FATAL:   first process with --sharens has already exited, could not execute process (pid 492836)
FATAL:   first process with --sharens has already exited, could not execute process (pid 492850)
INFO:    gocryptfs not found, will not be able to use gocryptfs
FATAL:   first process with --sharens has already exited, could not execute process (pid 73451)
FATAL:   first process with --sharens has already exited, could not execute process (pid 73440)
FATAL:   first process with --sharens has already exited, could not execute process (pid 73441)
FATAL:   first process with --sharens has already exited, could not execute process (pid 73442)
FATAL:   first process with --sharens has already exited, could not execute process (pid 73456)
INFO:    gocryptfs not found, will not be able to use gocryptfs
FATAL:   first process with --sharens has already exited, could not execute process (pid 73480)
FATAL:   first process with --sharens has already exited, could not execute process (pid 73467)
FATAL:   first process with --sharens has already exited, could not execute process (pid 73474)
FATAL:   first process with --sharens has already exited, could not execute process (pid 73488)
--------------------------------------------------------------------------
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[24949,1],1]
  Exit code:    255
--------------------------------------------------------------------------
```

JasonYangShadow:
> I'm testing against HPC-X with Oralce Cloud RoCE v2.
> 
>     * with `--sharens`, I couldn't run Intel MPI Benchamrk example.
> 
>     * without `--sharens`, I could run Intel MPI Benchamrk example (I got Shared memory error but UCX shm fallback feature works here). --> log [rpa.st/5IFQ](https://rpa.st/5IFQ)
> 
> 
> ```
> [rocky@inst-mvma1-instance-pool-20231220-0559 ~]$ UCX_NET_DEVICES=mlx5_2:1 mpirun -n 32 -N 16 -hostfile ~/hostlist.txt -x UCX_NET_DEVICES=mlx5_2:1 apptainer run --sharens hpcx-benchmark.sif alltoall
> Warning: Permanently added 'inst-vsoca-instance-pool-20231220-0559,10.0.1.15' (ECDSA) to the list of known hosts.
> INFO:    Instance stats will not be available - requires cgroups v2 with systemd as manager.
> INFO:    gocryptfs not found, will not be able to use gocryptfs
> FATAL:   first process with --sharens has already exited, could not execute process (pid 492811)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 492810)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 492820)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 492809)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 492825)
> INFO:    Instance stats will not be available - requires cgroups v2 with systemd as manager.
> INFO:    gocryptfs not found, will not be able to use gocryptfs
> --------------------------------------------------------------------------
> Primary job  terminated normally, but 1 process returned
> a non-zero exit code. Per user-direction, the job has been aborted.
> --------------------------------------------------------------------------
> FATAL:   first process with --sharens has already exited, could not execute process (pid 492843)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 492836)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 492850)
> INFO:    Instance stats will not be available - requires cgroups v2 with systemd as manager.
> INFO:    gocryptfs not found, will not be able to use gocryptfs
> FATAL:   first process with --sharens has already exited, could not execute process (pid 73451)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 73440)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 73441)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 73442)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 73456)
> INFO:    Instance stats will not be available - requires cgroups v2 with systemd as manager.
> INFO:    gocryptfs not found, will not be able to use gocryptfs
> FATAL:   first process with --sharens has already exited, could not execute process (pid 73480)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 73467)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 73474)
> FATAL:   first process with --sharens has already exited, could not execute process (pid 73488)
> --------------------------------------------------------------------------
> mpirun detected that one or more processes exited with non-zero status, thus causing
> the job to be terminated. The first process to do so was:
> 
>   Process name: [[24949,1],1]
>   Exit code:    255
> --------------------------------------------------------------------------
> ```

I saw that `-N` is deprecated and we just need to use `-n` ? also are you able to share me with the test env so that I can reproduce the issue?

PR #1848, opened 2023-12-06
Fix empty user's shell field in `/etc/passwd`

dennisklein:
This regression was introduced by https://github.com/apptainer/apptainer/commit/4a0d2b81b1c2784d7df196575dfffb47478532a7

Without the proposed patch:

```console
❯ build/install/bin/apptainer exec docker://rockylinux:8 tail -1 /etc/passwd
dklein:x:1000:1000:Dennis:/home/dklein:
```

With the patch:

```
❯ build/install/bin/apptainer exec docker://rockylinux:8 tail -1 /etc/passwd
dklein:x:1000:1000:Dennis:/home/dklein:/bin/zsh
```

We noticed the bug as the empty shell field causes a failure in the `salloc` command of the SLURM resource manager (https://bugs.schedmd.com/show_bug.cgi?id=18383).


internal/pkg/runtime/engine/apptainer/prepare_linux.go
```diff
@@ -1521,12 +1521,14 @@ func (e *EngineOperations) setUserInfo(useTargetIDs bool) {
...
 			e.EngineConfig.JSON.UserInfo.UID = int(pw.UID)
 			e.EngineConfig.JSON.UserInfo.GID = int(pw.GID)
+			e.EngineConfig.JSON.UserInfo.Shell = pw.Shell
 		}
 	} else {
 		e.EngineConfig.JSON.UserInfo.Username = pw.Name
 		e.EngineConfig.JSON.UserInfo.Gecos = pw.Gecos
 		e.EngineConfig.JSON.UserInfo.UID = int(pw.UID)
 		e.EngineConfig.JSON.UserInfo.GID = int(pw.GID)
+		e.EngineConfig.JSON.UserInfo.Shell = pw.Shell
```
DrDaveD, line 1531: While you're in there, can't these 5 lines of code all be under one case of `if err == nil` ?  I mean,
```
if useTargetIDs {
		pw, err = user.GetPwUID(uint32(e.EngineConfig.GetTargetUID()))
}
if err == nil {
		... 5 lines ...
}
```

DrDaveD:
I think there should also be an e2e test that checks the whole line of output in /etc/passwd.  I'm not sure exactly where to put it.  I suppose it would go in e2e/actions/regressions.go.

cclerget (approved):

internal/pkg/runtime/engine/apptainer/prepare_linux.go
dennisklein: Done

dennisklein:
> I think there should also be an e2e test that checks the whole line of output in /etc/passwd. I'm not sure exactly where to put it. I suppose it would go in e2e/actions/regressions.go.

Added a test, have a look. I tried to re-use the existing debian test image, but the test didn't fail for it contrary to the rockylinux image. But now I am stuck with some rate limit from dockerhub (even though I defined E2E_DOCKER_USERNAME/PASSWORD). Will have to wait I guess.

Issue #1844, opened 2023-12-05
Sylabs PRs to import for 1.3.0


DrDaveD:
I think it would be good to import the following Sylabs PRs for apptainer-1.3.0.
- sylabs/singularity#1957
- sylabs/singularity#1958
- sylabs/singularity#1971
- sylabs/singularity#2003
- sylabs/singularity#2033
- sylabs/singularity#2035
- sylabs/singularity#2287

JasonYangShadow:
@DrDaveD 

There are two PRs that can not be easily cherry picked
1. https://github.com/sylabs/singularity/pull/2033 it requires dependent commits as early as https://github.com/dtrudg/singularity/commit/88a85f10626c9a17cf4dc2dea02612c2eee91ecd, many changes involved. (Might have other dependent commits from other files)
2. https://github.com/sylabs/singularity/pull/1971 it also has many files diversed differently, for example, the https://github.com/dtrudg/singularity/commits/57ed2af44c2fe1f08be2c14ddc6ae5dac797d0be/internal/pkg/runtime/launcher/options.go `options.go` file has very different commits with apptainer's .

DrDaveD:
Ah, I have @vzokay working on the change to orags-go v2 so that's why I didn't list it.  That may be enough for number 1.  Let's check again after that's done.

For number 2, yes, ignore all the oci files because they're not relevant.  I think that should make it end up being very simple.

DrDaveD:
It turns out that the dependent PR we were waiting in, sylabs/singularity#1922, is quite a complex port, a bit too much for my just-learning helper.  @JasonYangShadow can you please add that to your list and see if you can do it instead?

JasonYangShadow:
> It turns out that the dependent PR we were waiting in, [sylabs/singularity#1922](https://github.com/sylabs/singularity/pull/1922), is quite a complex port, a bit too much for my just-learning helper. @JasonYangShadow can you please add that to your list and see if you can do it instead?

I've taken a look at the cherry pick. There are too many changes and differences between singualrity and apptainer. 
Changes can trace back to one year ago. 
One example is: 
https://github.com/dtrudg/singularity/commits/23fdfa85b097115da1e56c278738b9ddc858b5d1/internal/pkg/runtime/launcher/options.go

DrDaveD:
[...]

PR #1866, opened 2023-12-18
deps: replace oras-go, cherry pick sylabs #1922

JasonYangShadow:

cherry pick sylabs PR https://github.com/sylabs/singularity/pull/1922 and patch

> As we are using go-containerregistry for handling oci-sif push/pull, we can use it for oras sif push/pull and drop oras-go. This also requires a rewrite, but should ease the dependency situation going forward.

Addresses one of the PRs in
- #1844


JasonYangShadow:
after this PR gets merged, I will cherry pick this PR
https://github.com/sylabs/singularity/pull/2033

internal/pkg/client/ocisif/auth.go
```diff
@@ -0,0 +1,114 @@
...
+//   Apptainer a Series of LF Projects LLC.
+//   For website terms of use, trademark policy, privacy policy and other
+//   project policies see https://lfprojects.org/policies
+// Copyright (c) 2020, Control Command Inc. All rights reserved.
+// Copyright (c) 2020-2023, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package ocisif
```
DrDaveD, line 11: I don't think we should be adding any ocisif.  Is this required by any of our code?

DrDaveD: I see the AuthOptn function is used elsewhere, but maybe we should just be calling the package something else.

DrDaveD (changes requested):

internal/pkg/client/ocisif/auth.go
cclerget: Can we just move this file to `internal/pkg/client/oras/auth.go` ?

dtrudg: Also, please be careful with the copyright...

The auth keychain stuff contains 3rd party code from Google. See how this is mentioned in our repo here:

https://github.com/sylabs/singularity/blob/8c1e06338f5ada6c17403bc6899c6d8742317cd7/internal/pkg/remote/credential/ociauth/ociauth.go#L106

On this PR, that information has been dropped from the header in Apptainer:

https://github.com/apptainer/apptainer/pull/1866/files#diff-aead9acf55ec4678e340ed22d8da1139dcd08dcd3bfca4ad816e3ea1743498e8

DrDaveD (approved):

cclerget (approved):

PR #1887, opened 2024-01-03
Add new default `allow setuid-mount squashfs = iflimited`

### LINKED ISSUE #1874: Add `allow setuid-mount squashfs` third value `ifrestricted`, make it default
Instead of changing the default of `allow setuid-mount squashfs` to `no` as in #1845 and only warning about its effect on ECLs and other container restructions as in #1854, add a third value `ifrestricted` as the new default which only allows kernel squashfs mounts if container restrictions are also configured.  Those would be `limit container` and `allow container` restrictions in apptainer.conf or ECLs.

DrDaveD:
This adds a new middle option called `iflimited` to `allow setuid-mount squashfs` and makes it the default.  E2e tests are updated correspondingly.
- Fixes #1874

### DIFF
--- a/internal/pkg/util/fs/squashfs/squashfs.go
+++ b/internal/pkg/util/fs/squashfs/squashfs.go
@@ -10,9 +10,12 @@
 package squashfs
 
 import (
+	"github.com/apptainer/apptainer/internal/pkg/buildcfg"
+	"github.com/apptainer/apptainer/internal/pkg/syecl"
 	"github.com/apptainer/apptainer/internal/pkg/util/bin"
 	"github.com/apptainer/apptainer/pkg/sylog"
 	"github.com/apptainer/apptainer/pkg/util/apptainerconf"
+	"github.com/apptainer/apptainer/pkg/util/namespaces"
 )
 
 func getConfig() (*apptainerconf.File, error) {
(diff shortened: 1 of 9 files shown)


internal/pkg/util/fs/squashfs/squashfs.go
```diff
@@ -52,3 +55,45 @@ func GetMem() (string, error) {
...
+	} else if str == "iflimited" {
+		if len(cfg.LimitContainerOwners) > 0 ||
+			len(cfg.LimitContainerGroups) > 0 ||
+			len(cfg.LimitContainerPaths) > 0 {
+			setuidMountAllowed = true
+			sylog.Debugf("Kernel squashfs mount allowed because of limit container")
+		} else {
+			eclcfg, err := syecl.LoadConfig(buildcfg.ECL_FILE)
+			if err != nil {
+				sylog.Debugf("Kernel squashfs mount not allowed because error loading %s: %v", buildcfg.ECL_FILE, err)
```
cclerget, line 87: This should be a warning at least, ideally we should fail as it means there is something wrong with the installation

cclerget (changes requested):

internal/pkg/util/fs/squashfs/squashfs.go
DrDaveD: There's already a check for that elsewhere so I don't think it needs to be here.  If I rename ecl.toml I get a fatal error
```
FATAL:   /etc/apptainer/ecl.toml must be owned by root
```

PR #2004, opened 2024-02-09
reset `$HOME` if it is non-readable

### LINKED ISSUE #1959: Builds fail when $HOME isn't readable
### Version of Apptainer

1.2.5, at least

### Expected behavior

I expect builds to work even when the home directory isn't readable, as happens with action commands and when the home directory doesn't exist.  Warnings make sense.

### Actual behavior

Apptainer dies with these messsages:
```
WARNING: Unable to retrieve information for /root/.apptainer: stat /root/.apptainer: permission denied
WARNING: Cache disabled - cannot access parent directory of cache: stat /root/.apptainer: permission denied.
INFO:    Starting build...
FATAL:   While performing build: conveyor failed to get: while converting reference: loading registries configuration: reading registries.conf.d: lstat /root/.config/containers/registries.conf.d: permission denied
```

Set $HOME to a non-readable directory for build as an ordinary user:
```
$ HOME=/root apptainer build /tmp/centos7.sif docker://centos:7
```

It behaves the same for both suid mode and user namespace mode.

CentOS7

EPEL

JasonYangShadow:

reset `$HOME` value if it is non-readable. 

 - Fixes #1959 

## Test
```
vagrant@ubuntu:~/test$ HOME=/root apptainer build /tmp/centos7.sif docker://centos:7
Build target 'centos7.sif' already exists and will be deleted during the build process. Do you want to continue? [y/N] y
WARNING: Cache disabled - cache location / is not writable.
INFO:    Starting build...
Getting image source signatures
Copying blob 6717b8ec66cd skipped: already exists  
Copying config a61e0eb37c done   | 
Writing manifest to image destination
2024/02/13 13:16:27  info unpack layer: sha256:6717b8ec66cd6add0272c6391165585613c31314a43ff77d9751b53010e531ec
2024/02/13 13:16:27  warn rootless{usr/bin/newgidmap} ignoring (usually) harmless EPERM on setxattr "security.capability"
2024/02/13 13:16:27  warn rootless{usr/bin/newuidmap} ignoring (usually) harmless EPERM on setxattr "security.capability"
2024/02/13 13:16:27  warn rootless{usr/bin/ping} ignoring (usually) harmless EPERM on setxattr "security.capability"
2024/02/13 13:16:28  warn rootless{usr/sbin/arping} ignoring (usually) harmless EPERM on setxattr "security.capability"
2024/02/13 13:16:28  warn rootless{usr/sbin/clockdiff} ignoring (usually) harmless EPERM on setxattr "security.capability"
INFO:    Creating SIF file...
INFO:    Build complete: /tmp/centos7.sif
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Changes since v1.3.0-rc.1
   user, although it was ignored previously.
 - Fix `--sharens` failure on EL8.
 - Fix Harbor registry login failure.
+- Prevent container builds from failing when `$HOME` points to a non-readable directory.
 
 ## v1.3.0-rc.1 - \[2024-01-10\]
 
--- a/pkg/syfs/syfs.go
+++ b/pkg/syfs/syfs.go
@@ -23,6 +23,7 @@ import (
 	"sync"
 
 	"github.com/apptainer/apptainer/internal/pkg/util/env"
+	"github.com/apptainer/apptainer/internal/pkg/util/fs"
 	"github.com/apptainer/apptainer/pkg/sylog"
 )
 
@@ -74,6 +75,15 @@ func configDir(dir string) string {
 			homedir = cwd
 		} else {
 			homedir = user.HomeDir
+		}
+	}
+
+	if !fs.IsReadable(homedir) {
+		sylog.Debugf("Home dir: %s is not readable, will reset it to '/'", homedir)
+		homedir = "/"
+		err := os.Setenv("HOME", "/")
+		if err != nil {
+			sylog.Warningf("Unable to update `$HOME`: %s", err)
 		}
 	}
 


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Changes since v1.3.0-rc.1
   user, although it was ignored previously.
 - Fix `--sharens` failure on EL8.
 - Fix Harbor registry login failure.
+- Reset `$HOME` value if it is non-readable.
```
DrDaveD, line 27: That sort of reads as implying that the $HOME value itself is unreadable instead of the directory it points to.  Also the meaning of "Reset" isn't clear.   Also, again it doesn't really communicate to the user how it affects them.  So I suggest instead
```
- Prevent container builds from failing when `$HOME` points to a non-readable directory.
```

pkg/syfs/syfs.go
```diff
@@ -61,6 +62,15 @@ func configDir(dir string) string {
 	}
 
 	homedir := os.Getenv("HOME")
+	needUpdate := false
+	if homedir != "" && !fs.IsWritable(homedir) {
+		// apptainer process can not write the homedir, should ignore the manual value
+		sylog.Warningf("Environment variable `HOME` is set to a non-writable path:%s, will reset it", homedir)
+		// reset the homedir based on user's info and current working dir
+		homedir = ""
```
DrDaveD, line 70: But setting homedir to empty here is going to be immediately followed up just below by looking up the current user id from the password and making it non-empty again.  In the case of the original issue, the real user's home directory is non-readable, not just `$HOME`, so this will not fix that case.  Does it truly work to have an empty homedir?  Maybe it would work if it changed to "/" instead of being empty, and check done after the empty homedir code below.  I know that ssh for example will switch to "/" if the user's home directory isn't accessible.

pkg/syfs/syfs.go
```diff
@@ -61,6 +62,15 @@ func configDir(dir string) string {
...
+		needUpdate = true
```
DrDaveD, line 71: I don't see why this is deferred instead of updating $HOME right there.

pkg/syfs/syfs.go
```diff
@@ -61,6 +62,15 @@ func configDir(dir string) string {
 	}
 
 	homedir := os.Getenv("HOME")
+	needUpdate := false
+	if homedir != "" && !fs.IsWritable(homedir) {
```
DrDaveD, line 66: The reported issue is for non-readable, not non-writable.

DrDaveD (changes requested):

CHANGELOG.md
JasonYangShadow: thanks for the suggestion. I have updated it.

pkg/syfs/syfs.go
JasonYangShadow: thanks! I have updated the code.

pkg/syfs/syfs.go
JasonYangShadow: thanks! I forgot that the home directory is non-readable here, so setting the home dir value based on the current user's info will ultimately fail. So maybe we can set it to `/nonexistent` dir as shown in the `/etc/passwd` for user `nobody`? this actually has the similar behavior of setting it to `/`.

pkg/syfs/syfs.go
JasonYangShadow: thanks!, code is updated.

pkg/syfs/syfs.go
```diff
@@ -75,6 +77,12 @@ func configDir(dir string) string {
 		} else {
 			homedir = user.HomeDir
 		}
+	} else if homedir != nonexistentDir && !fs.IsReadable(homedir) {
```
DrDaveD, line 80: This shouldn't be an `else`, because the code just above might set the home directory to a non-readable place.

pkg/syfs/syfs.go
```diff
@@ -75,6 +77,12 @@ func configDir(dir string) string {
...
+		homedir = nonexistentDir
```
DrDaveD, line 81: I think that using `/` as a default will be less surprising to users, since that's what happens with ssh.  `/` always has to be readable.

DrDaveD (changes requested):

pkg/syfs/syfs.go
JasonYangShadow: yeah, correctly. Thanks for the correction.

pkg/syfs/syfs.go
JasonYangShadow: okay. sure.

pkg/syfs/syfs.go
```diff
@@ -77,6 +78,14 @@ func configDir(dir string) string {
 		}
 	}
 
+	if !fs.IsReadable(homedir) {
+		homedir = "/"
```
DrDaveD, line 82: Please insert a debug message here saying the home directory isn't readable, changing home directory to /.

DrDaveD (changes requested):

DrDaveD (approved):

DrDaveD:
Please cherry-pick to release-1.3.

PR #2036, opened 2024-02-21
Run image drivers with CAP_DAC_OVERRIDE

### LINKED ISSUE #2033: --nvccli fails for NVIDIA_DRIVER_CAPABILITIES=graphics
Hi,

after fixing #2028 I noticed another issue with `--nvccli`: Setting `NVIDIA_DRIVER_CAPABILITIES=graphics` does not work (and in fact fails earlier than #2028, so it's a separate issue).

I traced this to a problematic interaction with `fuse-overlayfs`, but I'm not sure if this is a `fuse-overlayfs` bug.

`nvidia-container-cli` will try to create the directory `/etc/nvidia/nvidia-application-profiles-rc.d` with mode 0555 (code [here](https://github.com/NVIDIA/libnvidia-container/blob/31c58490248068df28fee2619a4f50250de2bfd4/src/nvc_mount.c#L283)). Creating such a directory in fuse-overlayfs will fail as illustrated in this minimal example:

```bash
#!/bin/bash
mkdir lower upper work final
fuse-overlayfs -o "lowerdir=$(pwd)/lower,upperdir=$(pwd)/upper,workdir=$(pwd)/work" final
mkdir -m 0555 final/test || echo "mkdir failed!"
fusermount -u final
```

which gives

```
mkdir: cannot create directory ‘final/test’: Permission denied
mkdir failed!
```

The reason it fails is that fuse-overlayfs will create a "backing file" in its work directory with the specified mode (0555) ([code](https://github.com/containers/fuse-overlayfs/blob/c2dd76c6d7553971b7d2934c3bb4fa0b846e46b9/main.c#L2891)). It will then attempt to set some extended attributes on this file, which fails because it is not writable. This can be seen in this strace of fuse-overlayfs:

```
write(2, "ovl_mkdir(ino=1, name=test, mode"..., 38ovl_mkdir(ino=1, name=test, mode=365)
) = 38
fstatfs(4, {f_type=EXT2_SUPER_MAGIC, f_bsize=4096, f_blocks=245715567, f_bfree=20763722, f_bavail=8263639, f_files=62480384, f_ffree=56904629, f_fsid={val=[0x5ab33290, 0xff1c871c]}, f_namelen=255, f_frsize=4096, f_flags=ST_VALID|ST_RELATIME}) = 0
mkdirat(5, "2", 0555)                   = 0
openat2(5, "2", {flags=O_RDONLY, resolve=RESOLVE_IN_ROOT}, 24) = 9
fsetxattr(9, "trusted.overlay.opaque", "y", 1, 0) = -1 EPERM (Operation not permitted)
fsetxattr(9, "user.fuseoverlayfs.opaque", "y", 1, 0) = -1 EACCES (Permission denied)
unlinkat(5, "2", AT_REMOVEDIR)          = 0
```

I see three possible fixes:

1) Use the undocumented `-o xattr_permissions=2` parameter of fuse-overlayfs, which seems to create work/upper files always with the same mode and instead saves permissions as extended attributes. No idea if this has any security/portability implications.
2) File a bug with `nvidia-container-cli` maintainers and ask if they can do mkdir with 0755 mode and later chmod to 0555 - this works correctly
3) Ugly workaround: Create the `/etc/nvidia/nvidia-application-profiles-rc.d` directory ourselves (with mode 0755) prior to calling `nvidia-container-cli`.

### Version of Apptainer

```
apptainer version 1.3.0-rc.2
```

### Expected behavior

The container starts properly.

### Actual behavior

```
$ NVIDIA_DRIVER_CAPABILITIES=graphics apptainer run --nv --nvccli docker://ubuntu:22.04
INFO:    Setting 'NVIDIA_VISIBLE_DEVICES=all' to emulate legacy GPU binding.
INFO:    Setting --writable-tmpfs (required by nvidia-container-cli)
INFO:    Cleanup error: while stopping driver for /var/lib/apptainer/mnt/session/final: fuse-overlayfs exited
FATAL:   container creation failed: nvidia-container-cli failed with exit status 1: nvidia-container-cli: mount error: file creation failed: /var/lib/apptainer/mnt/session/final/etc/nvidia/nvidia-application-profiles-rc.d: permission denied
```

see above.

```sh
$ cat /etc/os-release
apptainer version 1.3.0-rc.2
```

From GitHub release page.

xqms:

This is required for fuse-overlayfs, which assumes it can create / modify files in its work directory without caring for permissions.

Also includes an e2e test for this case.

See https://github.com/containers/fuse-overlayfs/issues/414 for a related discussion.

 - Fixes #2033

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 Changes since v1.3.0-rc.2
 
+- Run image drivers with CAP_DAC_OVERRIDE in user namespace mode. This
+  fixes --nvccli with NVIDIA_DRIVER_CAPABILITIES=graphics, which
+  previously failed when using fuse-overlayfs.
 - Fix the use of `nvidia-container-cli` on Ubuntu 22.04 where an
   `ldconfig` wrapper script gets in the way. Instead, we use
   `ldconfig.real` directly.
--- a/e2e/gpu/gpu.go
+++ b/e2e/gpu/gpu.go
@@ -169,6 +169,14 @@ func (c ctx) testNvCCLI(t *testing.T) {
 			expectExit:  255,
 		},
 		{
+			// Test "graphics" capability without privileges (issue #2033)
+			name:       "UserGraphics",
+			profile:    e2e.UserProfile,
+			args:       []string{"--nvccli", imagePath, "true"},
+			env:        []string{"NVIDIA_DRIVER_CAPABILITIES=graphics"},
+			expectExit: 0,
+		},
+		{
 			// Require CUDA version >8 should be fine!
 			name:       "UserValidRequire",
 			profile:    e2e.RootProfile,
--- a/internal/pkg/image/driver/imagedriver.go
+++ b/internal/pkg/image/driver/imagedriver.go
@@ -363,6 +363,9 @@ func (d *fuseappsDriver) Mount(params *image.MountParams, mfunc image.MountFunc)
 				// Needed for nsenter
 				//  https://stackoverflow.com/a/69724124/10457761
 				uintptr(capabilities.Map["CAP_SYS_PTRACE"].Value),
+				// Required for fuse-overlayfs
+				//  see https://github.com/containers/fuse-overlayfs/issues/414#issuecomment-1956140097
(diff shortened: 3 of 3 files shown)


DrDaveD (approved):

DrDaveD:
Please also cherry-pick to release-1.3

PR #2129, opened 2024-04-05
hide `instance stats will not be available` message under `--sharens` mode

### LINKED ISSUE #2128: cgroups v2  required when using --sharens option for MPI jobs ?
Just a question.

When using  apptainer v 1.3.0 to launch MPI jobs using `--sharens`  on `Rocky Linux 8 ` i got the following WARNING/INFO:

```

```
But the job seems to run without any problem.

What this INFO exactly means ? 
Should one expect some problems related to apptainer instances if cgroups v2 is not enabled in the OS ?

JasonYangShadow:

hide `instance stats will not be available` message under `--sharens` mode

 - Fixes #2128

### DIFF
--- a/internal/pkg/runtime/launch/launcher_linux.go
+++ b/internal/pkg/runtime/launch/launcher_linux.go
@@ -1095,7 +1095,11 @@ func (l *Launcher) setCgroups(instanceName string) error {
 		return nil
 	}
 
-	sylog.Infof("Instance stats will not be available - requires cgroups v2 with systemd as manager.")
+	if l.cfg.ShareNSMode {
+		sylog.Debugf("Instance stats will not be available - requires cgroups v2 with systemd as manager.")
+	} else {
+		sylog.Infof("Instance stats will not be available - requires cgroups v2 with systemd as manager.")
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/pkg/runtime/launch/launcher_linux.go
```diff
@@ -1095,6 +1095,11 @@ func (l *Launcher) setCgroups(instanceName string) error {
 		return nil
 	}
 
+	if l.cfg.ShareNSMode {
+		sylog.Debugf("Instance stats will not be available because of sharens mode")
```
DrDaveD, line 1099: This is just a debug message so not so important, but shouldn't it say that instance stats will not be available because cgroups v2 is not available?

I would also just put the if around the message, since the very next thing is the return nil anyway.

DrDaveD (changes requested):

internal/pkg/runtime/launch/launcher_linux.go
DrDaveD: I can see how you can get that impression from the messages printed, but the function is really about enabling cgroups.  Sharens mode itself is not the reason why instance stats won't work.  That's just the reason why we're turning the message from an Info message into a Debug message.

DrDaveD (approved):

DrDaveD:
Please cherry-pick to release-1.3

PR #2257, opened 2024-06-03
Label interactive process with image filename

jeremyspencer39171:

When starting interactive container, append image filename to process name.
Such that the process name is of the form:
```
Apptainer runtime parent: example.sif
```

**Reasoning for this change:**
My team runs a service where users run containerised software as and when they need.
There are many different containers, and we'd like to get metrics about which are used, and how much.
In our case, the containers are not started as instances, so they don't appear in `apptainer instance list`.
Labelling the `Apptainer runtime parent` process with the filename gives a clear indication of which container is running.

If there is an easier/preferred mechanism to see which containers are currently running, please let me know.

This is my first attempt at `Go`, so apologies if I've made any rookie mistakes!

**Note:** Regarding the changelog, I made an entry under `Changes for v1.4.x`, I hope that's the correct place.
The docs in `CONTRIBUTING.md` talk about creating a new top level section, `Changes Since vX.Y.Z`, but I'm working under the impression the style has changed a little bit. Happy to update it though!

 - Fixes - No corresponding issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 ## Changes for v1.4.x
 
+- Label process for starter binary of interactive containers with image filename,
+  for example: `Apptainer runtime parent: example.sif`.
+
 ## v1.3.2 - \[2024-05-28\]
 
 ### Security fix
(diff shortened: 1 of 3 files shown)


DrDaveD (approved):
Looks good, thanks for the contribution.

PR #2351, opened 2024-07-11
Allow bind mount on r/o NFS filesystems to fix #2350

### LINKED ISSUE #2350: Fails to bind mount on existing file in read-only NFS mount
### Version of Apptainer

All versions that I checked (including singularity), but just in case:

```
$ apptainer --version
apptainer version 1.3.3-1
```

### Expected behavior

I want to run a container with a bind mount from path with a read-only NFS area, and then **also** mount a local file on top of an existing file in this read-only NFS area. Even though the target file system is read/only, it is accessible, so I expect to be able to bind another file on top of it, which doesn't requires any special permissions on the destination file system.

### Actual behavior

When doing that, `apptainer` fails with the error:
```
failed to create file /mnt/destfile: open /mnt/destfile: read-only file system
```

The reason is that if `apptainer` check if the file exists by opening it in with `O_WRONLY|O_CREAT|O_EXCL` flags:

https://github.com/apptainer/apptainer/blob/f79d37102700db92bbccb54a45b1b1b7b6a27f0f/internal/pkg/util/fs/layout/manager.go#L101-L108

Then in the following section, it succeeds only if the file was created, **or** if the open failed because the file existed.

https://github.com/apptainer/apptainer/blob/f79d37102700db92bbccb54a45b1b1b7b6a27f0f/internal/pkg/util/fs/layout/manager.go#L390-L397

It seems that in the Linux kernel, when you try to try to open a file with those flags, it will fail with `EROFS (Read-only file system)`. From stace:
```
openat(AT_FDCWD, "/readonly/destfile", O_WRONLY|O_CREAT|O_EXCL|O_CLOEXEC, 0644) = -1 EROFS (Read-only file system)
```

At least on my kernel, after I run an command that performs `stat(2)` on the file, its attributes go to the cache, and then it fails with `EEXIST (File exists)`. 
```
openat(AT_FDCWD, "/readonly/destfile", O_WRONLY|O_CREAT|O_EXCL|O_CLOEXEC, 0644) = -1 EEXIST (File exists)
```
I'm not sure if this is consistent across all Linux kernels, but I checked it until at least 5.14.21. 

1. Mount some NFS area as r/o

       $ mount -o ro SERVER:/PATH /mnt
2. Bind mount the NFS area, and then also a local file on a file that exists in this NFS area.

       $ apptainer exec -B /mnt/:/readonly,/tmp/sourcefile:/readonly/destfile image.sif echo Running
       FATAL:   container creation failed: mount hook function failure: hook function for tag layer returns error: failed to create file /mnt/destfile: failed to create file /mnt/destfile: open /mnt/destfile: read-only file system

3. Stat or ls the file on the NFS area to make its attributes go to the cache

        $ /bin/ls /mnt/destfile
        /mnt/destfile

4. Now the bind mount would work

        $ apptainer exec -B /mnt/:/readonly,/tmp/sourcefile:/readonly/destfile image.sif echo Running
        Running

5. If you `umount` and `mount` the NFS area, or free the dentries and inodes cache, the problem will be back.

        $ echo 2 > /proc/sys/vm/drop_caches

```sh
$ cat /etc/os-release
NAME="SLES"
VERSION="12-SP5"
VERSION_ID="12.5"
PRETTY_NAME="SUSE Linux Enterprise Server 12 SP5"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:12:sp5"
```

But also SLES15SP4.

From source.

arozenfe:

When opening a file on a r/o NFS filesystem with `O_WRONLY|O_CREAT|O_EXCL` flags, and the flag is not in the kernel's cache, it will fail with `EROFS (Read-only file system)` instead of `EEXIST (File exists)`, even when the file exists. When apptainer encounters any error except for `EEXIST`, it will fail bind mount on that file. 

This fix introduces a new check for the file's existence, if opening it failed with any error except for `EEXIST`.

 - Fixes #2350

### DIFF
--- a/internal/pkg/util/fs/layout/manager.go
+++ b/internal/pkg/util/fs/layout/manager.go
@@ -389,7 +389,13 @@ func (m *Manager) sync() error {
 			}
 			if err := m.VFS.WriteFile(path, entry.content, entry.mode); err != nil {
 				if !os.IsExist(err) {
-					return fmt.Errorf("failed to create file %s: %s", path, err)
+					// It is possible that on an NFS r/o mount, opening an existing file
+					// would return EROFS instead of EEXIST. So even if we had a different
+					// error, we check first if the file exists.
+					// If it doesn't, we return the original error.
+					if _, err1 := os.Stat(path); err1 != nil {
+						return fmt.Errorf("failed to create file %s: %s", path, err)
+					}
 				}
 				// skip content write or owner change, not created by us
 				entry.created = true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
   `--writable-tmpfs`.
 - If an error "no descriptor found for reference" is seen while getting
   an oci container, retry the operation up to five times.
-- Make fakeroot Recommended for suse rpm
+- Make fakeroot Recommended for SUSE rpms.
+- Allow bind mounts onto existing files on r/o NFS filesystems.
 
 ## v1.3.3 - \[2024-07-03\]
 


DrDaveD (changes requested):
I think this is a confusing fix.  I would rather have it check for `err == syscall.EROFS`.  There is a precedent in [container_linux.go](https://github.com/apptainer/apptainer/blob/main/internal/pkg/runtime/engine/apptainer/container_linux.go#L1111).

Or maybe it needs to be `errors.Is(err, syscall.EROFS)`.   The documentation for [`os.IsExist()`](https://pkg.go.dev/os#IsExist) says it is preferred to use `errors.Is()`.

arozenfe:
The thing is that `syscall.EROFS` can also mean that the code actually couldn't **create** a non-existing file on the read-only filesysystem, which is fine in this case. I'm checking a case where the file **does** exist and I want to mount on top of it.

I was also thinking that if the file exists, it doesn't matter which error it failed opening the file with - we should be able to bind mount over any file on the destination providing it exists, regardless of whether we succeeded or failed opening it.

arozenfe:
If you find that confusing, maybe you'd prefer to add the `os.Stat` call to `func (v *defaultVFS) WriteFile` before `os.OpenFile`. If the file exists, there's no reason to try to open it, and we can just return `nil`. If `os.Stat` fails, for any reason, only then we can try to open the file. It might be clearer that way.
https://github.com/apptainer/apptainer/blob/e1d4157f5b0221ecfd3d6beb9a97e130d7d36b7e/internal/pkg/util/fs/layout/manager.go#L101-L107

DrDaveD:
It does make some sense to me to do a Stat() first.  However it isn't at all clear that just doing a Stat() is always going to be equivalent of doing a Write.  It looks to me like this function is writing out a file, so we should also verify that the content already matches what is there.  It isn't obvious that these functions are always only used for making bind mount points.

It would be good to have input from @cclerget on this.

arozenfe:
But because you open a file with `O_CREATE` in conjunction with `O_EXCL`, the open should fail anyway if the file already exists. I agree that the name of the function is a bit misleading, but that's how it works. Adding the Stat() will not change the behavior, it will just fix this specific corner case.

internal/pkg/util/fs/layout/manager.go
```diff
@@ -389,7 +389,9 @@ func (m *Manager) sync() error {
 			}
 			if err := m.VFS.WriteFile(path, entry.content, entry.mode); err != nil {
 				if !os.IsExist(err) {
-					return fmt.Errorf("failed to create file %s: %s", path, err)
+					if _, err1 := os.Stat(path); err1 != nil {
```
cclerget, line 392: The approach makes sense, however like @DrDaveD suggested, it would be good to check if the path is a regular file when `err1 == nil`, because it could also be anything except a regular file, and in this case the bind mount will fail later, so better to check that here and return an appropriate error when it's the case

cclerget (changes requested):

internal/pkg/util/fs/layout/manager.go
arozenfe: There's not problem bind mounting on non-regular file. I tried to bind mount a regular file on a socket file and a named pipe and had no issue, it successfully mounted my regular file on both of them.

cclerget: Might worth to check if the existing path is not a directory at least because mount will fail in this case

arozenfe: If the destination is a directory, opening it will fail with `EEXIST` anyway and won't even go into the `if !os.IsExist(err)` condition. By the way, apptainer will also fail in the mount stage also if the source is a directory and the destination is a file. There's nothing anywhere at the code that checks if the type of the source and the mountpoint match anyway.

cclerget: Ok thanks for the clarifications

cclerget (approved):

internal/pkg/util/fs/layout/manager.go
DrDaveD: I would like to at least see a comment explaining what's going on here.

DrDaveD (changes requested):
Please rebase on the current main and add a CHANGELOG entry

DrDaveD:
@arozenfe this is waiting on an update from you.

arozenfe:
I know, I'm not able to do that in the next two weeks, will finish it then.

DrDaveD:
We hope to have a release in a couple more weeks so hopefully you can get to this before then.

DrDaveD (approved):

PR #1284, opened 2023-04-13
Add build options --resolv and --hosts

ShamrockLee:
## Description of the Pull Request (PR):

This patch adds the flags --resolv and --hosts for `apptainer build` to determine the path to be mounted on the container `/etc/resolv.conf` and `/etc/hosts` when building the image.

When an empty string is passed, it disables the corresponding binding.

This allows unprivileged users to build containers without relying on
the top-level directories, addressing the image-building part of #1283.

Changed function interfaces:

In util.go: createStageFile: add `destSubpath` as the second input
argument.
(This function is currently used to prepare the temporary files
to mount to `/etc/resolv.conf` and `/etc/hosts`; the impact to other
parts of code should be minimal)

Status and questions:
*   The implementation works.
*   The tests haven't been added yet. Where should I write a test for new build flags?
*   Would it be a security concern to allow unprivileged users to specify the files to mount if Apptainer is built with SUID support? If we need to check that, is there other parts of code I can reference to implement the condition?

### This fixes or addresses the following GitHub issues:

 - Partially fixes #1283

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/apptainer/apptainer/blob/main/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTORS.md)

PR #2394, opened 2024-07-29
build: Skip staging /etc/resolv.conf if there is an explicit bind option

posch:
By default, "apptainer build" mounts the host's /etc/resolv.conf. To mount a different resolv.conf file, you can now use:

  apptainer build -B /some/other/resolv.conf:/etc/resolv.conf ...

One use case is when apptainer is used to build containers in a sandbox that has no /etc/resolv.conf (e.g. nix-build).

This is a follow-up to https://github.com/apptainer/apptainer/pull/1284 which was not merged. @DrDaveD suggested to implement this using --bind or --no-mount.

It's also a follow-up to https://github.com/apptainer/apptainer/pull/1312 that wasn't merged either. @DrDaveD suggested to try an implementation based on --no-mount.

This implementation uses the existing --bind option as suggested in PR1284. --no-mount is not a build argument, but only implemented for exec/run/etc. Also, --bind seems to be the more general approach.

The goal is to run "apptainer build" in a sandbox that does not have an /etc/resolv.conf file. This currently fails with:

```
FATAL   [U=0,P=28]         runBuildLocal()               While performing build: failed to read /etc/resolv.conf: open /etc/resolv.conf: no such file or directory
```

With this change, one can create a custom resolv.conf somewhere inside the nix sandbox and pass it to apptainer build using -B. The build then skips normal staging of /etc/resolv.conf and succeeds.

 - Fixes "build" in nix sandbox.

### DIFF
--- a/internal/pkg/build/build.go
+++ b/internal/pkg/build/build.go
@@ -328,6 +328,24 @@ func (b Build) cleanUp() {
 	}
 }
 
+// Search list of binds, return true iff there is one for the given destination
+func haveBindFor(binds []string, destination string) bool {
+	for _, bindpath := range binds {
+		splitted := strings.Split(bindpath, ":")
+		src := splitted[0]
+		dst := ""
+		if len(splitted) > 1 {
+			dst = splitted[1]
+		} else {
+			dst = src
+		}
+		if destination == dst {
+			return true
+		}
+	}
+	return false
+}
+
 // Full runs a standard build from start to finish.
 func (b *Build) Full(ctx context.Context) error {
 	sylog.Infof("Starting build...")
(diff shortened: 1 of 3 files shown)


internal/pkg/build/build.go
```diff
@@ -412,11 +430,15 @@ func (b *Build) Full(ctx context.Context) error {
...
 
 		// create stage file for /etc/resolv.conf and /etc/hosts
-		sessionResolv, err := createStageFile("/etc/resolv.conf", stage.b, "Name resolution could fail")
-		if err != nil {
-			return err
-		} else if sessionResolv != "" {
-			defer os.Remove(sessionResolv)
+		// skip, if there is an explicit --bind for resolv.conf
+		sessionResolv := ""
+		if !haveBindFor(stage.b.Opts.Binds, "/etc/resolv.conf") {
```
DrDaveD, line 435: While you're there, please do the same for /etc/hosts.conf just below.

DrDaveD (changes requested):
Please redo the commit with `-s` for the DCO and force push.

internal/pkg/build/build.go
posch: Done

DrDaveD (changes requested):
Looks good but I'm sorry I forgot to ask you to add a CHANGELOG entry.  Put it in the 1.3.x section; I will ask you to cherry-pick to the release-1.3 branch after this is merged.

posch:
> Looks good but I'm sorry I forgot to ask you to add a CHANGELOG entry. Put it in the 1.3.x section;
Done.

DrDaveD (approved):

DrDaveD:
Please also cherry-pick to the release-1.3 branch in a new PR

PR #2450, opened 2024-08-21
Fix mconfig -s to build static binaries (release-1.3)

DrDaveD:
Cherry-pick of #2445 to the 1.3 branch.


DrDaveD:
Leaving this for 1.4 only since it causes crashes.

PR #2490, opened 2024-09-06
refactor: minimise containers/image imports

JasonYangShadow:
## Description of the Pull Request (PR):

This PR re-implemented sylabs PRs https://github.com/sylabs/singularity/pull/2376

which had an original description of
> Refactor code so that we minimise the number of times that we import from containers/image, without yet rewriting the image pulling & caching code that uses containers/image. 
Use the ggcr AuthConfig over the containers/image DockerAuthconfig.
Use our own TransportOptions over the containers/image SystemContext.
Where we still have functional containers/image code, translate to its types as near as possible.
A first step for https://github.com/sylabs/singularity/issues/2022

### This fixes or addresses the following GitHub issues:

Addresses multiple of the PRs in

- https://github.com/apptainer/apptainer/issues/2126

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/apptainer/apptainer/blob/main/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTORS.md)

PR #2500, opened 2024-09-11
refactor: containers/image -> ggcr step 2

JasonYangShadow:
This pulls in and adapts
- sylabs/singularity#2379

to convert the use of containers/image to the google go-container registry (which was already being used in some places but not everywhere).   Continues #2490.

The original PR had a description of
> Replaces use of containers/image/v5 with ggcr for fetching OCI images from all OCI sources.
> 
> Note that there are three areas of work pending:
> 
>     * Regression - there are no progress bars during the fetch of an OCI image.
> 
>     * Digests of images are still initially retrieved using containers/image/v5 code. ggcr does not have suitable internal
>       handling of DockerHub HEAD to get a digest without impacting API limits. Custom code will be needed.
> 
>     * The e2e tests still use containers/image/v5 to prepare docker archive images etc.
> 
> 
> A significant advantage of this refactoring is that when the cache is enabled, the rootfs of a container being built/pulled is extracted directly from the OCI cache. An temporary OCI layout is now only created in `TMPDIR` if the cache is disabled.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -383,7 +383,7 @@ jobs:
 
       - name: Upload artifacts
         if: env.run_tests
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         with:
           name: e2e-artifact
           path: builddir/e2e-cmd-report.txt
--- a/LICENSE_DEPENDENCIES.md
+++ b/LICENSE_DEPENDENCIES.md
@@ -917,12 +917,6 @@ The dependencies and their licenses are as follows:
 
 **License URL:** <https://github.com/titanous/rocacheck/blob/master/LICENSE>
 
-## github.com/urfave/cli
-
-**License:** MIT
-
-**License URL:** <https://github.com/urfave/cli/blob/master/LICENSE>
-
 ## go.mozilla.org/pkcs7
 
 **License:** MIT
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -27,7 +27,7 @@ import (
 	"github.com/apptainer/apptainer/internal/pkg/client/oras"
 	"github.com/apptainer/apptainer/internal/pkg/client/shub"
 	"github.com/apptainer/apptainer/internal/pkg/instance"
-	"github.com/apptainer/apptainer/internal/pkg/ocitransport"
+	"github.com/apptainer/apptainer/internal/pkg/ociimage"
 	"github.com/apptainer/apptainer/internal/pkg/runtime/launch"
 	"github.com/apptainer/apptainer/internal/pkg/util/env"
 	"github.com/apptainer/apptainer/internal/pkg/util/uri"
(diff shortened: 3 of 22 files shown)


JasonYangShadow:
@DrDaveD it looks like the ci builds scripts won't work somehow with this significant change. 
1. Unprivileged test won't work. showing errors related to `lchown invalid arguments`, which I believe is related to subuid/subgid mapping, as discussed here https://github.com/containers/podman/issues/12715, I saw that we deleted the `/etc/subuid & /etc/subgid` to trigger root-mapped namespace here https://github.com/apptainer/apptainer/blob/fe9dadd1d2c01d8ba01cc6783c447ce2f37b23ac/scripts/ci-unpriv-test#L21, it looks like this mode still does not meet the permission. If we are running in `privileged` mode, it runs good.
2. To run with minimum go version 1.21.9, it requires minimum host go version being 1.17.13, some old distros (ubuntu 20.04 & debian 11) do not meet this requirements.

DrDaveD:
When I compile this branch and try to do an unprivileged build without subuid/subgid mapping set up I get:
```
$ apptainer build image.sif image.def
INFO:    User not listed in /etc/subuid, trying root-mapped namespace
INFO:    The %post section will be run under the fakeroot command
INFO:    Starting build...
FATAL:   While performing build: packer failed to pack: while unpacking rootfs: error unpacking rootfs: unpack entry: run/utmp: apply hdr metadata: restore chown metadata: /scratch/tmp/build-temp-3699279536/rootfs/run/utmp: lchown /scratch/tmp/build-temp-3699279536/rootfs/run/utmp: invalid argument
```
I don't expect chown to work in this case, because it is running in only an ordinary unprivileged user namespace at that point; the fakeroot command is only used in the %post section.  That error message comes from [umoci](https://github.com/opencontainers/umoci/blob/v0.4.7/oci/layer/tar_extract.go#L119) only when `mapOptions.Rootless` is false.  So the problem is that this PR replaces in oci_unpack.c
```
if namespaces.IsUnprivileged() {
		sylog.Debugf("setting umoci rootless mode")
```
with
```
if os.Geteuid() != 0 {
```

DrDaveD (changes requested):
Don't forgot the commit co-author

JasonYangShadow:
> When I compile this branch and try to do an unprivileged build without subuid/subgid mapping set up I get:
> 
> ```
> $ apptainer build image.sif image.def
> INFO:    User not listed in /etc/subuid, trying root-mapped namespace
> INFO:    The %post section will be run under the fakeroot command
> INFO:    Starting build...
> FATAL:   While performing build: packer failed to pack: while unpacking rootfs: error unpacking rootfs: unpack entry: run/utmp: apply hdr metadata: restore chown metadata: /scratch/tmp/build-temp-3699279536/rootfs/run/utmp: lchown /scratch/tmp/build-temp-3699279536/rootfs/run/utmp: invalid argument
> ```
> 
> I don't expect chown to work in this case, because it is running in only an ordinary unprivileged user namespace at that point; the fakeroot command is only used in the %post section. That error message comes from [umoci](https://github.com/opencontainers/umoci/blob/v0.4.7/oci/layer/tar_extract.go#L119) only when `mapOptions.Rootless` is false. So the problem is that this PR replaces in oci_unpack.c
> 
> ```
> if namespaces.IsUnprivileged() {
>         sylog.Debugf("setting umoci rootless mode")
> ```
> 
> with
> 
> ```
> if os.Geteuid() != 0 {
> ```

thanks so much for the help!

DrDaveD (approved):

Issue #2126, opened 2024-04-02
Import selected PRs from SingularityCE 4.1.x


DrDaveD:
Add `--authfile` option:
- [x] sylabs/singularity#2227
- [x] sylabs/singularity#2225
- [x] sylabs/singularity#2236
- [x] sylabs/singularity#2238

Refactor to use go-container registry:
- [x] sylabs/singularity#2376
- [x] sylabs/singularity#2379
- [x] sylabs/singularity#2407
- [x] sylabs/singularity#2390

Miscellaneous:
- [x] sylabs/singularity#2230
- [ ] (not needed) sylabs/singularity#2329
- [x] sylabs/singularity#2374
- [x] sylabs/singularity#2380
- [x] sylabs/singularity#2408
- [ ] (not needed) Consider the host cleanup commit from sylabs/singularity#2451

Fixes backported to 4.1 branch:
- [x] sylabs/singularity#2569
- [ ] (not needed) sylabs/singularity#2573
- [ ] (not needed) sylabs/singularity#2699
- [x] sylabs/singularity#2777

Lint:
- [ ] (not needed) sylabs/singularity#2281
- [x] sylabs/singularity#2373
- [x] sylabs/singularity#2409

DrDaveD:
@JasonYangShadow why don't you figure on working on the first two batches having to do with `--authfile` and the go-containerregistry refactor and I'll figure on doing at least some of the miscellaneous.  We'll see how things go after those are done.

JasonYangShadow:
> @JasonYangShadow why don't you figure on working on the first two batches having to do with `--authfile` and the go-containerregistry refactor and I'll figure on doing at least some of the miscellaneous. We'll see how things go after those are done.

Sure, no problem!

DrDaveD:
@JasonYangShadow great job on finishing the first two batches.  Please move on next to the backported fixes.  I should to be able to start to do some of the miscellaneous ones next week.

PR #2539, opened 2024-10-03
fix: Address layer / image extraction issues in user namespaces, from sylabs #2699

JasonYangShadow:

This PR re-implements the Singularity PR: https://github.com/sylabs/singularity/pull/2699

> Note - there are no e2e tests for the nested containers / nested namespaces situations that are fixed by the commits in this PR. However, our e2e tests do confirm the changes don't cause regressions in non-nested cases.
The e2e framework doesn't offer a great way of executing singularity nested, and I've opened an issue (https://github.com/sylabs/singularity/issues/2700) to address this to verify functionality more generally than a messy one-off for this PR would handle.
fix: use rootless umoci inside user namespace
If we are running from within a user namespace, then use rootless OCI layer extrraction with umoci.
This permits the extraction to complete when singularity is run under unshare -r.
fix: honor --userns in unsquashfs wrapping
If singularity is executed with --userns/-u then it should also use a user namespace where it executes unsquashfs in a wrapped manner.
Previously the unsquashfs wrapping was without --userns/-u in a setuid installation. This caused extraction to fail from within a non-root-mapped user namespace (e.g. unshare -c).

Addresses multiple of the PRs in

- https://github.com/apptainer/apptainer/issues/2126

### DIFF
--- a/internal/pkg/build/sources/oci_unpack.go
+++ b/internal/pkg/build/sources/oci_unpack.go
@@ -81,7 +81,8 @@ func UnpackRootfs(_ context.Context, srcImage v1.Image, destDir string) (err err
 	}
 
 	// Allow unpacking as non-root
-	if namespaces.IsUnprivileged() {
+	insideUserNS, _ := namespaces.IsInsideUserNamespace(os.Getpid())
+	if namespaces.IsUnprivileged() || insideUserNS {
 		sylog.Debugf("setting umoci rootless mode")
 		mapOptions.Rootless = true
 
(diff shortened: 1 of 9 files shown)


internal/pkg/build/sources/oci_unpack.go
```diff
@@ -81,7 +81,8 @@ func UnpackRootfs(_ context.Context, srcImage v1.Image, destDir string) (err err
 	}
 
 	// Allow unpacking as non-root
-	if namespaces.IsUnprivileged() {
+	insideUserNS, _ := namespaces.IsInsideUserNamespace(os.Getpid())
+	if namespaces.IsUnprivileged() || insideUserNS {
```
DrDaveD, line 85: This is I think the key change that went into the original PR, but it turns out to be redundant for apptainer because `namespaces.IsUnprivileged()` already covers all the cases that `namespaces.IsInsideUserNamespace(os.Getpid())` covers.  In fact, I don't think apptainer has the problem at all that was being addressed by that PR, that of running under `unshare -r` and `unshare -c`.  They both work for me (the latter on EL9 because it doesn't exist on EL8).

DrDaveD (commented):
I'm closing this, apptainer doesn't need it.

PR #2559, opened 2024-10-09
Avoid timeouts when cleaning up after building gocryptfs-encrypted SIF files

### LINKED ISSUE #2529: Unprivileged encrypted image build shows timeout messages
### Version of Apptainer

1.3.4

### Expected behavior

I expect to be able to cleanly build an encrypted image unprivileged.

### Actual behavior

Building an encyrpted image includes these lines in the output:
```
INFO:    Creating SIF file...
INFO:    Terminating gocryptfs after timeout
INFO:    Timeouts can be caused by a running background process
```

The image seems to be created OK, but there should be a clean shutdown of gocryptfs, not a timeout.

```
apptainer build --encrypt --passphrase /tmp/image-e.sif docker://rockylinux:8
```
and enter any passphrase when prompted.

Rocky8

dnf install from EPEL.

DrDaveD:
When cleaning up from building gocryptfs-encrypted SIF files, this unmounts the gocryptfs filesystem before stopping the gocryptfs image driver, to avoid an unclean shutdown that causes timeouts.

- Fixes #2529

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Skip attempting to bind inaccessible mount points when handling the
   `mount hostfs = yes` configuration option.
 - Avoid timeouts when cleaning up from building gocryptfs-encrypted SIF files.
+- Fix bug that prevented build with `--passphrase` or `--pem-path` but
+  no `--encrypt` from implying fakeroot.
 
 ## Changes for v1.3.x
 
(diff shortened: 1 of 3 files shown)


DrDaveD:
Also I found that using `--pem-path` or `--passphrase` without `--encrypt` was implying encryption but didn't always implying the use of fakeroot.  That is also fixed here.

JasonYangShadow (approved):

cclerget (approved):

Issue #2546, opened 2024-10-04
Import selected PRs from SingularityCE 4.2.x


DrDaveD:
### Cleanup

- [x] sylabs/singularity#2919
- [x] sylabs/singularity#2957

### Testing

- [x] sylabs/singularity#2984
- [x] sylabs/singularity#2989
- [x] sylabs/singularity#2990
- [x] sylabs/singularity#2991
- [x] sylabs/singularity#3107

### New configuration options

- [x] sylabs/singularity#2915
- [x] sylabs/singularity#2956
- [x] sylabs/singularity#2981

### Miscellaneous

- [x] sylabs/singularity#2789
- [x] sylabs/singularity#2855
- [x] sylabs/singularity#2924
- [x] sylabs/singularity#3192 (at least the corrections to apptainer.conf, consider the new feature)
- [x] sylabs/singularity#3225 (selected commits)
- [x] sylabs/singularity#3266

### Lint

- [x] sylabs/singularity#3222 (just the "nolint lint" commit, and I think only one applies) - Skip because removing the `//nolint:gosec tags causes gosec ilnt errors.

JasonYangShadow:
@DrDaveD 

https://github.com/sylabs/singularity/pull/3225

regarding the cherry-pick of the above PR, Apptainer uses https://pkg.go.dev/github.com/apptainer/sif/v2 rather than https://pkg.go.dev/github.com/sylabs/sif/v2 which has version `2.19.0`, we can not cherry-pick it.

DrDaveD:
We need to update our sif to 2.19.2.  Can you please submit a PR there that cherry-picks everything missing, to bring it up to date?

JasonYangShadow:
> We need to update our sif to 2.19.2. Can you please submit a PR there that cherry-picks everything missing, to bring it up to date?

https://github.com/apptainer/sif/pull/245

created one PR syncing from sylabs sif repo.

PR #2590, opened 2024-10-25
fix: copy relative symlinks correctly between stages (from sylabs #2924)

JasonYangShadow:

This PR re-implements the Singularity PR: https://github.com/sylabs/singularity/pull/2924

> In a build stage, a valid relative symlink can be created that points to a location in the rootfs that is above the location of the symlink. E.g.:
/usr/mydir/mysubdir/true -> ../../../bin/true
This symlink is valid in the rootfs. It resolves to /bin/true.
If we attempt to copy /usr/mydir into a new build stage, then prior to this PR the symlink would cause the copy to fail.
We have been using CopyFromTar to copy from src <stage1>/usr/mydir, to <stage2>/usr/mydir. The CopyFromTar routines enforce that the target of a hard link or symlink must be under dst, or a fatal error is raised to avoid a breakout attack.
When copying between stages our boundary is not dst, but the top level of the destination stage's rootfs. We must allow links to be created that are above dst, but within the stage rootfs.
To resolve the issue, bring over enough code from
docker/docker/pkg/archive that we can implement a CopyFromTarWithRoot, that performs the same as CopyFromTar, except that we can specify a destination root directory. Any hard links and symlink targets under this destination root directory are valid.

Addresses one of the PRs in

- https://github.com/apptainer/apptainer/issues/2546


e2e/testdata/regressions/issue_2607.def
```diff
@@ -0,0 +1,16 @@
...
+mkdir -p /foo
+ln -s ../bar /foo/bar
+
+
+Bootstrap: oras
+From: ghcr.io/apptainer/alpine:3.15.0
+Stage: two
+
+%files from one
+/foo
```
DrDaveD, line 16: Add a newline to this file

DrDaveD (changes requested):
Needs rebase for CHANGELOG again.

DrDaveD (approved):

PR #2614, opened 2024-11-11
upgrade apptainer/sif/v2 (from sylabs #3225) and update to go1.22

JasonYangShadow:

This PR re-implements the Singularity PR: https://github.com/sylabs/singularity/pull/3225

> Bumps [github.com/sylabs/sif/v2](https://github.com/sylabs/sif) from 2.18.0 to 2.19.0.

Addresses one of the PRs in

- https://github.com/apptainer/apptainer/issues/2546


DrDaveD (changes requested):
We have been avoiding upgrading to go-1.22 until it is available in el8 & el9.  Fortunately, it is now available there.  Slightly unfortunately, it is version 1.22.7, not 1.22.8.  Please change all the versions to 1.22.7.  The `hstgo_version` should be just `1.22`.

There also needs to be a CHANGELOG item saying go1.22 is now required.

DrDaveD:
Actually I don't see the go 1.22 upgrade yet on EL9.  The bugzilla issue where it was upgraded for EL8 is https://bugzilla.redhat.com/show_bug.cgi?id=2292787.  Hopefully it will be on EL9 soon since it's on EL8.  I don't think we should wait any longer.

DrDaveD:
apptainer/sif 2.19.2 unfortunately requires golang 1.22.8, but I'm going to back it to an earlier version via apptainer/sif#249 and make a 2.19.3.

JasonYangShadow:
> apptainer/sif 2.19.2 unfortunately requires golang 1.22.8, but I'm going to back it to an earlier version via [apptainer/sif#249](https://github.com/apptainer/sif/pull/249) and make a 2.19.3.

thanks for the PR. It looks like 1.22.5 is okay to go. Tested that `hstgo_version` needs to be updated to 1.22.5, 1.22 will raise an error.

scripts/ci-deb-build-test
```diff
@@ -84,6 +84,7 @@ su testuser -c '
     fi
   fi
   go version
+  export GOTOOLCHAIN=auto
```
DrDaveD, line 87: I am skeptical about this.  The script already has its own code for downloading & compiling a newer go version when needed.  What happens when it is not set (here and for rpm)?

JasonYangShadow: @DrDaveD 
see the error in the pipeline
https://github.com/apptainer/apptainer/actions/runs/11798829974/job/32865935843?pr=2614
it'll use container's golang toolchai.

DrDaveD: If you look earlier in that output, it is already downloading the source and binary of go 1.22.5.   That is required to build the resulting source rpm on EPEL/Fedora (the output of that ci script is also used for the released packages on github), because online downloads are not allowed on that build system.   There's something else wrong.  We must not set GOTOOLCHAIN=auto.

I reproduced the problem interactively under docker containers.  The problem is that the first mconfig stores the go path in the Makefile.  The best solution I can come up with is to re-run mconfig after the go binary is downloaded:
```
diff --git a/scripts/ci-rpm-build-test b/scripts/ci-rpm-build-test
index 90c5b421d..59da62771 100755
--- a/scripts/ci-rpm-build-test
+++ b/scripts/ci-rpm-build-test
@@ -54,6 +54,8 @@ su testuser -c '
       curl -f -L -sS https://golang.org/dl/$GOBIN | tar -xzf - -C /local
       PATH=/local/go/bin:$PATH
     fi
+    # re-do the mconfig to locate the new go
+    ./mconfig --only-rpm
   fi
   go version
   ./scripts/download-dependencies
```

Debian does not have the same problem since it doesn't need to use mconfig to build a .spec file.  I removed `export GOTOOLCHAIN=auto` from scripts/ci-deb-build-test and it worked fine in my test.  Did you see a failure there?

DrDaveD (changes requested):

scripts/ci-deb-build-test
JasonYangShadow: thanks! looks like it works.

DrDaveD (approved):

PR #2680, opened 2024-12-13
fix CVE-2024-45337

mslacken:
Manual update to golang.org/x/crypto v0.31.0, other libraries are updated per `go mod tidy`
Fixes https://www.cve.org/CVERecord?id=CVE-2024-45337


DrDaveD:
Thanks for the PR but Apptainer does not use that portion of the crypto library so the CVE is not relevant.  Also, dependabot has already made #2676 containing this update.

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@ For older changes see the [archived Singularity change log](https://github.com/a
 - Change message log level from warning to debug when environment variables
   set inside a container or by APPTAINERENV have a different value than the
   environment variable on the host.
+- Update to golang.org/x/crypto v0.31.0 to fix CVE-2024-45337
```
DrDaveD, line 80: There's no need for a CHANGELOG entry since the CVE isn't relevant, please remove it.

DrDaveD (changes requested):
On the other hand I can't merge #2676 because it includes updates to other modules that increase the golang version to 1.23.x which I don't want to do yet.   So we might as well update just this one, but it also needs to have `scripts/update-license-dependencies.sh` run.

DrDaveD (approved):

PR #2535, opened 2024-10-02
cgroups: rework checks for cgroup support, from sylabs #2569

JasonYangShadow:
## Description of the Pull Request (PR):

This PR re-implements the Singularity PR: https://github.com/sylabs/singularity/pull/2569

which had an original description of
> Rework the location of checks for cgroup management support, and add fall-back logic, so that when all of the following apply:
A system is using cgroups v2
Singularity is configured to use systemd for cgroups management
The system configuration / environment doesn't set usable XDG_RUNTIME_DIR / DBUS_SESSION_BUS_ADDRESS env vars
The user has not explicitly requested resource limits (e.g --cpus)
... It is still possible to:
Start an instance as non-root (without the cgroup needed for instance stats)
run / shell / exec a container as non-root in OCI-Mode

### This fixes or addresses the following GitHub issues:

Addresses multiple of the PRs in

- https://github.com/apptainer/apptainer/issues/2126

#### Before submitting a PR, make sure you have done the following:

- [ ] Read the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- [ ] Make sure all commits are signed-off with `git commit -s`, see the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md).
- [ ] Added changes to the [CHANGELOG](https://github.com/apptainer/apptainer/blob/main/CHANGELOG.md), if necessary according to the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md).
- [ ] Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)).
- [ ] Based this PR against the appropriate branch according to the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md).
- [ ] Added myself as a contributor to the [Contributors File](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTORS.md).

PR #2712, opened 2025-01-17
Don't disable instance cgroups with user namespaces

### LINKED ISSUE #2684: Instance stats not working even when cgroups v2 is available
### Version of Apptainer

1.3.5

### Expected behavior

I expect there to not be a warning about a lack of instance stats when cgroups v2 are available under systemd.

### Actual behavior

I get this message when starting an instance on el9 where cgroups v2 is enabled by default:
```
```

```
apptainer instance start docker://rockylinux:9 myi
```

el9

From EPEL.

DrDaveD:
This stops disabling instance cgroups when running unprivileged.

- Fixes #2684
- The bug was introduced in #2535 which has not yet been released so there's no need for a  CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,13 +23,11 @@ For older changes see the [archived Singularity change log](https://github.com/a
   by squashfuse_ll.
 - Add automated tests for OpenSUSE Leap and Tumbleweed and Debian Bookworm.
 - Fixed typo in `nvliblist.conf` (`libnvoptix.so.1` -> `libnvoptix.so`)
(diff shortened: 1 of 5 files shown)


DrDaveD:
There's no need for a new CHANGELOG entry, but I revised the previous one because it was misleading.

JasonYangShadow (approved):
thanks! lgtm!

JasonYangShadow:
@DrDaveD feel free to merge the PR, thanks for the PR.

cclerget (approved):

PR #2715, opened 2025-01-21
Split up the minor dependabot group

DrDaveD:
This removes the "minor" dependabot group because some of them require updates to too-new golang and others don't.  By splitting them up, we'll be able to update the ones that don't require a too-new goolang.

This group was added in #2633 along with others.


DrDaveD:
It also includes a fix for the tumbleweed CI check failure

ikaneshiro (approved):

PR #2731, opened 2025-01-22
Support binfmt_misc also with rootless fakeroot

### LINKED ISSUE #2730: The detection of emulation does not work with rootless fakeroot
When trying to build an image under Rosetta (amd64 emulation on arm64), you get:

`FATAL:   While checking build target: while determining 'test.sif' format: the image's architecture (amd64) could not run on the host's (arm64)`

This is strange, because it is _supposed_ to be there in "binfmt_misc" (same error with qemu)

Adding some error reporting (original code ignores it), sheds some light on what is going on:

`WARNING: open /proc/sys/fs/binfmt_misc/status: no such file or directory`

There is a problem with the `/proc` mount, and checking what is supported by the kernel.

afbjorklund:

Support binfmt_misc also with rootless fakeroot

Currently it only works _without_ subuid/subgid

 - Fixes #2730

- https://github.com/lima-vm/lima/discussions/3135

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ The Singularity Project has been
 and re-branded as Apptainer.
 For older changes see the [archived Singularity change log](https://github.com/apptainer/singularity/blob/release-3.8/CHANGELOG.md).
 
+## v1.4.x changes
+
+Changes since 1.4.0-rc.1
+
+- Fix running and building containers of different architectures
+  than the host via binfmt_misc when using rootless fakeroot.
+
 ## v1.4.0 Release Candidate 1 - \[2025-01-21\]
 
 Changes since 1.3.6
--- a/internal/pkg/runtime/engine/fakeroot/engine_linux.go
+++ b/internal/pkg/runtime/engine/fakeroot/engine_linux.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path/filepath"
 	"runtime"
 	"syscall"
 
@@ -266,6 +267,7 @@ func fakerootSeccompProfile() *specs.LinuxSeccomp {
 func (e *EngineOperations) StartProcess(_ int) error {
 	const (
 		mountInfo    = "/proc/self/mountinfo"
+		binfmtMisc   = "/proc/sys/fs/binfmt_misc"
 		selinuxMount = "/sys/fs/selinux"
 	)
 
@@ -294,9 +296,25 @@ func (e *EngineOperations) StartProcess(_ int) error {
 	if err != nil {
 		return fmt.Errorf("failed to mount %s to /root: %s", e.EngineConfig.Home, err)
 	}
-	err = syscall.Mount("proc", "/proc", "proc", syscall.MS_NOSUID|syscall.MS_NOEXEC|syscall.MS_NODEV, "")
+	tmpdir, err := os.MkdirTemp("", "bind-mount-*")
+	if err != nil {
+		return fmt.Errorf("failed to create temporary directory: %s", err)
+	}
+	err = syscall.Mount("proc", tmpdir, "proc", syscall.MS_NOSUID|syscall.MS_NOEXEC|syscall.MS_NODEV, "")
 	if err != nil {
 		return fmt.Errorf("failed to mount proc filesystem: %s", err)
+	}
+	err = syscall.Mount(binfmtMisc, filepath.Join(tmpdir, "sys", "fs", "binfmt_misc"), "", syscall.MS_BIND, "")
+	if err != nil {
+		return fmt.Errorf("failed to mount binfmt_misc filesystem: %s", err)
+	}
+	err = syscall.Mount(tmpdir, "/proc", "", syscall.MS_MOVE, "")
+	if err != nil {
+		return fmt.Errorf("failed to mount proc filesystem: %s", err)
+	}
+	err = os.Remove(tmpdir)
+	if err != nil {
+		return fmt.Errorf("failed to remove temporary directory: %s", err)
 	}
 
 	// fix potential issue with SELinux (https://github.com/apptainer/singularity/issues/4038)
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -8,6 +8,7 @@
 - Alexander Grund <alexander.grund@tu-dresden.de>
 - Amanda Duffy <aduffy@lenovo.com>
 - Ana Guerrero Lopez <aguerrero@suse.com>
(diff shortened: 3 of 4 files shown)


internal/pkg/runtime/engine/fakeroot/engine_linux.go
```diff
@@ -294,10 +296,47 @@ func (e *EngineOperations) StartProcess(_ int) error {
...
+			continue
+		}
+		content[entry.Name()] = b
+	}
 	err = syscall.Mount("proc", "/proc", "proc", syscall.MS_NOSUID|syscall.MS_NOEXEC|syscall.MS_NODEV, "")
 	if err != nil {
 		return fmt.Errorf("failed to mount proc filesystem: %s", err)
 	}
+	// save a copy of binfmt_misc
+	binfmtMisc = "/tmp/sys/fs/binfmt_misc"
```
DrDaveD, line 323: Using a path that's shared by all processes on a machine is not a good idea.  They will clash.

However I'm not sure what else to suggest as a good solution to this problem.  @cclerget can you?

DrDaveD (changes requested):
Thanks for your first Apptainer PR!  We'll have to come up with a different approach, though, because the /tmp path will clash between different apptainer commands on the same machine.  

Once we do have an acceptable approach, we will need you to add yourself to CONTRIBUTORS.md and add a CHANGELOG.md entry as the checklist shows.  An e2e test would be very useful but I suspect it will be too challenging since we don't have different architectures to test with during the e2e run.

internal/pkg/runtime/engine/fakeroot/engine_linux.go
afbjorklund: The usual path is _also_ shared by all processes (being a kernel one), but there is a race during creation now.

i.e. Previously it was leaving the old settings, so I now delete any old directory - that probably needs a flock

DrDaveD: The usual path is a special kernel-created path that is different for every process, like a mount namespace.  Apptainer uses a mount namespace too including a tmpfs filesystem but I'm not sure if it is created at that point.

afbjorklund: It needs to be mounted, but unfortunately binft_misc is not yet part of user namespaces so it is not allowed as rootless (like /proc)

afbjorklund: But that's ok, since we don't need to register anything from within the container. A read-only copy of what is already loaded in the kernel is fine...

afbjorklund:
> An e2e test would be very useful but I suspect it will be too challenging since we don't have different architectures to test with during the e2e run.

I could not find any existing e2e tests with qemu, which was probably why the functionality was broken in the first place (since rootless has been there a while)?

internal/pkg/runtime/engine/fakeroot/engine_linux.go
cclerget: @afbjorklund why not just inject a bind mount for `/proc/sys/fs/binfmt_misc` on command line when using `--fakeroot`  instead of duplicating the directory content here?

DrDaveD: @cclerget Wouldn't that be a problem since `/proc` is a new mount?  A bind at this point would just come from the new `/proc`, and a bind before would be covered up by the new `/proc`.  Maybe it could be bound to a temporary place before and bound again from there to the original place afterward.  But where's a good temporary place?

afbjorklund: The new /proc has an empty directory for the binfmt_misc, which is what lead to the "status" not found error. The rootless user cannot mount the real binfmt_misc, since it is not covered by the namespace. And trying to populate the directory directly leads to errors, since it is not a normal directory (you go through "register")

afbjorklund: I did not try a temporary mountpoint.

cclerget: doing the equivalent of this in Go should do it:

```
tmpdir=$(mktemp -d bind-mount-XXXX)
mount -t proc proc $tmpdir
mount --bind /proc/sys/fs/binfmt_misc/ ${tmpdir}/sys/fs/binfmt_misc
mount --move ${tmpdir} /proc
rm -rf $tmpdir
```

afbjorklund: Took a while, but doing the implementation as suggested (except rmdir, not rm -rf). Seems to work fine!

DrDaveD (changes requested):
Ok, this code is looking pretty good now.

It just needs you to add yourself to CONTRIBUTORS.md and an entry in CHANGELOG.md.  For the latter add above the 1.4.0 Release Candidate 1 line:
```
## v1.4.x changes

Changes since 1.4.0-rc.1

- your entry

```

I prefer one squashed commit, but if that's not easy for you I'll just do a squash & merge.

I also think we could and should add some binfmt_misc e2e tests, including trying it with the different fakeroot modes.  If that seems like too much of a challenge just make another issue for that and have it refer to this PR.

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ The Singularity Project has been
 and re-branded as Apptainer.
 For older changes see the [archived Singularity change log](https://github.com/apptainer/singularity/blob/release-3.8/CHANGELOG.md).
 
+## v1.4.x changes
+
+Changes since 1.4.0-rc.1
+
+- Make sure to mount binfmt_misc on the new proc filesystem created
+  when running with fakeroot, to make emulated cross-builds possible.
```
DrDaveD, line 13: Please make this more user focused, for example like this:
```
- Fix running or building containers of different architectures via
  binfmt_misc with /etc/subuid fakeroot.
```

DrDaveD (changes requested):

DrDaveD:
> Squashing changelogs with the actual changes sounds like it would be a mess for cherry-picks etc, but OK
> 
> Like if a 1.4.0-rc.2 is relased before this is merged, then it would need to be rebased (only for CHANGELOG)

Actually after this I will want you to submit a cherry-picked PR to the release-1.4 branch, but we'll want all the same things there so it's easier to cherry-pick with fewer commits.  Then when we release 1.4.0-rc.2 we'll replace that top line.

DrDaveD (approved):

DrDaveD:
Please make another PR cherry-picking this one to the release-1.4 branch.

PR #2732, opened 2025-01-22
Work around c23 bool restriction

DrDaveD:
I ran into a problem compiling on Fedora rawhide where it complained about `true` and `false` being restricted keywords in c23.  This works around the problem by changing things to start with upper case.

### DIFF
--- a/cmd/starter/c/include/starter.h
+++ b/cmd/starter/c/include/starter.h
@@ -74,9 +74,9 @@ enum goexec {
 #endif
 
 typedef enum {
-    false,
-    true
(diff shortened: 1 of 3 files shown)


afbjorklund:
Can't you include `<stdbool.h>` instead ?

DrDaveD:
> Can't you include `<stdbool.h>` instead ? (or otherwise TRUE and FALSE, oldskool)

I considered that, but concluded that it would still fool the go code.

cclerget (approved):

PR #2791, opened 2025-02-19
Bundle squashfs-tools with apptainer libexec

### LINKED ISSUE #2785: Consider bundling a `mksquashfs` with more compression options etc
One _could_ add some information to documentation and packaging, on how to build "[squashfs-tools](https://github.com/plougher/squashfs-tools)" with more compression support.

On some platforms, the support for alternatives to gzip have been disabled by packaging (or age) and excludes the newer alternatives.

The implementation should be as simple as putting it first in the PATH, possibly choosing a different name than `/usr/bin/mksquashfs`?

The software normally builds without too many complications, all that is needed is to override some variables for better support...

* https://github.com/apptainer/apptainer/issues/1022

```
Makefile:GZIP_SUPPORT ?= 1
Makefile:XZ_SUPPORT ?= 0
Makefile:LZO_SUPPORT ?= 0
Makefile:LZ4_SUPPORT ?= 0
Makefile:ZSTD_SUPPORT ?= 0
```

Having zstd and lz4 "always" available would be nice, assuming that one is also providing a `squashfuse_ll` with similar support.

They both feature much better (5-10x) performance than gzip, and in the case of zstd it also provides compression similar to xz.

* https://github.com/apptainer/apptainer/pull/2707

* https://github.com/apptainer/apptainer-userdocs/issues/308

----

See also: https://facebook.github.io/zstd/#benchmarks

Compressor name	| Ratio	| Compression	| Decompress.
--- | --- | --- | ---
zlib 1.2.11 -1	| 2.743	| 95 MB/s	| 400 MB/s
zstd 1.5.6 -1	| 2.887	| 510 MB/s	| 1580 MB/s
zstd 1.5.6 --fast=1	| 2.437	| 545 MB/s	| 1890 MB/s
zstd 1.5.6 --fast=3	| 2.239	| 650 MB/s	| 2000 MB/s
lzo1x 2.10 -1	| 2.106	| 650 MB/s	| 825 MB/s
lz4 1.9.4	| 2.101	| 700 MB/s	| 4000 MB/s

Since 1.0: https://engineering.fb.com/2016/08/31/core-infra/smaller-and-faster-data-compression-with-zstandard/

CLI | Ratio | Comp Speed | Decomp Speed
--- | --- | --- | ---
`gzip` | 3.11 | 19.66 MB/s | 125.55 MB/s
`zstd` | 3.14 | 100.1 MB/s | 459.40 MB/s

afbjorklund:
For access to all compression methods (same as `squashfuse_ll`),
and for the new -percentage option for showing a progress bar.

This means that the squashfs-tools (or "squashfs") package
no longer needs to be pre-installed from the distribution.

Bundle squashfs-tools with apptainer, similar to squashfuse.

Enable all compression methods with their libraries installed.

 - Fixes #2785

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -323,7 +323,7 @@ jobs:
           go-version: 1.22.7
 
       - name: Fetch deps
-        run: sudo apt-get -q update && sudo DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential squashfs-tools squashfuse fuse-overlayfs fakeroot fuse2fs libseccomp-dev cryptsetup dbus-user-session
+        run: sudo apt-get -q update && sudo DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential squashfuse fuse-overlayfs fakeroot fuse2fs libseccomp-dev cryptsetup dbus-user-session
 
       - name: Build and install Apptainer
         run: |
@@ -353,7 +353,7 @@ jobs:
           go-version: 1.22.7
 
       - name: Fetch deps
-        run: sudo apt-get -q update && sudo DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential squashfs-tools libseccomp-dev cryptsetup dbus-user-session
+        run: sudo apt-get -q update && sudo DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential libseccomp-dev cryptsetup dbus-user-session
 
       - name: Build and install Apptainer
         run: |
@@ -402,8 +402,8 @@ jobs:
         run: |
           set -e
           sudo apt-get -q update
-          sudo DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential uidmap squashfs-tools squashfuse fuse-overlayfs fakeroot fuse2fs libseccomp-dev cryptsetup dbus-user-session
-          sudo DEBIAN_FRONTEND=noninteractive apt-get install -y autoconf automake libtool pkg-config libfuse3-dev zlib1g-dev
+          sudo DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential uidmap squashfuse fuse-overlayfs fakeroot fuse2fs libseccomp-dev cryptsetup dbus-user-session
+          sudo DEBIAN_FRONTEND=noninteractive apt-get install -y autoconf automake libtool pkg-config libfuse3-dev zlib1g-dev liblzo2-dev liblz4-dev liblzma-dev libzstd-dev
 
       - name: Download, compile, and install dependent packages
         if: env.run_tests
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -22,7 +22,6 @@ sudo apt-get install -y \
     libseccomp-dev \
     pkg-config \
     uidmap \
-    squashfs-tools \
     fakeroot \
     cryptsetup \
     tzdata \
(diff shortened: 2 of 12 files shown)


DrDaveD (changes requested):
What's here looks pretty good but also look for changes in .github/workflows/*.yml.  Of course also look at the ci failures because they're not all fixable in the github action workflows.

INSTALL.md
```diff
@@ -208,7 +206,8 @@ sudo apt-get install -y autoconf automake libtool pkg-config libfuse3-dev \
 On RHEL or derivatives:
 
 ```sh
-sudo dnf install -y fuse3-devel lzo-devel lz4-devel
+sudo dnf install -y fuse3-devel zlib-devel lzo-devel lz4-devel \
+    xz-devel libzstd-devel
```
DrDaveD, line 210: These 3 additional libraries were not explicitly listed because they're already included by
```
dnf groupinstall -y 'Development Tools'
```
So I don't think they need to be here or in scripts/ci-rpm-build-test.

DrDaveD (changes requested):

INSTALL.md
afbjorklund: Seems a bit arbitrary to me, hope they don't change it then. But OK, will leave as it was.

.github/workflows/ci.yml
```diff
@@ -323,7 +323,7 @@ jobs:
           go-version: 1.22.7
 
       - name: Fetch deps
-        run: sudo apt-get -q update && sudo DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential squashfs-tools squashfuse fuse-overlayfs fakeroot fuse2fs libseccomp-dev cryptsetup dbus-user-session
+        run: sudo apt-get -q update && sudo DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential fakeroot libseccomp-dev cryptsetup dbus-user-session
```
DrDaveD, line 326: There are more places in .github/workflows/ci.yml where these packages should be deleted.

DrDaveD (changes requested):

.github/workflows/ci.yml
afbjorklund: Okay, will do the cleanup for that separately I think.

tools/install-unprivileged.sh
```diff
@@ -328,7 +328,7 @@ mkdir tmp
 cd tmp
 
 if [ "$DIST" = el7 ]; then
-	OSUTILS="$OSUTILS lzo squashfs-tools"
+	OSUTILS="$OSUTILS lzo"
```
DrDaveD, line 331: It was a good catch to see that this was being included here, but the problem is that this script needs to also support installing older versions of apptainer as well.   You could either leave this in for now and add a comment to remove it later, or you could test to see if the main install has already installed mksquashfs.

Then the line `mv tmp/usr/*bin/*squashfs utils/libexec` below shouldn't be removed but just made to be optional like the lines just above and below it.

.github/workflows/ci.yml
DrDaveD: If you'd like to leave it separate the easiest way is to put it an another commit in this PR.

DrDaveD (changes requested):

tools/install-unprivileged.sh
afbjorklund: Leaving the install-unprivileged.sh script as-is

DrDaveD (approved):

DrDaveD:
Please submit a PR to cherry-pick this to release-1.4 and another to remove installation of the other bundled tools in the CI.

PR #2891, opened 2025-04-07
refine the error message of container creation failure when apptainer-suid is not installed

### LINKED ISSUE #2871: Using a network bridge as non-root user doesn't work on Ubuntu 24.04.
I am running a container with `--net --network=bridge` as non-root users.  In the past this worked with the following steps:

```
sudo apptainer config global --set "allow net groups" foo
sudo apptainer config global --set "allow net networks" bridge
```
Now users of the group "foo" could run a container with `--net --network=bridge`.

Since upgrading the host system to Ubuntu 24.04 (previously it was 18.04), this now results in an error.

## Steps to reproduce

```
sudo apptainer config global --set "allow net groups" foo
sudo apptainer config global --set "allow net networks" bridge
apptainer pull docker://ubuntu:24.04  # I think it doesn't really matter which image, I get the same with alpine

# run this with a user in the "foo" group:
apptainer shell --net --network=bridge ubuntu_24.04.sif
```

## Expected behaviour

The container is created with the network bridge.

## Actual behaviour

The `apptainer shell` command fails with 

> INFO:    Terminating squashfuse_ll after timeout
> INFO:    Timeouts can be caused by a running background process
> ERROR:   could not delete networks: plugin type="firewall" failed (delete): could not initialize iptables protocol 0: exec: "iptables": executable file not found in $PATH
> FATAL:   container creation failed: invalid argument

## Additional information

- I'm using Apptainer 1.4.0
- Host OS is Ubuntu 24.04
- iptables is installed on the host (located in `/usr/sbin/iptables`)
- I tried installing iptables in the container as well but it didn't help
- It _does_ work when running with sudo, so maybe only a PATH issue with `/usr/sbin` for non-root users?
- It is working on Ubuntu 18.04/Apptainer 1.0.3

Not sure if relevant but I gave it a try with Apptainer 1.0.3 on Ubuntu 24.04.  It also fails but with a different error:

> ERROR:   could not delete networks: plugin type="firewall" failed (delete): could not initialize iptables protocol 0: could not get iptables version: exit status 111
> FATAL:   container creation failed: plugin type="bridge" failed (add): failed to locate iptables: could not get iptables version: exit status 111

JasonYangShadow:

refine the error message of container creation failure when apptainer-suid is not installed

 - Fixes #2871

### UserNS enabled case
apptainer is built with `--without-suid`
```
➜  test apptainer shell --net --network=bridge ubuntu_24.04.sif
INFO:    Cleanup error: while unmounting /usr/local/var/apptainer/mnt/session/final directory: no such file or directory, while unmounting /usr/local/var/apptainer/mnt/session/rootfs directory: no such file or directory
FATAL:   container creation failed: network requires root or a suid installation with /etc/subuid --fakeroot; non-root users can only use --network=none unless permitted by the administrator
```

### suid enabled case
apptainer is built with `--with-suid`
```
➜  test apptainer shell --net --network=bridge ubuntu_24.04.sif 
Apptainer> exit
exit
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
   For instance RHEL 9 uses v2 and RHEL 10 uses v3 as their default values.
 - Fix a bug introduced in 1.4.0 that caused arm64 to be mis-converted to arm64v8
   and resulted in a failure when pulling OCI containers.
+- Add a clear error message if someone tries to use privileged network options
+  while not using setuid mode.
 
 Changes since 1.4.0
 
--- a/internal/pkg/runtime/engine/apptainer/container_linux.go
+++ b/internal/pkg/runtime/engine/apptainer/container_linux.go
@@ -2725,7 +2725,7 @@ func (c *container) prepareNetworkSetup(system *mount.System, pid int) (func(con
 	}
 
 	allowedNetUnpriv := false
-	if euid != 0 && !forceFakerootNet {
+	if euid != 0 && !forceFakerootNet && !c.userNS {
 		// Is the user permitted in the list of unpriv users / groups permitted to use CNI?
 		allowedNetUser, err := user.UIDInList(euid, c.engine.EngineConfig.File.AllowNetUsers)
 		if err != nil {


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
   older CPUs, although it can have a cost in performance on newer CPUs.
   It is still possible to set `GOAMD64` to a newer microarchitecture (v2+).
   For instance RHEL 9 uses v2 and RHEL 10 uses v3 as their default values.
+- Refine the error message of container creation failure when apptainer-suid
```
DrDaveD, line 20: A better message for this fix would be:
```
- Add a clear error message if someone tries to use privileged network options
  while not using setuid mode.
```
However I don't think this is the correct fix, see below.

internal/pkg/runtime/engine/apptainer/container_linux.go
```diff
@@ -2804,7 +2804,7 @@ func (c *container) prepareNetworkSetup(system *mount.System, pid int) (func(con
 			if euid != 0 {
 				dropPrivilege, err := priv.Escalate()
 				if err != nil {
-					return err
+					return fmt.Errorf("while escalating the privilege of container process, err: %s, this is usually caused by not installing apptainer-suid", err)
```
DrDaveD, line 2807: It shouldn't have been able to get that far.  There was already a check [earlier in that function](https://github.com/apptainer/apptainer/blob/release-1.4/internal/pkg/runtime/engine/apptainer/container_linux.go#L2758) that should have caught this condition.  It looks like there's a bug there, please investigate.

DrDaveD (changes requested):

DrDaveD:
`allowedNetUnpriv` is true and I don't think it should ever be set to true unless it is a suid installation without the `--userns` option.  I *think* in that function a suid installation is equivalent to `c.userNS`; please verify.  If so, that whole section calculating `allowedNetUnpriv` should probably have `&& !c.userNS`.

internal/pkg/runtime/engine/apptainer/container_linux.go
JasonYangShadow: thanks! I updated the code.

DrDaveD (approved):

DrDaveD:
Please cherry-pick to release-1.4

PR #2911, opened 2025-04-13
Add buildkit bootstrap to build Dockerfile as OCI

### LINKED ISSUE #2910: Build Dockerfile directly from apptainer bootstrap
Some people have a workflow where they update a Dockerfile, and then use that as a bootstrap...

This requires either using docker-daemon, or pushing the results to a registry (over the network?)

It could be useful if apptainer could do both steps, and create a sif image from the build context.

This also allows using a modern build framework (BuildKit), instead of legacy docker or podman.

https://github.com/moby/buildkit

```
buildctl build \
    --frontend=dockerfile.v0 \
    --local context=. \
    --local dockerfile=.

buildctl build ... --output type=oci,dest=path/to/output.tar
```

----

Here is what it could experimentally look like:

```
Bootstrap: buildkit
From: build
```

Where "build" is a path to a directory with the Dockerfile, and optionally any other "build context".

The entire directory is transferred to the buildkit worker, and the result is saved as an `oci-archive`.

There is a build cache, so only those build steps that actually _need_ to be rebuilt are executed again.

```Dockerfile
FROM alpine
RUN true
```

```console
$ apptainer build --force buildkit.sif buildkit.def
INFO:    Starting build...
INFO:    Building OCI image...
[+] Building 0.5s (6/6) FINISHED                                                                                        
 => [internal] load build definition from Dockerfile                                                               0.0s
 => => transferring dockerfile: 58B                                                                                0.0s
 => [internal] load metadata for docker.io/library/alpine:latest                                                   0.4s
 => [internal] load .dockerignore                                                                                  0.0s
 => => transferring context: 2B                                                                                    0.0s
 => [1/2] FROM docker.io/library/alpine:latest@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3  0.0s
 => => resolve docker.io/library/alpine:latest@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3  0.0s
 => CACHED [2/2] RUN true                                                                                          0.0s
 => exporting to oci image format                                                                                  0.0s
 => => exporting layers                                                                                            0.0s
 => => exporting manifest sha256:43ea733cb28a8fb3b7076fc319053818f34468d4ef76722f2b8bf2aa1996167f                  0.0s
 => => exporting config sha256:d128407b81aff69eb707920fab5960c2cd6c258e7dbfe4decb3234fb7d3ea10f                    0.0s
 => => sending tarball                                                                                             0.0s
INFO:    Extracting OCI image...
INFO:    Inserting Apptainer configuration...
INFO:    Creating SIF file...
[=============================================================================================================] 100 % 0s
INFO:    Build complete: buildkit.sif
```

The requirement is to install BuildKit, and start the BuildKit server (possibly rootless).

It is possible to have workers for different architectures, to do multi-arch image builds.

```console
$ buildctl debug workers
ID				PLATFORMS
t4ddg3lhk5x535vl2ahk78bbv	linux/amd64,linux/amd64/v2,linux/amd64/v3,linux/arm64,linux/riscv64,linux/ppc64,linux/ppc64le,linux/s390x,linux/386,linux/mips64le,linux/mips64,linux/arm/v7,linux/arm/v6
tjjkniyozr2aie96sx52bq2um	linux/amd64,linux/amd64/v2,linux/amd64/v3,linux/arm64,linux/riscv64,linux/ppc64,linux/ppc64le,linux/s390x,linux/386,linux/mips64le,linux/mips64,linux/arm/v7,linux/arm/v6
```

Most of the code is shared with the OCI bootstrap, except for the new "buildImage" call.

afbjorklund:

This will call out to `buildctl`, to contact `buildkitd`.

The location can be set by using BUILDKIT_HOST env var.

It is mostly a proof-of-concept on top of scratch, so it needs more features from the regular OCI (oci-archive) build.

Preferrably they could _share_ all that code, by splitting up the OCI bootstrap into more steps that could be called ?

 - Fixes #2910


DrDaveD:
Let's discuss this at a community meeting.

One thought I had is that the Apptainer definition file is so simple, I wonder if there should additionally be an implied buildkit bootstrap if an input directory on the build command line contains a Dockerfile, instead of treating it as a sandbox.

Also, I think that if the required executables (buildctl or runc) are missing or daemon (buildkitd) isn't running, there should be clear error messages saying what to do about it.

There will need to be tests.

afbjorklund:
> Also, I think that if the required executables (buildctl or runc) are missing or daemon (buildkitd) isn't running, there should be clear error messages saying what to do about it.

The error messages are similar to when `docker` is not installed, or when `dockerd` is not running. But there could be more explicit tests, even if it requires some testing and parsing.

afbjorklund:
> One thought I had is that the Apptainer definition file is so simple, I wonder if there should additionally be an implied buildkit bootstrap if an input directory on the build command line contains a Dockerfile, instead of treating it as a sandbox.

It would need to look for both Dockerfile and "Containerfile", but I added a parameter for e.g. Dockerfile-foo

But BuildKit also allows you to come with your own syntax, so it could be possible to add support for any of:

https://github.com/moby/buildkit#exploring-llb

* Dockerfile (currently included by default)
* [Buildpacks](https://github.com/tonistiigi/buildkit-pack)
* [Mockerfile](https://matt-rickard.com/building-a-new-dockerfile-frontend/)
* [Gockerfile](https://github.com/po3rin/gockerfile)
... etc ...

DrDaveD (changes requested):
Needs a rebase now that your other PR was merged

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ For older changes see the [archived Singularity change log](https://github.com/a
   Unfortunately the feature does not work with "docker-archive" or "docker-daemon".
 - Change the default `arm` variant to `v7`, and stop using the GOARM environment
   variable. The variables GOOS, GOARCH and GOARM are only used when building.
+- Add new bootstrap `buildkit:` for building SIF images directly from
```
DrDaveD, line 16: Also mention the buildkit:// URL.

DrDaveD (changes requested):
Please improve the error messages for when buildctl and docker aren't present.

Also make another issue calling for tests.

afbjorklund:
> Please improve the error messages for when buildctl and docker aren't present.

Now it outputs a more friendly error message, if you forgot  to start the daemon:

`FATAL:   Cannot connect to the BuildKit daemon at unix:///run/buildkit/buildkitd.sock. Is the buildkit daemon running?`

Instead of the default error, which could be interpreted as more cryptic perhaps:

`error: listing workers for Build: failed to list workers: Unavailable: connection error: desc = "transport: error while dialing: dial unix /run/buildkit/buildkitd.sock: connect: no such file or directory"`

---

The other error messages are more similar, like if you don't have root access:

`permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.45/version": dial unix /var/run/docker.sock: connect: permission denied`

`error: listing workers for Build: failed to list workers: Unavailable: connection error: desc = "transport: error while dialing: dial unix /run/buildkit/buildkitd.sock: connect: permission denied"`

Then you can either wrap it in `sudo apptainer`, or set up rootless per [instructions](https://github.com/moby/buildkit/blob/master/docs/rootless.md).

https://apptainer.org/docs/user/main/docker_and_oci.html (Docker, for comparison)

```shell
export BUILDKIT_HOST=unix://$XDG_RUNTIME_DIR/buildkit/buildkitd.sock
```

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
   Unfortunately the feature does not work with "docker-archive" or "docker-daemon".
 - Change the default `arm` variant to `v7`, and stop using the GOARM environment
   variable. The variables GOOS, GOARCH and GOARM are only used when building.
+  The full buildkit log file is included in the image, for traceability.
+  It is also showed on the console, as a progress update while building.
```
DrDaveD, line 17: Looks like the beginning of the changelog entry got lost here.

internal/pkg/build/sources/conveyorPacker_buildkit.go
```diff
@@ -0,0 +1,348 @@
...
+		defer buildlog.Close()
+
+		err = buildkitBuild(ctx, env, platform, output, cp.bk, buildlog)
+		if err != nil {
+			return err
+		}
+
+		ref = "oci-archive:" + output
+	} else {
+		err := dockerBuild(ctx, env, platform, output, cp.bk)
```
DrDaveD, line 327: What happens if neither buildctl nor docker are installed?  I think there should be an error message explaining that one or the other is needed.

DrDaveD (changes requested):

internal/pkg/build/sources/conveyorPacker_buildkit.go
afbjorklund: Currently you will see the last error. `exec: "docker": executable file not found in $PATH`

It would be possible to use the docker client API instead of the docker client CLI, but it would make the fallback code different from the buildkit code (that is using CLI instead of API, to cut down on the dependency tree...)
It would also create yet another dependency on the github.com/docker/docker (moby), which is currently there for the `docker-daemon` support (only), that is quite big and might become unwanted at some future time.

But I think just catching the error and doing another error message would be good enough...

As mentioned above, the error message when the server is not running is not much better (more like: worse) than when the client is missing. And most users would install both of them at the same time anyway, even though it is theoretically possible to have a local client talking to a remote server on some other machine or cluster. The permission errors are also likely to happen, but resolution can be either "sudo" or "rootless".

`FATAL:   While performing build: packer failed to pack: while building image: exec: "docker": executable file not found in $PATH`
`FATAL:   While performing build: conveyor failed to get: error saving the docker archive file: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?`

CHANGELOG.md
afbjorklund: Probably a victim of the rebase, when the code was shuffled.

internal/pkg/build/sources/conveyorPacker_buildkit.go
afbjorklund: Added an error message.

`FATAL:   While performing build: packer failed to pack: while building image: neither "buildctl" nor "docker" could be found in the PATH, please install one of them to build`

internal/pkg/build/sources/conveyorPacker_buildkit.go
```diff
@@ -0,0 +1,352 @@
+// Copyright (c) Contributors to the Apptainer project, established as
+//   Apptainer a Series of LF Projects LLC.
+//   For website terms of use, trademark policy, privacy policy and other
+//   project policies see https://lfprojects.org/policies
+// Copyright (c) 2020, Control Command Inc. All rights reserved.
+// Copyright (c) 2018-2024, Sylabs Inc. All rights reserved.
```
DrDaveD, line 6: Is there anything significant in this file that was copied from other files, or is it essentially from scratch?

DrDaveD (approved):
Is there anything else you want to change before this is merged?

internal/pkg/build/sources/conveyorPacker_buildkit.go
afbjorklund: It is a copy of (big) parts of "net", and some functions (from it) are called directly as well...

It would be possible to refactor those into common functions, but since they were hardcoded in "net"

afbjorklund:
> So which way do you want to go for the tests?

We can add the tests later, if that is OK with you.

DrDaveD:
Ok, please make an issue for that.

PR #2922, opened 2025-04-22
Record metadata for original image used from oci

### LINKED ISSUE #2764: Reliable way to reproduce a .sif image
If I build an image file using the `latest` tag, there doesn't appear to be reliable way to discover the version (or ideally the SHA digest) that was used so that it can be reproduced at a later date.

My initial attempt was to use `apptainer inspect` command to get the `org.label-schema.version` label, but this appears to be missing from some images.

afbjorklund:

Add two new fields to the metadata, when building from `docker:`

`org.opencontainers.image.base.name`
`org.opencontainers.image.base.digest`

The fields used for Apptainer are the same that is used in the OCI image specification:

https://github.com/opencontainers/image-spec/blob/main/annotations.md

These record the information that is available, but hidden away:

```console
$ docker pull alpine
Using default tag: latest
latest: Pulling from library/alpine
Digest: sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c
Status: Image is up to date for alpine:latest
docker.io/library/alpine:latest
```

In addition to the basic metadata that was previously recorded:

```
org.label-schema.usage.singularity.deffile.bootstrap: docker
org.label-schema.usage.singularity.deffile.from: alpine
```

```
org.opencontainers.image.base.digest: sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c
org.opencontainers.image.base.name: docker.io/library/alpine:latest
```

We don't need to record the Id (ConfigFile) or the Manifest checksums:

docker inspect
```
        "Id": "sha256:aded1e1a5b3705116fa0a92ba074a5e0b0031647d9c315983ccba2ee5428ec8b",
        "RepoTags": [
            "alpine:latest"
        ],
        "RepoDigests": [
            "alpine@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c"
        ],
```

podman inspect
```
          "Id": "aded1e1a5b3705116fa0a92ba074a5e0b0031647d9c315983ccba2ee5428ec8b",
          "Digest": "sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c",
          "RepoTags": [
               "docker.io/library/alpine:latest"
          ],
          "RepoDigests": [
               "docker.io/library/alpine@sha256:1c4eef651f65e2f7daee7ee785882ac164b02b78fb74503052a26dc061c90474",
               "docker.io/library/alpine@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c"
          ],
```

But we should record the full name of the image, and the full digest.

https://hub.docker.com/layers/library/alpine/latest/images/sha256-a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c

https://www.redhat.com/en/blog/be-careful-when-pulling-images-short-name

Note: 1c4eef651f65 is the amd64 manifest, that was picked from the image index

```
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 1022,
         "digest": "sha256:1c4eef651f65e2f7daee7ee785882ac164b02b78fb74503052a26dc061c90474",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
```

But we should record the digest that is the same for all platforms.

```
org.label-schema.build-arch: amd64
org.label-schema.build-date: Tuesday_22_April_2025_18:46:13_CEST
```

 - Fixes #2764

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 ## v1.5.x changes
 
+- Record image digest metadata (sha256 from `RepoDigests`), for OCI registry images.
+  Also add the image name (ref) of the image from "docker", with registry and tag.
+  This is useful for traceability, when using `docker.io` or a tag like `latest`.
+  Unfortunately the feature does not work with "docker-archive" or "docker-daemon".
+
 ## v1.4.x changes
 
 - Check for existence of `/run/systemd/system` when verifying cgroups can be
--- a/internal/pkg/build/metadata.go
+++ b/internal/pkg/build/metadata.go
@@ -341,6 +341,12 @@ func addBuildLabels(labels map[string]string, b *types.Bundle) error {
 		}
 	}
 
+	// Digest of image
+	if b.Opts.Tag != "" && b.Opts.Digest != "" {
+		labels["org.opencontainers.image.base.name"] = b.Opts.Tag
+		labels["org.opencontainers.image.base.digest"] = b.Opts.Digest
+	}
+
 	// Architecture of build
 	buildarch := oci.ArchMap[b.Opts.Arch]
 	labels["org.label-schema.build-arch"] = buildarch.Arch
--- a/internal/pkg/build/oci/oci.go
+++ b/internal/pkg/build/oci/oci.go
@@ -171,6 +171,23 @@ func parseURI(uri string) (types.ImageReference, *GoArch, error) {
 	return imgRef, arch, err
 }
 
+// RepoDigest returns the (tag and) digest of a docker image
+func RepoDigest(ctx context.Context, uri string, topts *ociimage.TransportOptions) (tag string, digest string, err error) {
+	ref, _, err := parseURI(uri)
+	if err != nil {
+		return "", "", fmt.Errorf("unable to parse image name %v: %v", uri, err)
+	}
+	if ref.Transport().Name() != "docker" {
+		return "", "", nil
+	}
+	// nolint:staticcheck
+	d, err := docker.GetDigest(ctx, ociimage.SystemContextFromTransportOptions(topts), ref)
+	if err != nil {
+		return "", "", err
+	}
+	return ref.DockerReference().String(), d.String(), nil
+}
+
 // ImageDigest obtains the digest of a uri's manifest
 func ImageDigest(ctx context.Context, uri string, topts *ociimage.TransportOptions) (digest string, err error) {
 	ref, arch, err := parseURI(uri)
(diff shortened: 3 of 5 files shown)


internal/pkg/build/metadata.go
```diff
@@ -341,6 +341,12 @@ func addBuildLabels(labels map[string]string, b *types.Bundle) error {
 		}
 	}
 
+	// Digest of image
+	if b.Opts.Tag != "" && b.Opts.Digest != "" {
+		labels["org.label-schema.image-tag"] = b.Opts.Tag
+		labels["org.label-schema.image-digest"] = b.Opts.Digest
```
DrDaveD, line 347: Are these label names made up or are they part of some standard?

afbjorklund: They are made up, or well the "tag" and "digest" is part of the Docker format

afbjorklund: Note that both words are highly ambiguous, but here they are both clearly defined.

"Tag" is the **full** reference name, and "Digest" is the **full** image digest (with algorithm).

DrDaveD: I don't think it's OK to make up org.label-schema namespace labels, because those are defined by `org.label-schema`.  It should be allowed to make `org.apptainer` labels.  I see the `org.label-schema` space is now [deprecated](http://label-schema.org/rc1/) and replaced by [`org.opencontainers.image`](https://github.com/opencontainers/image-spec/blob/main/annotations.md).  That says that that namespace MUST NOT be used by extensions.  On the other hand it looks like singularity and apptainer have already added on to `org.label-schema.usage` with sublabels.

afbjorklund: I suppose that we could also add `org.opencontainers.image.created` in addition to the old build-date?

But there doesn't seem to exist any field for the full `platform`, I can't even find any reference to build-arch

DrDaveD (changes requested):

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ For older changes see the [archived Singularity change log](https://github.com/a
   while not using setuid mode.
 - Allow multi-arch oci-archive files that have a nested index with the manifest.
   This is the default format (both for Docker and OCI) when using `nerdctl save`.
+- Record image digest metadata (sha256 from `RepoDigests`), for OCI registry images.
+  Also add the image name (ref) of the image from "docker", with registry and tag.
+  This is useful for traceability, when using `docker.io` or a tag like `latest`.
+  Unfortunately the feature does not work with "docker-archive" or "docker-daemon".
```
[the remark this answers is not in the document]
afbjorklund: I will move the changelog to the 1.5.0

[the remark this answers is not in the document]
afbjorklund: I thought it would make a nice match with the Dockerfile build log that I added in the other PR

DrDaveD (approved):

PR #2962, opened 2025-05-10
Add implementation of ipfs gateway download

### LINKED ISSUE #2961: Add support for downloading images from IPFS through a HTTP gateway
Previously there was some discussion recarding IPFS support through `/ipfs`:

* https://github.com/apptainer/apptainer/issues/134#issuecomment-2183979619

But that requires mounting a fuse filesystem, which might not be doable.

Another approach is to link with go-ipfs or to call the `ipfs get` command...

----

But it would also be possible to support it through HTTP, like in curl 8.4.0+:

https://blog.ipfs.tech/ipfs-uri-support-in-curl/

https://curl.se/docs/ipfs.html

Then it only needs to rewrite the URL, and use the same code as in "net".

A bonus is that the sha256 is in the URL, so it doesn't need to do any HEAD call...

```console
$ ipfs add --cid-version=1 --wrap-with-directory alpine.sif 
added bafybeigmf4qbfae5zyacg6tp6q67fcb7w3ouw7gmvtbaxqpg5rmmznaf6i alpine.sif
added bafybeibpisnayrx6rb3eicvsxu6lench2xogxurrgonj67jujgofa2bpge 
 3.41 MiB / 3.41 MiB [=========================================================================================] 100.00%
$ apptainer run ipfs://bafybeigmf4qbfae5zyacg6tp6q67fcb7w3ouw7gmvtbaxqpg5rmmznaf6i
INFO:    Downloading network image
3.4MiB / 3.4MiB [==============================================================================================] 100 %0s
Apptainer> 
```

The CID can be [converted](https://cid.ipfs.tech/#bafybeigmf4qbfae5zyacg6tp6q67fcb7w3ouw7gmvtbaxqpg5rmmznaf6i) to a hex sha256, so that the cache directory looks "normal".

```
$ ipfs-inspect bafybeigmf4qbfae5zyacg6tp6q67fcb7w3ouw7gmvtbaxqpg5rmmznaf6i
base32-cidv1-dag-pb-(sha2-256:32:CC2F2012809DCE00237A6FF43DF2883FB6DD4B7CCCACC20BC1E6EC58CCB405F2)
```

It is a [multihash](https://multiformats.io/), but only the base32 decoding is required - the rest can be CIDv1-only.

Only the gateway needs to be connected to the p2p network (which can be made private).

afbjorklund:

Set IPFS_GATEWAY to a shared HTTP gateway, or run `ipfs daemon`.

Upload images like so: `ipfs add --cid-version 1 -w lolcow.sif`.

 - Fixes #2961

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ For older changes see the [archived Singularity change log](https://github.com/a
   set is "all".
 - debugsource rpms are now generated in addition to debuginfo rpms on
   RHEL-derived and Fedora operating systems.
+- Add support for downloading SIF images from an IPFS peer-to-peer cluster using
+  a HTTP gateway (similar to the existing support for IPFS in the `curl` tool).
+  The location of the address can set in the `IPFS_GATEWAY` environment variable,
+  any local gateway will be picked up automatically (e.g. `http://127.0.0.1:8080`)
 
 ## v1.4.x changes
 
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/apptainer/apptainer/docs"
 	"github.com/apptainer/apptainer/internal/pkg/cache"
+	"github.com/apptainer/apptainer/internal/pkg/client/ipfs"
 	"github.com/apptainer/apptainer/internal/pkg/client/library"
 	"github.com/apptainer/apptainer/internal/pkg/client/net"
 	"github.com/apptainer/apptainer/internal/pkg/client/oci"
@@ -105,6 +106,10 @@ func handleOras(ctx context.Context, imgCache *cache.Handle, cmd *cobra.Command,
 	return oras.Pull(ctx, imgCache, pullFrom, tmpDir, ociAuth, noHTTPS, reqAuthFile)
 }
 
+func handleIpfs(ctx context.Context, imgCache *cache.Handle, pullFrom string) (string, error) {
+	return ipfs.Pull(ctx, imgCache, pullFrom, tmpDir)
+}
+
 func handleLibrary(ctx context.Context, imgCache *cache.Handle, pullFrom string) (string, error) {
 	r, err := library.NormalizeLibraryRef(pullFrom)
 	if err != nil {
@@ -160,6 +165,8 @@ func replaceURIWithImage(ctx context.Context, cmd *cobra.Command, args []string)
 		image, err = handleLibrary(ctx, imgCache, args[0])
 	case uri.Oras:
 		image, err = handleOras(ctx, imgCache, cmd, args[0])
+	case uri.IPFS:
+		image, err = handleIpfs(ctx, imgCache, args[0])
 	case uri.Shub:
 		image, err = handleShub(ctx, imgCache, args[0])
 	case ociimage.SupportedTransport(t):
--- a/docs/content.go
+++ b/docs/content.go
@@ -67,6 +67,7 @@ Enterprise Performance Computing (EPC)`
       docker://   a Docker/OCI registry (default Docker Hub)
       shub://     an Apptainer registry (default Singularity Hub)
       oras://     an OCI registry that holds SIF files using ORAS
+      ipfs://     an IPFS cluster, using a HTTP gateway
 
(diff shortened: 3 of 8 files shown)


internal/pkg/client/net/pull.go
```diff
@@ -111,8 +111,7 @@ func DownloadImage(ctx context.Context, filePath string, netURL string) error {
 	return nil
 }
 
-// pull will pull a http(s) image into the cache if directTo="", or a specific file if directTo is set.
-func pull(ctx context.Context, imgCache *cache.Handle, directTo, pullFrom string) (imagePath string, err error) {
+func head(pullFrom string) (hash string, err error) {
```
DrDaveD, line 114: How does it help to introduce this function if it is only used once?

DrDaveD (changes requested):
Please add the new URL type in appropriate help messages in docs/content.go.

DrDaveD:
Also please add a CHANGELOG.md entry.

internal/pkg/client/net/pull.go
afbjorklund: It was because I wanted to reuse the "pull" code, but in the end I was not able to anyway...

So all it accomplished was to make the functions more similar - but not identical (or shared)

afbjorklund:
Refactored and added a unit test for the the decoding util, from the CIDv1 to the SHA-256...

https://cid.ipfs.tech/#bafybeicn7i3soqdgr7dwnrwytgq4zxy7a5jpkizrvhm5mv6bgjd32wm3q4

When adding the file with --wrap-with-directory, as suggested above, you get two different CID:

https://ipfs.io/ipfs/bafybeicn7i3soqdgr7dwnrwytgq4zxy7a5jpkizrvhm5mv6bgjd32wm3q4/welcome-to-IPFS.jpg

https://ipfs.io/ipfs/bafkreie7ohywtosou76tasm7j63yigtzxe7d5zqus4zu3j6oltvgtibeom?filename=welcome-to-IPFS.jpg

This is similar to how it works in git, when the "tree" gets one SHA-1 and the "blob" gets one SHA-1.

----

In the apptainer implementation, we expect that there is only one SIF file in each directory.

So if you try to download the directory CID, it is created as a symbolic link to the file CID:

```
$ ipfs add --cid-version 1 -w lolcow.sif
added bafybeigsdv6ijdarejvzmx5qwqqpdjsbhuaplpym2nij72dolvmr3i634m lolcow.sif
added bafybeiex3hgq4uslhdpk4iz6ni2oel2o6dt3jqd45lb6vz6ka7krw7rqca 
$ apptainer run ipfs://bafybeiex3hgq4uslhdpk4iz6ni2oel2o6dt3jqd45lb6vz6ka7krw7rqca/lolcow.sif
INFO:    Downloading network image
$ apptainer run ipfs://bafybeigsdv6ijdarejvzmx5qwqqpdjsbhuaplpym2nij72dolvmr3i634m
```

So the second invocation it used the ipfs cache, since it was referring to the _same_ SIF file:

`~/.apptainer/cache/ipfs/97d9cd0e524b38deae233e6a34e22f4ef0e7b4c07ceac3eae7ca07d51b7e3010 -> d21d7c848c11226b965fb0b420f1a6413d00f5bf0cd3509fe86e5d591da3dbe3`
`~/.apptainer/cache/ipfs/d21d7c848c11226b965fb0b420f1a6413d00f5bf0cd3509fe86e5d591da3dbe3`

We can calculate the checksum of the file in the cache, and see that it matches the CID.

Note: https://docs.ipfs.tech/concepts/content-addressing/#cids-are-not-file-hashes

`sha256:1582da68e80fcfc7f7da5cf0679403a89dbd16e10fe7a8258a00b76033d83ffa`
`multi:bafybeigsdv6ijdarejvzmx5qwqqpdjsbhuaplpym2nij72dolvmr3i634m`

But assuming that we **do** use the same encoding and version and codec, it _will_ match:

```
$ ipfs-inspect bafybeigsdv6ijdarejvzmx5qwqqpdjsbhuaplpym2nij72dolvmr3i634m
base32-cidv1-dag-pb-(sha2-256:256:D21D7C848C11226B965FB0B420F1A6413D00F5BF0CD3509FE86E5D591DA3DBE3)
```

Issue #3012, opened 2025-06-07
unable to install current pre-built Debian package (1.4.1) on current Debian testing (trixie)


davegomboc:
It seems that the .deb is looking specifically for libfuse3-3:amd64.  This package is not available on the current Debian testing: I do have [libfuse3-4:amd64](https://packages.debian.org/trixie/libfuse3-4) and also [libfuse2t64:amd64](https://packages.debian.org/trixie/libfuse2t64) installed.

```
dave@mc25deb:/opt/apptainer$ sudo apt install -y ./apptainer_1.4.1_amd64.deb 
Note, selecting 'apptainer' instead of './apptainer_1.4.1_amd64.deb'
Solving dependencies... Error!
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

Unsatisfied dependencies:
 apptainer : Depends: libfuse3-3 (>= 3.2.3) but it is not installable
Error: Unable to correct problems, you have held broken packages.
Error: The following information from --solver 3.0 may provide additional context:
   Unable to satisfy dependencies. Reached two conflicting decisions:
   1. apptainer:amd64=1.4.1 is selected for install
   2. apptainer:amd64 Depends libfuse3-3 (>= 3.2.3)
      but none of the choices are installable:
      [no choices]
dave@mc25deb:/opt/apptainer$ dpkg --list | grep libfuse
ii  libfuse2t64:amd64                                   2.9.9-9                             amd64        Filesystem in Userspace (library)
ii  libfuse3-4:amd64                                    3.17.2-2                            amd64        Filesystem in Userspace (library) (3.x version)
dave@mc25deb:/opt/apptainer$
```

afbjorklund:
The current packages are built for Debian 11 (oldstable), they should work on 12 (stable) but apparently not on 13.

DrDaveD:
Yes we need to update the packaging to optionally allow libfuse3-4.

DrDaveD:
Come to think of it, although that should be helpful for building from source, I'm not actually sure it will be sufficient for running an older binary on the newer OS version.  Does Debian not have the ability to install older libraries for backward compatibility?

afbjorklund:
I think you get that for the stable releases, but not (minor) for testing?

Just looking at https://packages.debian.org/search?keywords=libfuse

DrDaveD:
[...]

PR #3070, opened 2025-07-23
Some changes to dist/debian

Steffen-Gr:

This PR adds `conffiles`, eventual downloading of the `go` tarball, and minor additions for smoother packaging e.g. with `pbuilder`.

- [-] Make sure all commits are signed-off with `git commit -s`, see the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md).
- [-] Added changes to the [CHANGELOG](https://github.com/apptainer/apptainer/blob/main/CHANGELOG.md), if necessary according to the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md).
- [-] Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md)).
- [-] Added myself as a contributor to the [Contributors File](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTORS.md).

### DIFF
--- a/dist/debian/apptainer.conffiles
+++ b/dist/debian/apptainer.conffiles
@@ -1,0 +1,16 @@
+/etc/apparmor.d/apptainer
+/etc/apptainer/apptainer.conf
+/etc/apptainer/capability.json
+/etc/apptainer/cgroups/cgroups.toml
+/etc/apptainer/dmtcp-conf.yaml
+/etc/apptainer/ecl.toml
+/etc/apptainer/global-pgp-public
+/etc/apptainer/network/00_bridge.conflist
+/etc/apptainer/network/10_ptp.conflist
+/etc/apptainer/network/20_ipvlan.conflist
+/etc/apptainer/network/30_macvlan.conflist
+/etc/apptainer/network/40_fakeroot.conflist
+/etc/apptainer/nvliblist.conf
+/etc/apptainer/remote.yaml
+/etc/apptainer/rocmliblist.conf
+/etc/apptainer/seccomp-profiles/default.json
--- a/dist/debian/control
+++ b/dist/debian/control
@@ -7,14 +7,23 @@ Uploaders:
 Build-Depends:
  debhelper (>= 9),
  dh-autoreconf,
+ devscripts,
  help2man,
  libarchive-dev,
  libssl-dev,
  uuid-dev,
- devscripts,
  libseccomp-dev,
- cryptsetup,
  golang-go (>= 2:1.13~~),
+ libseccomp-dev,
+ uidmap,
+ fakeroot,
+ cryptsetup,
+ tzdata,
+ dh-apparmor,
+ curl,
+ wget,
+ git,
+ libsubid-dev,
  autoconf,
  automake,
  libtool,
@@ -25,7 +34,6 @@ Build-Depends:
  liblz4-dev,
  liblzma-dev,
  libzstd-dev,
- dh-apparmor
 Standards-Version: 3.9.8
 Homepage: http://apptainer.org
 Vcs-Git: https://github.com/apptainer/apptainer.git
@@ -52,7 +60,10 @@ Description: container platform focused on supporting "Mobility of Compute" form
 
 Package: apptainer-suid
 Architecture: any
-Depends: ${misc:Depends}, ${shlibs:Depends}, apptainer (= ${binary:Version})
+Depends:
+ ${misc:Depends},
+ ${shlibs:Depends},
+ apptainer (= ${binary:Version})
 Description: setuid-root portion of apptainer
  This package is the optional setuid-root portion of Apptainer.  
  Without it, Apptainer will run only unprivileged command flows.
--- a/dist/debian/rules
+++ b/dist/debian/rules
@@ -26,6 +26,7 @@ endif
 
 # For locally compiled go
 MINGO_VERSION = $(shell scripts/get-min-go-version)
+TARBALL = go$(MINGO_VERSION).src.tar.gz
 GOROOT = $${TMPDIR:-/tmp}/appdebgo/go
 GOCACHE = $${TMPDIR:-/tmp}/appdebgo/cache
 
@@ -62,6 +63,7 @@ endif
 endif
 
 MAKEPARALLEL = $(shell nproc --ignore=2 || echo 1 )
+
 override pkgdir = debian/tmp
 override apppkgdir = debian/apptainer
 
@@ -74,11 +76,14 @@ override_dh_auto_configure:
 	@export GOPATH=$(GOROOT)/bin PATH=$(GOROOT)/bin:$$PATH GOCACHE=$(GOCACHE); \
 	  set -ex; \
 	  if ! ./mlocal/scripts/check-min-go-version go $(MINGO_VERSION); then \
-	    if [ -d $(GOROOT) ]; then rm -rf $(GOROOT); fi; \
+	    wget -nc -O debian/$(TARBALL) https://dl.google.com/go/$(TARBALL); \
+	    if [ -d $(GOROOT) ]; then \
+	      rm -rf $(GOROOT); \
+	    fi; \
 	    mkdir -p $(GOROOT); \
 	    HERE=$$PWD; \
 	    cd $(GOROOT)/..; \
-	    tar -xf $$HERE/debian/go$(MINGO_VERSION).src.tar.gz; \
+	    tar -xf $$HERE/debian/$(TARBALL); \
 	    cd go/src; \
 	    ./make.bash; \
 	    cd $$HERE; \
--- a/dist/debian/source/include-binaries
+++ b/dist/debian/source/include-binaries
@@ -1,0 +1,6 @@
+# update this with actual filenames
+#debian/e2fsprogs-1.47.2.tar.gz
+#debian/fuse-overlayfs-1.14.tar.gz
+#debian/gocryptfs-2.5.1.tar.gz
+#debian/squashfs-tools-4.6.1.tar.gz
+#debian/squashfuse-0.6.0.tar.gz


dist/debian/rules
```diff
@@ -74,11 +76,14 @@ override_dh_auto_configure:
 	@export GOPATH=$(GOROOT)/bin PATH=$(GOROOT)/bin:$$PATH GOCACHE=$(GOCACHE); \
 	  set -ex; \
 	  if ! ./mlocal/scripts/check-min-go-version go $(MINGO_VERSION); then \
-	    if [ -d $(GOROOT) ]; then rm -rf $(GOROOT); fi; \
+	    wget -nc -O debian/$(TARBALL) https://dl.google.com/go/$(TARBALL); \
```
DrDaveD, line 79: Doesn't this go against the requirement of not using the network during package build?  I had assumed the golang tarball would be supplied externally.

dist/debian/source/include-binaries
```diff
@@ -0,0 +1,6 @@
+# update this with actual filenames
```
DrDaveD, line 1: I don't think we want to include specific version numbers here directly because they'll quickly get out of date.  I think this file should be generated.  Can it be generated by something in `rules`?  Maybe there could be an option on the `download-dependencies` script to create it, or another script that uses similar logic to extract it from the .spec.

dist/debian/apptainer.conffiles
```diff
@@ -0,0 +1,16 @@
...
+/etc/apptainer/global-pgp-public
+/etc/apptainer/network/00_bridge.conflist
+/etc/apptainer/network/10_ptp.conflist
+/etc/apptainer/network/20_ipvlan.conflist
+/etc/apptainer/network/30_macvlan.conflist
+/etc/apptainer/network/40_fakeroot.conflist
+/etc/apptainer/nvliblist.conf
+/etc/apptainer/remote.yaml
+/etc/apptainer/rocmliblist.conf
+/etc/apptainer/seccomp-profiles/default.json
```
DrDaveD, line 16: I am afraid that people will forget to update this when new files are added.  Can these paths include wildcards, or can this file be generated?

dist/debian/source/include-binaries
Steffen-Gr: See my note above - is there an original place where those version numbers get injected into the system? That one would feed `apptainer.spec.in` and could be used to create `debian/source/include-binaries` - if the src pkg creator decides to have them below `debian/`.
Also some logic would be required to extract the `MINGO_VERSION` from `.mconfig` or whatever is the ultimate source for that one.
For Debian packaging, this could be done by a "release src pkg" script (perhaps in `dist/debian/`?) that might ask the right questions, and gets the right numbers from their canonical source.
OTOH, `.github/workflows/ci.yml` must also be generated somehow... I'm obviously missing the starting point.

dist/debian/control
```diff
@@ -7,14 +7,23 @@ Uploaders:
...
+ libseccomp-dev,
+ uidmap,
+ fakeroot,
+ cryptsetup,
+ tzdata,
+ dh-apparmor,
+ curl,
+ wget,
+ git,
+ libsubid-dev,
```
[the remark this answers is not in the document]
Steffen-Gr: Yes, adding `debian/control.in` and some preprocessing would also do. This would result in individual source packages for individual distros, but the #3012 fix would have a similar effect.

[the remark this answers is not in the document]
Steffen-Gr: Since I'm doing backports mostly, I'm used to tenporarily modify "modern" `debian/*` files to match the situation with older distros. Here, for deb 11 and ubuntu22.04, the dependency line in `debian/control` will be commented out, and in `debian/rules` `SC_OPTIONS = --without-libsubid` will be set as the default (instead of `SC_OPTIONS =`).
If the decision is made to create an individual source package for every distro, that's also fine (I've come up with a `prepare-srcpkg` script to loop over a list of distros). In the end, this decision is a basic one...

dist/debian/source/include-binaries
Steffen-Gr: What's also nice: there's no immediate way to use go1.23.6 in Bullseye. (For Bookworm, there's a 1.22 for the supported architectures.) 1.23.6 needs 1.20.x to build, but we only have 1.18.
What I'm currently looking into is "go version" not failing despite the old version,
```
make[1]: Entering directory '/build/apptainer-1.4.1'
+ go version
go: downloading go1.23.6 (linux/amd64)
go: download go1.23.6: golang.org/toolchain@v0.0.1-go1.23.6.linux-amd64: Get "https://proxy.golang.org/golang.org/toolchain/@v/v0.0.1-go1.23.6.linux-amd64.zip": dial tcp: lookup proxy.golang.org on [::1]:53: read udp [::1]:60862->[::1]:53: read: connection refused
make[1]: *** [debian/rules:76: override_dh_auto_configure] Error 1
make[1]: Leaving directory '/build/apptainer-1.4.1'
make: *** [debian/rules:73: build] Error 2
```
I think I'm going back to 1.22 first. (maybe it'd be best to "just use" the provided tarball?)

dist/debian/rules
DrDaveD: The `.spec.in` is the manually maintained source of truth for where to get those packages.  Everything else derives from there.

A script that is used in common in rpm and debian package building to download if necessary, unpack, and compile go is probably a good idea.

I tested `wget -nc` and you're right that if the file already exists it does not even attempt to go out to the internet to fetch it.  You would need to make sure that wget is a build-time dependency for both rpm & debian.

dist/debian/source/include-binaries
DrDaveD: > I'm obviously missing the starting point.

The starting point is `apptainer.spec.in`, and `mconfig` controls generating everything.  So there could be a script under `dist/debian` that `mconfig` invokes, or it could be included directly in `mconfig` itself as the `apptainer.spec` generation code is.

> there's no immediate way to use go1.23.6 in Bullseye

Oh, good point.  I hadn't noticed because our debian11 github action has internet access and it directly downloads the 1.23.6 binary distribution to save time over compiling from source.  You don't really have that option so I guess it would need to do a double compile.  Messy.

dist/debian/control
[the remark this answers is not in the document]
DrDaveD: `mconfig --without-libsubid` shouldn't need to be explicitly set when the library isn't available because if the library is missing it will automatically skip trying to use it.

dist/debian/source/include-binaries
Steffen-Gr: I have given up - without network access things will fail even for Debian12. With network access, there's no point in providing any tarballs with the source package. I'll revert that - in the end, there will be not much left from this draft (except the conffiles - which the spec file doesn't list in detail `%config(noreplace) %{_sysconfdir}/%{name}/*` )

dist/debian/apptainer.conffiles
Steffen-Gr: There's no simple way except bootstrapping (build once, get list, build using list), I think.

dist/debian/control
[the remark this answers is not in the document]
Steffen-Gr: I didn't fully understand the logic yet - so this means that `--without` is used to **override** what would be found during build (not everyone uses a sandbox, so this makes sense), and one might have some "nice to have as build-dependency" (this concept doesn't exist in Debian, to my limited knowledge - one could fake it by using `libsubid-dev | popcon` which is ugly).
Or modify control, for a fixed list of distros (which is also kind of ugly).

dist/debian/source/include-binaries
Steffen-Gr: How would I drop this commit? (I don't plan to un-draft the PR, I just want to know.)

Re upgrading, I found this instruction somewhere for another target version (would it work from each initial version?):
```
      The go1.21.13 command runs the go command from Go 1.21.13.
      To install, run:
        $ go install golang.org/dl/go1.21.13@latest
        $ go1.21.13 download
      then use the go1.21.13 command as if it were your normal go command.
```
The binary packages https://go.dev/dl/go${v}.linux-${a}.tar.gz do not contain a versioned `bin/go` command AFAICT, so this must get something else... (golang.org/dl/ redirects to go.dev/dl/ AFAIK).
If we know the build architecture (`dpkg-architecture -q ...`) and translate that (! important for ppc64le-> ppc64el and maybe others) we could also download the binary package into some working place below `debian/`, to avoid the bootstrap hassle?

DrDaveD: > How would I drop this commit? (I don't plan to un-draft the PR, I just want to know.)

I assume you're asking about how to undo things in git.  I usually do `git reset HEAD^1` to uncommit my last commit, make my changes, and then do `git push -f` to update a PR branch in progress.  You have multiple commits so you could do `git stash` after each one to save the changes, or you could rename your branch to something else with `git branch -m`, then create a fresh branch, and cherry-pick your old commits one at a time.  You can use `git cherry-pick -n <hash>` to cherry-pick without committing.

> Re upgrading, I found this instruction somewhere for another target version (would it work from each initial version?):
> 
> ```
>       The go1.21.13 command runs the go command from Go 1.21.13.
>       To install, run:
>         $ go install golang.org/dl/go1.21.13@latest
>         $ go1.21.13 download
>       then use the go1.21.13 command as if it were your normal go command.
> ```

That's pretty cool.  I hadn't known that.

> The binary packages [https://go.dev/dl/go${v}.linux-${a}.tar.gz](https://go.dev/dl/go$%7Bv%7D.linux-$%7Ba%7D.tar.gz) do not contain a versioned `bin/go` command AFAICT, so this must get something else... (golang.org/dl/ redirects to go.dev/dl/ AFAIK). If we know the build architecture (`dpkg-architecture -q ...`) and translate that (! important for ppc64le-> ppc64el and maybe others) we could also download the binary package into some working place below `debian/`, to avoid the bootstrap hassle?

I think the problem is that we want one source package to work for multiple architectures so that might not be a good idea.  On the other hand I guess if you're giving up on building without network access then it doesn't much matter.  On the third if you want common code between debian & rpm, the rpm build definitely needs to be able to work without a network.

dist/debian/apptainer.conffiles
DrDaveD: I think they should be available in the source directory after build.  Or at least after the `override_dh_auto_install` and before it starts to create the package they should be in the installation location.

dist/debian/source/include-binaries
Steffen-Gr: I was close to becoming insane, but eventually I found some hints.
Starting at some time (supposedly in the 1.21 series?), running `go` (even `go version`) in the presence of a `go.mod` file would result in an auto-upgrade **to the version specified in that file**.
This finally explains why `go version` wouldn't just tell us our version was too old, as `go` would attempt (and fail without network access!) to silently upgrade itself in the background. Which is perhaps not exactly desirable?
Fortunately, there is an environment variable to control this: ${GOTOOLCHAIN}.
I did some experimenting on my laptop: setting it to `auto` will reproduce the auto-upgrade behaviour (if and only if there's a `go.mod` file, even with a single `go 1.23.6` line!), while `local` will not do the upgrade.
Thus, if full control over the manual installation is desired, we need to `export GOTOOLCHAIN=local` as well, together with the other `GO*` settings. This will be part of my next tests.
I'm still wondering whether a forced but controlled upgrade (as shown above) would be the best choice.

DrDaveD: > > the rpm build definitely needs to be able to work without a network.
>
> I'm in doubt now that this would work at all - after spotting some go mod vendor in scripts/download-dependencies (the debian version of which I'm now shamelessly using). I'm afraid that at least gocryptfs will need external modules.

`download-dependencies` of course needs internet, but it is expected to be done before package building.  It extends the gocryptfs tarball to include the vendor files.  So I don't see the problem.

Steffen-Gr: Hm, something along the lines of `gocryptfs-2.5.1-repacked.tar.gz` would have been nice, indeed. For the effect of `go mod vendor`, just to avoid the network access, no upstream maintainer would accept such a thing. (I remember patches for `squashfuse` being around for a long time - those must have been tricky to maintain... Since these are no longer required, the approach of approaching upstream seems to make sense.) - Over and out.

DrDaveD: Yeah the `squashfuse` case was fixed when I volunteered to be a maintainer.  Apptainer releases a source tarball containing `go mod vendor`.   I already maintain a portion of the other project that I'm working on packaging (`openbao`) so a PR to release a vendored source tarball will get accepted.  There's a good chance that `gocryptfs` would accept a PR too since it is pretty actively maintained.

PR #3077, opened 2025-07-29
DRAFT: two approaches in one: Debian packaging with and without network

### LINKED ISSUE #3069: Instructions to build for Debian/Ubuntu using enhancements?
I'd like to build Apptainer for Debian/Ubuntu, with the enhancements that come e.g. in the SRPM, but cannot find any instructions (except the 23kB `.spec` file).

### Version of Apptainer

1.4.1 or later

Debian {11,12,13} Ubuntu {22.04,24.04} on multiple architectures, including arm64 and riscv64

Not yet.
I'd like to use the full content of the `.src.rpm` file to integrate the enhancements, but `DEBIAN_PACKAGE.md` and `INSTALL.md` don't even mention them. I searched the whole source tree (e.g, for `gocryptfs`) - for older versions, there must have been instructions (otherwise I wouldn't have even looked for/into the SRPM). The `debian/*` tree doesn't seem to be helpful either.

Steffen-Gr:

This is the culmination of my investigations and coding efforts in #3070, and meant to replace it as another "eternal draft" (once again I forgot to sign off my commit..)

- Fixes #3069

Two approaches in one:
(1) `prepare-withnetwork` will prepare the debian/ tree for building with networking **enabled** (`pbuilder` option `--use-network yes`), by providing suitable `control`, `rules` and (empty) `source/include-binaries`.
(2) `prepare-withoutnetwork` will prepare the debian/ tree for building with networking **disabled** (`pbuilder` without `--use-network`), by providing suitable `control`, `rules` and `source/include-binaries`. Tarballs for go and dependencies must be provided!

Both approaches have been tested on amd64 Ubuntu 24.04 ("Noble").
Make sure to provide a `go` version >= 1.20.6!

Again, comment, ignore, steal or reject :)


DrDaveD (changes requested):
Thanks for figuring out what it takes to build with and without a network.

However, I don't like duplicating these files because it makes maintenance more difficult; every change has to go into two places.   I would much rather use `.in` files and a script that can read those in and write out the processed output.

And in this case it does not seem even `.in` file is necessary.  I did a `diff` between the `withnetwork` and `withoutnetwork` files and they are trivial.  It doesn't hurt to include `curl` and `wget` even when there's no network in `control`, and the `rules` could check to see if a target file exists and if not do the wget or call download-dependencies.

But maybe you will need a `control.in` after all because the debian11, debian12, and ubuntu22 CI checks still failed.  Maybe we don't care about debian11 so close to its end of life, but the other two need to succeed.

What does the `.ex` extension on `apptainer.conffiles` indicate?

You don't need to make a new PR to update.  Just do `git reset HEAD^1` to back up, and after you do `git commit -s` do `git push -f` to overwrite what's there.

DrDaveD:
I'm closing this as stale.  Reopen if you want to make more progress on it.

PR #3078, opened 2025-07-30
Update to minimum go 1.24.1

DrDaveD:
golang 1.24.4 is now available on el8 & el9, so update to that minimum version.


DrDaveD:
But ubuntu 22.04 only has 1.24.1 so use that as a minimum instead.

PR #3499, opened 2026-05-17
Add short-hand mksquashfs compression build flags

### LINKED ISSUE #3498: Experiment with shorthand flags for squashfs compression
When you use the `gzip` command, it has some simple flags like `--fast` and `--best`.

It would be nice with something similar for apptainer build, even beyond legacy gzip...

Currently there is a `--mksquashfs-args` generic option, that can do anything/nothing.

https://apptainer.org/docs/user/main/build_a_container.html#alternative-compressors

The assumption is that both compress **and** decompress can be up to 5x faster than now*.

\* by switching the default compressor from gzip to zstd, before any levels or comp flags.

Follow-up from:

* #2785

-----

According to my own experiments, lz4 is the fast and zstd is the best compression.

But note also that lz4 has the _worst_ size, and for instance xz has the _worst_ speed:

Image

_the type of image also impacts the result, so need to test with several more of them_

Switching the default from `gzip -9` to `zstd -9` has a good impact on both, though.

The compression/decompression speed is faster, but the image size is still smaller...

-----

Compression | `mksquashfs-args`
---  | ---
None | `-no-compression # -noI -noD -noF -noX`
_Legacy options:_  | 
Fastest |`-comp lzo`
Fast |`-comp gzip -Xcompression-level 1`
Default |`-comp gzip # -Xcompression-level 9`
Best | `-comp xz`
_Modern options:_ | 
Fastest | `-comp lz4`
Fast | `-comp zstd -Xcompression-level 1`
Default | `-comp zstd -Xcompression-level 9`
Best |`-comp zstd # -Xcompression-level 15`

Higher levels of zstd are exponentially more expensive to compress, same as with xz.

They also have a performance penalty for decompress, high with xz - less with zstd.

afbjorklund:

fixing or enhancing.

 - Fixes #3498

### DIFF
--- a/cmd/internal/cli/build.go
+++ b/cmd/internal/cli/build.go
@@ -38,6 +38,10 @@ var buildArgs struct {
 	keyServerURL        string
 	webURL              string
 	mksquashfsArgs      string
+	none                bool
+	fast                bool
+	best                bool
+	legacy              bool
 	encrypt             bool
 	fakeroot            bool
 	fakefakeroot        bool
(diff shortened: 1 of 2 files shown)


DrDaveD:
Since compression is not the main function of this application, I don't think the flag names should be so compact.  I think it should be something like `--compress-level {none|fast|best|legacy}`.

I'm also still concerned about backward compatibility if we change the default.

DrDaveD:
I suggest not changing the default just yet but making a separate issue for that, in case 1.6.0 is released on a comparatively short time frame.

afbjorklund:
I will open a new PR later, when decided on the flag names and how they would interact.

Issue #3486, opened 2026-05-12
Corrupt SIF output with a combination of proot and xz


afbjorklund:
### Version of Apptainer

What version of Apptainer (or Singularity) are you using?

`apptainer version 1.5.0`

### Expected behavior

What did you expect to see when you do...?

I was comparing different compression methods, and tested legacy "xz":

```
apptainer --debug build --mksquashfs-args="-comp xz" python.sif docker://python
```

`310M	python.sif`

### Actual behavior

What actually happened? Why was it incorrect?

The compression is aborted after 41%, no error. The output SIF is corrupt:

```
VERBOSE [U=1000,P=598688]  create()                      Executing /usr/local/libexec/apptainer/bin/proot -S / /usr/local/libexec/apptainer/bin/mksquashfs /tmp/build-temp-372428726/rootfs /tmp/bundle-temp-887462477/squashfs-3910298712 -noappend -comp xz
```

`52K	python.sif`

### Steps to reproduce this behavior

How can others reproduce this issue/problem?

Setting `APPTAINER_IGNORE_PROOT` seems to fix the issue.

Indicating that the issue is with `proot` (and `xz`)?

### What OS/distro are you running

Ubuntu 22.04.5 LTS

liblzma 5.2.5-2ubuntu1

### How did you install Apptainer

Write here how you installed Apptainer (or Singularity). Eg. RPM, source.

source

`/usr/local/libexec/apptainer/bin/proot`

`/usr/local/libexec/apptainer/bin/mksquashfs`

DrDaveD:
I tried to reproduce it on a fresh Ubuntu 22.04 VM and apptainer [installed from the Ubuntu PPA](https://apptainer.org/docs/admin/latest/installation.html#install-ubuntu-packages).  It did not have a problem.  Can you try doing that on your machine?

afbjorklund:
It was the same behaviour with the PPA binaries, it went to 47% instead of 41% but I don't know if that has any significance...

DrDaveD:
There seems to be something different about our machines then.  It's the same Ubuntu 22.04.5 LTS and liblzma5 5.2.5-2ubuntu1.  Maybe you can have it run under a debugger or something to find out why it is dying.

afbjorklund:
Since it is not a recommended compression and only happens rarely, it is not urgent. But still strange so will keep in backlog.

afbjorklund:
[...]

PR #3516, opened 2026-05-22
Update to PRoot-5.4.0-rootless.3

DrDaveD:
This fixes the SIF file corruption reported in #3486.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@ Changes since 1.5.x
 
 ## v1.5.x changes
 
-Changes since v1.5.0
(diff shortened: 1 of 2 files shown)


afbjorklund (approved):
Now it fails "properly", i.e. returns the exit code from the underlying mksquashfs

`FATAL:   While performing build: while creating squashfs: /usr/local/libexec/apptainer/bin/mksquashfs command failed: exit status 139: `

Now, wonder that error means (the theory is that it is related to memory issues)

i.e. it is SIGSEGV, but what part of the xz wrapper that is actually segfaulting

PR #3518, opened 2026-05-26
Skip downloading PRoot on unsupported architectures 

### LINKED ISSUE #3488: riscv64 Debian trixie build of 1.5.0 fails, while 1.4.5 had passed
### Version of Apptainer

1.5.0

### Expected behavior

Since 1.4.5 built for Debian Trixie on riscv64, I expected the same to happen with 1.5.0

### Actual behavior

```
make[2]: Entering directory '/build/apptainer-1.5.0/PRoot-5.4.0-rootless.2/src'
.//arch.h:55:10: error: #error "Unsupported architecture"
```

The Changelog suggests that since 1.4.0 riscv64 support would be "complete", but something must have changed since 1.4.5

Attempt to build e.g. in a riscv64 VM (I am using UTM)

Debian 13 Trixie

not yet...

DrDaveD:
- Fixes #3488

### DIFF
--- a/scripts/download-dependencies
+++ b/scripts/download-dependencies
@@ -40,6 +40,12 @@ sed -n -e '/^# URL:/{s/^# //;s/%%/%/g;p}' -e 's/^Source[1-9][0-9]*: //p' -e 's/^
 	# take the file name from the base of the URL
 	LINE="${LINE/*\//}"
     fi
+    if [[ "$LINE" = *PRoot* ]]; then
+        # skip architectures known to not be supported by PRoot
+        case $(arch) in
+            ppc*|s390*|riscv*) continue;;
+        esac
+    fi
     if [ -n "$GITHUB_TOKEN" ] && [[ "$URL" = *github.com/*.patch ]]; then
 	# use github API instead because the public URL downloads 
 	# sometimes hit server busy (429) when run in github actions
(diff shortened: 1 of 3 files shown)


scripts/download-dependencies
```diff
@@ -40,6 +40,12 @@ sed -n -e '/^# URL:/{s/^# //;s/%%/%/g;p}' -e 's/^Source[1-9][0-9]*: //p' -e 's/^
 	# take the file name from the base of the URL
 	LINE="${LINE/*\//}"
     fi
+    if [[ "$LINE" = *PRoot* ]]; then
+        # skip architectures known to not be supported by PRoot
+        case $(arch) in
+            ppc*|s390*|riscv*) continue;;
```
afbjorklund, line 46: You might want to use the explicitly supported architectures instead?

So that it doesn't fail when introducing new architectures like loong64

DrDaveD: I thought about that but decided I'd rather consider them individually as they came along so I don't accidentally exclude any.  It could be done either way though, you're right about that.

afbjorklund (approved):

scripts/download-dependencies
AEI-SteffenG: Maybe allowing the PRoot build to fail (make ... || true) and the result not to be found would be close to one-fits-all.

DrDaveD: But then we wouldn't notice if it fails to build when it should build.

PR #3522, opened 2026-05-28
Avoid overwriting /etc/passwd in case of root user

### LINKED ISSUE #3502: User's name instead of `root` when running with `--fakeroot`
### Expected behavior
```
$ /usr/bin/apptainer exec --fakeroot --cleanenv --containall docker://ubuntu id    
uid=0(root) gid=0(root) groups=0(root),65534(nogroup),65534(nogroup)

$ /usr/bin/apptainer --version
apptainer version 1.4.5
```

### Actual behavior

With apptainer 1.5.0 it started showing my user name rather than showing `root`

```
$ /usr/bin/apptainer exec --fakeroot --cleanenv --containall docker://ubuntu id 
uid=0(mahnerak) gid=0(root) groups=0(root),65534(nogroup),65534(nogroup)

$ /usr/bin/apptainer --version
apptainer version 1.5.0
```

```sh
$ cat /etc/os-release

PRETTY_NAME="Ubuntu 22.04.5 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.5 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

mahnerak:

Especially in case of fakeroot, if the uid inside the apptainer is 0 (root) we want to avoid mapping it back to the original username.

 - Fixes #3502 

### DIFF
--- a/internal/pkg/util/fs/files/passwd.go
+++ b/internal/pkg/util/fs/files/passwd.go
@@ -87,6 +87,10 @@ func Passwd(path string, home string, uid int, customLookup UserGroupLookup) (co
 			if err != nil {
 				return nil, err
 			}
+			// If user already exists in container, change their username, gecos
+			// and home dir to those of the original user. Except for uid 0,
+			// where the original user is the unprivileged host user (e.g.
+			// fakeroot) and not root, so keep the container's existing values.
 			name := pwInfo.Name
 			gecos := pwInfo.Gecos
 			if uid == 0 {
--- a/internal/pkg/util/fs/files/passwd.go
+++ b/internal/pkg/util/fs/files/passwd.go
@@ -87,8 +87,13 @@ func Passwd(path string, home string, uid int, customLookup UserGroupLookup) (co
 			if err != nil {
 				return nil, err
 			}
-			// If user already exists in container, change their username and home dir to those of the original user
-			lines[i] = makePasswdLine(pwInfo.Name, uid32, gid32, pwInfo.Gecos, homeDir, entry.Shell())
+			name := pwInfo.Name
+			gecos := pwInfo.Gecos
+			if uid == 0 {
+				name = entry.Username()
+				gecos = entry.Info()
+			}
+			lines[i] = makePasswdLine(name, uid32, gid32, gecos, homeDir, entry.Shell())
 			break
 		}
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 Changes since 1.5.x
 
+- Fix fakeroot overwriting root's username in `/etc/passwd` with the host
+  user's name, a regression introduced in v1.5.0.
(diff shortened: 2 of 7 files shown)


DrDaveD (changes requested):
I'm hesitant to make such a significant change because of potential unwanted side effects.   I see at least one unexpected change in behavior, in that since it skips using an internally-generated passwd file, you cannot change the home directory in it with the `--home` option when running as root as it does now.  It does still change `$HOME` and the value of `~`, but it does not change the value of `~root`.  One could argue that no-one is likely to be depending on that, plus it already doesn't happen if the image or overlay is writable.  But I would prefer to make the minimal change by just undo-ing the change in passwd.go done in #3386 when uid is 0.

mahnerak:
> But I would prefer to make the minimal change by just undo-ing the change in passwd.go done in #3386 when uid is 0.

My initial version just added an `if` statement there, but I noticed the check for `(uid == 0 && is_writable_...)`, which seemed counterintuitive. As you noted, writable overlays already exhibit different behavior (which I believe should not be the case).

I am updating this PR to only add the `uid == 0` check to the #3386 logic to ship the bug fix as soon as possible. However, I still see value in discussing how we might simplify the flow as proposed in my original version.

peci1:
I've locally tested this PR and I confirm it solves #3502.

internal/pkg/util/fs/files/passwd.go
```diff
@@ -87,8 +87,13 @@ func Passwd(path string, home string, uid int, customLookup UserGroupLookup) (co
 			if err != nil {
 				return nil, err
 			}
-			// If user already exists in container, change their username and home dir to those of the original user
-			lines[i] = makePasswdLine(pwInfo.Name, uid32, gid32, pwInfo.Gecos, homeDir, entry.Shell())
+			name := pwInfo.Name
```
peci1, line 94: Maybe add an explaining comment here?

peci1 (approved):

internal/pkg/util/fs/files/passwd.go
DrDaveD: Yes, I would like to see an explaining comment talking about the two cases.  Otherwise everything looks good.

DrDaveD (changes requested):

DrDaveD:
Please address the check_source failure.  You can put `//nolint:gosec` as a comment on the end of that line.

internal/pkg/util/fs/files/passwd.go
mahnerak: Sounds good!

DrDaveD:
Please rebase on the current main branch

DrDaveD (approved):

PR #3535, opened 2026-06-01
Work around segmentation fault in mksquashfs under proot

### LINKED ISSUE #3486: Corrupt SIF output with a combination of proot and xz

DrDaveD:
Work around segmentation fault sometimes seen in mksquashfs under proot by setting some environment variables to change the behavior of malloc.

- Fixes #3486

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ Changes since 1.5.x
 
 ## v1.5.x changes
 
+- Work around segmentation fault sometimes seen while mksquashfs under
+  proot is creating a SIF file.
 - Update bundled PRoot to version 5.4.0-rootless.3 in order to fix a
   problem where SIF files could be corrupted when mksquashfs died with
   a signal.  The proot command was not passing back an error exit code.
--- a/internal/pkg/image/packer/squashfs.go
+++ b/internal/pkg/image/packer/squashfs.go
@@ -73,6 +73,7 @@ func (s Squashfs) create(files []string, dest string, opts []string) error {
 	}
 
 	prog := s.MksquashfsPath
+	extraEnv := []string{}
 	if namespaces.IsUnprivileged() {
 		// building as unprivileged user, make the files appear as root
 		ignoreProot := os.Getenv("APPTAINER_IGNORE_PROOT")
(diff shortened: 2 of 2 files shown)


afbjorklund:
Somewhat unobvious, but we increase the number of threads "possible" and disable mmap.

```
       M_ARENA_MAX
              If this parameter has a nonzero value, it defines a hard limit on the maximum number of arenas that can be created.  An arena represents  a  pool
              of memory that can be used by malloc(3) (and similar) calls to service allocation requests.  Arenas are thread safe and therefore may have multi‐
              ple concurrent memory requests.  The trade-off is between the number of threads and the number of arenas.  The more arenas you  have,  the  lower
              the per-thread contention, but the higher the memory usage.

              The default value of this parameter is 0, meaning that the limit on the number of arenas is determined according to the setting of M_ARENA_TEST.


       M_ARENA_TEST
              This parameter specifies a value, in number of arenas created, at which point the system configuration will be examined to determine a hard limit

              The default value for the M_ARENA_TEST parameter is 2 on systems where sizeof(long) is 4; otherwise the default value is 8.
```

afbjorklund (approved):

PR #3543, opened 2026-06-03
Enable extglob for %files section

sieben-gea:

I noticed that the path expansion in the %files section was done by executing a script with mvdan.cc/sh, which struck me as unnecessarily unsafe. Instead, we can use the mvdan.cc/sh/v3/expand package that is used by the package internally as well, and skip the rest of the shell, reducing the possibility of unexpected command injections via the %files section.

Doing that, I also enabled extglob for the %files section, allowing to make it more concise. Unfortunately negative matching is still not possible, see https://github.com/mvdan/sh/issues/1158#issuecomment-3316224634.

	- I did not run test which required me to run the test suite with sudo privileges, because I did not see how this was necessary for the changes I made and it struck me as unsafe. Maybe for local testing you should consider providing an option to skip privileged tests for local testing.

### DIFF
--- a/internal/pkg/build/files/files.go
+++ b/internal/pkg/build/files/files.go
@@ -10,53 +10,36 @@
 package files
 
 import (
-	"bytes"
-	"context"
-	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
 	securejoin "github.com/cyphar/filepath-securejoin"
-	"mvdan.cc/sh/v3/interp"
+	"mvdan.cc/sh/v3/expand"
 	"mvdan.cc/sh/v3/syntax"
 )
 
-const filenameExpansionScript = `
-for n in %[1]s ; do
-	echo -n "$n\0"
-done
-`
-
 func expandPath(path string) ([]string, error) {
-	var output, stderr bytes.Buffer
-
 	path = strings.ReplaceAll(path, " ", "\\ ")
-	cmdline := fmt.Sprintf(filenameExpansionScript, path)
-	parser, err := syntax.NewParser().Parse(strings.NewReader(cmdline), "")
+	parsedPath, err := syntax.NewParser().Document(strings.NewReader(path))
 	if err != nil {
 		return nil, err
 	}
 
-	runner, err := interp.New(
-		interp.StdIO(nil, &output, &stderr),
-	)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


DrDaveD (changes requested):
Thanks for the contribution!

Github is not running checks, I think because you're a first time contributor, but it's also not giving me the option to approve running them.  Please rebase on the current main to fix the CHANGELOG conflict, maybe that will do it.

sieben-gea:
I did the rebase, but it seems you need to enable the workflows manually, which is a very sensible security default! [According to the GitHub docs](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/approve-runs-from-forks), you need to do that by clicking on the merge status button in the upper right corner.

DrDaveD:
This time the button for me to approve the workflow was there.

internal/pkg/build/files/files.go
```diff
@@ -10,53 +10,36 @@
...
 	}
 
-	// parse expanded output and ignore empty strings from consecutive null bytes
-	paths := make([]string, 0, len(strings.Split(output.String(), "\\0")))
+	expandedPaths := expand.FieldsSeq(&expandConfig, parsedPath)
 
-	for _, s := range strings.Split(output.String(), "\\0") {
-		if s == "" {
-			continue
+	var paths []string
```
DrDaveD: The linter wants you to pre-allocate this to the size of expandedPaths.  You could use
```
    paths := make([]string, 0, len(expandedPaths))
```

DrDaveD (changes requested):

internal/pkg/build/files/files.go
DrDaveD: @sieben-gea This is waiting on you.  Also needs another rebase at this point.

sieben-gea: Oh, I am sorry I didn’t have the time to work on it right away!
Since `expandedPaths` is an iterator `len(expandedPaths)` is not valid, as the length of the iterator is only known after the loop. I can loop through it twice to please the linter, but that seems a bit strange. I’ll just pre-allocate to a constant array size, but for longer arrays, `append` will create a new larger array anyway.

DrDaveD (approved):

PR #3546, opened 2026-06-03
Avoid overwriting /etc/passwd in case of root user (release-1.5)

mahnerak:
Cherry-picks #3522 to release-1.5

### DIFF
--- a/internal/pkg/util/fs/files/passwd_test.go
+++ b/internal/pkg/util/fs/files/passwd_test.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/apptainer/apptainer/internal/pkg/test"
 	"github.com/apptainer/apptainer/internal/pkg/util/user"
+	"github.com/ccoveille/go-safecast"
 )
 
 func TestPasswd(t *testing.T) {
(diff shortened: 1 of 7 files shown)


internal/pkg/util/fs/files/passwd_test.go
```diff
@@ -12,12 +12,13 @@ package files
...
 	"fmt"
 	"os"
-	"os/user"
 	"path/filepath"
 	"strings"
 	"testing"
 
 	"github.com/apptainer/apptainer/internal/pkg/test"
+	"github.com/apptainer/apptainer/internal/pkg/util/user"
+	"github.com/ccoveille/go-safecast/v2"
```
DrDaveD: The move to go-safecast/v2 hasn't been backported to release-1.5.  Please remove the `v2`.

DrDaveD (changes requested):

internal/pkg/util/fs/files/passwd_test.go
mahnerak: Ah, that explains why I did not have linter warnings on my side when pushing #3522

DrDaveD (approved):

PR #3573, opened 2026-06-12
Switch to get opensuse containers from registry.opensuse.org and update opensuse-leap to 16.0

DrDaveD:
The latest opensuse/leap container on dockerhub is 15.6, and its download repository has disappeared from download.opensuse.org/repositories/filesystem.  Switch opensuse containers to instead use registry.opensuse.org which has newer versions of opensuse.

(On the other hand, the place that `tools/install-unprivileged.sh -o` downloads from under download.opensuse.org/distribution/leap still has a 15.6 and does not have a 15.7)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,14 +227,14 @@ jobs:
 
       - name: Install and test unprivileged for openSUSE leap
         env:
-          OS_TYPE: opensuse/leap
+          OS_TYPE: registry.opensuse.org/opensuse/leap
           OS_VERSION: latest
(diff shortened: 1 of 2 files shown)


DrDaveD:
OpenSUSE leap 15.6 reached end of life in April 2026 and there was no 15.7.  I will update install-unprivileged.sh accordingly.

DrDaveD:
Also up the number of retries from 3 to 5 in install-unprivileged.sh and reduce the bytes per second needed to meet the minimum bandwidth requirement

cclerget (approved):

PR #3579, opened 2026-06-18
Fix build issues

### LINKED ISSUE #3347: Re-enable FIPS support in EL builds
For [bugzilla 2437258](https://bugzilla.redhat.com/show_bug.cgi?id=2437258) the EPEL/Fedora packaging was updated to enable FIPS support.  Import that change to do `export GODEBUG=fips140=on` near the beginning of the %build step.

DrDaveD:
Fix some issues that happened with the 1.5.1 apptainer build.  Stop building PRoot on fedora 45 x86_64, and improve the fedora build instructions.  Set `GOEXPERIMENT=strictfipsruntime` when the go version includes "Red Hat".

- Fixes #3347

### DIFF
--- a/scripts/download-dependencies
+++ b/scripts/download-dependencies
@@ -44,6 +44,16 @@ sed -n -e '/^# URL:/{s/^# //;s/%%/%/g;p}' -e 's/^Source[1-9][0-9]*: //p' -e 's/^
         # skip architectures known to not be supported by PRoot
         case $(arch) in
             ppc*|s390*|riscv*) continue;;
+            x86*) if [ -f /etc/os-release ]; then
+                    . /etc/os-release
+                    if [ "$ID" = fedora ] && [ "$VERSION_ID" -ge 45 ]; then
+                        # On fedora45 x86_64 building PRoot dies with
+                        # "relocation truncated to fit: R_X86_64_PC32 against `.rodata'"
+                        # See https://github.com/proot-me/proot/issues/414
+                        continue
+                    fi
+                fi
+                ;;
         esac
     fi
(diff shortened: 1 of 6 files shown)


scripts/download-dependencies
```diff
@@ -44,6 +44,15 @@ sed -n -e '/^# URL:/{s/^# //;s/%%/%/g;p}' -e 's/^Source[1-9][0-9]*: //p' -e 's/^
         # skip architectures known to not be supported by PRoot
         case $(arch) in
             ppc*|s390*|riscv*) continue;;
+            x86*) if [ -f /etc/os-release ]; then
+                    . /etc/os-release
+                    if [ "$ID" = fedora ] && [ "$VERSION_ID" -ge 45 ]; then
+                        # On fedora45 x86_64 building PRoot dies with
+                        # "relocation truncated to fit: R_X86_64_PC32 against `.rodata'"
```
afbjorklund, line 51: is there an upstream issue about this build issue? Seems like a regression

afbjorklund (approved):
LGTM, would be good to link to the proot issue so that it can be fixed (eventually)

scripts/download-dependencies
DrDaveD: There isn't.  I maintain the immediate PRoot upstream so it doesn't help for there to be an issue there, and its upstream unfortunately seems to not be maintained, with no commits in 3 years ... ah, but there were a few commits last week.  I guess I will make an issue.

DrDaveD: proot-me/proot#414

PR #3591, opened 2026-06-22
Update AGENTS.md and security policy

DrDaveD:
The first commit is based on one from
- sylabs/singularity#4067

and then the second adds policy regarding the Apptainer threat model and how container images are trusted by the user.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,97 @@
+# AGENTS.md
+
+This file provides instructions for AI agents working in the apptainer
+repository.
+
+## Overview
+
+Apptainer is a container runtime, focused on HPC and scientific computing
+use-cases.
+
+## Quick-Start
+
+Install prerequisites listed in [INSTALL.md](INSTALL.md)
+
+```sh
+./mconfig -v                        # Configure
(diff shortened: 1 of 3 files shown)


afbjorklund:
I'm a little unsure about best way to review this...

It seems the best would be to feed it to an agent, and see what it makes of it - how it affects the session

afbjorklund:
Easier to compare with Singularity CE...

It seems like only the Code Structure is "missing", and then there is the difference in name and CI and OCI... OK.

```diff
@@ -38,3 +38,3 @@
 - C code: the `cmd/starter/` runtime starter is C. Do not modify C code
-  without understanding the security implications — the starter runs as
+  without understanding the security implications -- the starter can run as
   a setuid binary.
```

1. The good ole emdash (AI™)
2. Saying that setuid is optional

Reason for [code structure](https://github.com/sylabs/singularity/blob/main/AGENTS.md#code-structure) was:

> A summary of the repository layout is useful, as otherwise agents will do a huge amount of grepping or similar to find things.

afbjorklund (approved):

PR #3609, opened 2026-07-01
build: use an os.Root for Mkdir / Writefile operations, from sylabs 4178

### LINKED ISSUE #3594: Import Sylabs 4178
I heard about this SingularityCE PR that I would like ported to Apptainer:
- sylabs/singularity#4178

luissimas:

This pulls in the sylab PR https://github.com/sylabs/singularity/pull/4178 and fixes some issues introduced by it.

> In builds, we frequently create directories and write files into a rootfs dir. We can use os.Root to better restrict these operations to the rootfs directory.
> 
> This code was developed out of exploratory work using Claude Code, that has been manually reviewed and edited. The Co-authored-by trailer is added on this basis.
> 
> Co-authored-by: Claude Code

 - Fixes #3594

### DIFF
--- a/internal/pkg/build/sources/base_environment.go
+++ b/internal/pkg/build/sources/base_environment.go
@@ -15,7 +15,7 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/apptainer/apptainer/internal/pkg/util/fs"
+	"github.com/apptainer/apptainer/pkg/build/types"
 	"github.com/apptainer/apptainer/pkg/sylog"
 )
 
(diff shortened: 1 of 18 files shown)


DrDaveD (approved):

internal/pkg/build/sources/base_environment.go
```diff
@@ -115,32 +115,32 @@ func makeDirs(b *types.Bundle) error {
 
 func makeSymlinks(b *types.Bundle) error {
 	if _, err := b.Rootfs.Stat("singularity"); err != nil {
-		if err = b.Rootfs.Symlink("/.singularity.d/runscript", "singularity"); err != nil {
```
dtrudg: Edit - deleted... misread some git history across different branches / projects here.

dtrudg:
Note that... 

>Ensures paths used on os.Root methods are relative and not absolute,
which would cause a runtime error when using os.Root methods.

... is not technically true for some of the changes you have made. The `oldname` for `Root.Symlink` doesn't actually have to be within the root: https://pkg.go.dev/os#Root.Symlink and can be absolute.

I've deleted a previous comment, which had an incorrect interpretation from looking at a confused history. Sorry for the noise.

luissimas:
> Note that...
> 
>  >  Ensures paths used on os.Root methods are relative and not absolute,
>  > which would cause a runtime error when using os.Root methods.
>
> ... is not technically true for some of the changes you have made. The oldname for Root.Symlink doesn't actually have to be within the root: https://pkg.go.dev/os#Root.Symlink and can be absolute.

You're right, `Root.Symlink` doesn't require `oldname` to be within the root.

But note that the call to `Root.Symlink` only happens after checking the error returned by a [call](https://github.com/luissimas/apptainer/blob/a29f81caa3daa01f6bfbf0fb2596b4548e45ba80/internal/pkg/build/sources/base_environment.go#L117) to `Root.Stat` to check for the existence of the symlink. The original check makes sense to avoid creating the symlink twice, and assumes that the call to `os.Stat` will return an error if the symlink is not created. The problem is that `Root.Stat` does follow symlinks and [checks that symlinks do not escape the root](https://go.dev/blog/osroot#introducing-osroot).

Given that, the call to `Root.Stat` on these symlinks would always fail, triggering the re-creation of the symlink. This caused the following error observed in a [previous CI run for this PR](https://github.com/apptainer/apptainer/actions/runs/28515375111/job/84525690003):

```
INFO:    Starting build...
INFO:    Verifying bootstrap image image.sif
INFO:    Extracting local image...
FATAL:   While performing build: packer failed to pack: while inserting base environment: build: failed to make environment symlinks: symlinkat /.singularity.d/runscript singularity: file exists
```

I believe the fix in this case is to either make the symlink relative again (which AFAIK does not impact the end result of the .sif file), or check for the existence of the symlink file using `Root.Lstat` (which does not follow symlinks). I haven't tested the second option, though.

I haven't tried reproducing this error we saw in CI with singularity 4.5, but I did reproduce another issue related to `os.Root`.

```
$ cat alma.def
Bootstrap: yum
OSVersion: 9
MirrorURL: http://repo.almalinux.org/almalinux/%{OSVERSION}/BaseOS/x86_64/os
Include: dnf
$ sudo ./singularity build alma.sif alma.def
INFO:    Starting build...
WARNING: Your host system is using the sqlite RPM database backend.
WARNING: Bootstrapping of older distributions that use the bdb backend will fail.
FATAL:   While performing build: conveyor failed to get: while generating yum config: while creating /tmp/build-temp-365577488/rootfs/etc/bootstrap-yum.conf: openat /etc/bootstrap-yum.conf: path escapes from parent
```

 This one is fixed by https://github.com/apptainer/apptainer/commit/a29f81caa3daa01f6bfbf0fb2596b4548e45ba80. I was going to open an issue to let you know but you were faster than me 😅.

dtrudg:
> I believe the fix in this case is to either make the symlink relative again (which AFAIK does not impact the end result of the .sif file), or check for the existence of the symlink file using `Root.Lstat` (which does not follow symlinks). I haven't tested the second option, though.

Yeah, a stronger `Lstat` check probably makes sense there. If it's a link, we should really be checking the link has the correct target... and if not, recreate it correctly. If it's a file, then it'd be an extremely old container, but probably valid (I'd have to go spelunking in old version code to verify).

> I haven't tried reproducing this error we saw in CI with singularity 4.5, but I did reproduce another issue related to `os.Root`.
> 
> ```
> $ cat alma.def
> Bootstrap: yum
> OSVersion: 9
> MirrorURL: http://repo.almalinux.org/almalinux/%{OSVERSION}/BaseOS/x86_64/os
> Include: dnf
> $ sudo ./singularity build alma.sif alma.def
> INFO:    Starting build...
> WARNING: Your host system is using the sqlite RPM database backend.
> WARNING: Bootstrapping of older distributions that use the bdb backend will fail.
> FATAL:   While performing build: conveyor failed to get: while generating yum config: while creating /tmp/build-temp-365577488/rootfs/etc/bootstrap-yum.conf: openat /etc/bootstrap-yum.conf: path escapes from parent
> ```
> 
> This one is fixed by [a29f81c](https://github.com/apptainer/apptainer/commit/a29f81caa3daa01f6bfbf0fb2596b4548e45ba80). I was going to open an issue to let you know but you were faster than me 😅.

Thanks for letting me know.

luissimas:
> Yeah, a stronger `Lstat` check probably makes sense there. If it's a link, we should really be checking the link has the correct target... and if not, recreate it correctly. If it's a file, then it'd be an extremely old container, but probably valid (I'd have to go spelunking in old version code to verify).

Ah, I didn't knew about the file case at all, thanks! IIUC this should now be handled by the changes in https://github.com/apptainer/apptainer/pull/3609/commits/9cfadab251abbd31738ea01c6e549e0712eb5a89, but if I missed something please let me know.

dtrudg:
Hi @luissimas - just a note that I picked the fix into SingularityCE (https://github.com/sylabs/singularity/pull/4237)

Would you like to be added to the CONTRIBUTORS.md file over there?

luissimas:
Hi @dtrudg.

> Would you like to be added to the CONTRIBUTORS.md file over there?

Yes, that'd be nice. Thanks!

PR #3658, opened 2026-07-25
Default to /var/tmp if /tmp is on tmpfs

### LINKED ISSUE #3657: Auto-detect tmp-on-tmpfs for APPTAINER_TMPDIR
Now that more and more distributions are using the systemd tmpfs for /tmp,
maybe this should be automatically detected to avoid APPTAINER_TMPDIR

Currently the packaging needs an override `APPTAINER_TMPDIR=/var/tmp`,
due to /tmp being mounted with "nodev" (and possible too small, being RAM)

* https://github.com/lima-vm/lima/pull/1085 (fedora)

* https://github.com/lima-vm/lima/pull/5331 (ubuntu)

----

Previously Fedora 18+

https://fedoraproject.org/wiki/Features/tmp-on-tmpfs

Now also Ubuntu 24.10+

https://documentation.ubuntu.com/release-notes/26.04/summary-for-lts-users/#systemd-259

BEFORE (24.04)

```
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda1        96G  1.9G   94G   2% /
```

AFTER (26.04)

```
TARGET SOURCE FSTYPE OPTIONS
/tmp   tmpfs  tmpfs  rw,nosuid,nodev,nr_inodes=1048576,inode64,usrquota

Filesystem      Size  Used Avail Use% Mounted on
tmpfs           2.0G     0  2.0G   0% /tmp
```

afbjorklund:
The tmpfs mount on /tmp by systemd typically has "nodev" and is too small for use by Apptainer. Use /var/tmp instead.

fixing or enhancing.

 - Fixes #3657

### DIFF
--- a/cmd/internal/cli/apptainer.go
+++ b/cmd/internal/cli/apptainer.go
@@ -251,7 +251,7 @@ var commonOldNoHTTPSFlag = cmdline.Flag{
 var commonTmpDirFlag = cmdline.Flag{
 	ID:           "commonTmpDirFlag",
 	Value:        &tmpDir,
-	DefaultValue: os.TempDir(),
+	DefaultValue: tempDir(),
 	Hidden:       true,
 	Name:         "tmpdir",
 	Usage:        "specify a temporary directory to use for build",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Changes since 1.5.x
 - Add a `sync writable extfs` directive to apptainer.conf. When enabled,
   writable extfs image mounts use the `sync` mount option.
 - Added support for NVIDIA Tegra to `nvliblist.conf`
+- The `APPTAINER_TMPDIR` now defaults to /var/tmp, if /tmp is on tmpfs.
 
 ## v1.5.x changes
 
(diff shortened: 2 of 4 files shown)


cmd/internal/cli/apptainer.go
```diff
@@ -296,6 +296,17 @@ func getCurrentUser() *user.User {
 	return usr
 }
 
+func tempDir() string {
+	tmp := os.TempDir()
```
DrDaveD: I think that if somebody explicitly sets `$TMPDIR` to `/tmp` it should continue to use that even if it is of type tmpfs.  `os.TempDir()` returns `$TMPDIR` if it is set, so I think this means that the thing to do is to not call `os.TempDir()` but implement it yourself and only check for tmpfs if it's about to default to `/tmp`.

DrDaveD (changes requested):

DrDaveD:
It looks like tmpDir can also be set in `cmd/internal/action_flags.go` although it's not clear to me why it needs to be in there if it is already at the apptainer top level.

DrDaveD (approved):

PR #3667, opened 2026-07-31
disable record length test in mount string parsing

rroohhh:
This allows for multiple entries in `APPTAINER_MOUNT` with different number of options. See the added test for an example. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ Changes since 1.5.x
 
 - Fix the help text for the `--mount` option not wrapping at 80 columns,
   by using the shorter `src` and `dst` aliases in the example.
+- Fixed a bug that prevented multiple mount entries in the `APPTAINER_MOUNT`
+  env var from having different numbers of options.
 
 ## v1.5.3 - \[2026-07-21\]
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -119,6 +119,7 @@
 - Richard Neuboeck <hawk@tbi.univie.ac.at>
 - Rémy Dernat <remy.dernat@umontpellier.fr>
 - Robert Clarke <robert.clarke@bristol.ac.uk>
+- Robin Heinemann <robin.ole.heinemann@gmail.com>
 - Rushil Patel <rushil.patel@gsacapital.com>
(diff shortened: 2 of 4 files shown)


DrDaveD:
So am I understanding correctly that this supports having newlines within a `--mount` option to specify multiple mounts?  I'm not sure we want to set a precedent for that for just this one option.  Can you give a good reason why this should be different than other options?  You can alternatively use multiple `--mount` options.

rroohhh:
This is about setting it via the `APPTAINER_MOUNT` environment variable, which is already document as supporting multiple, newline delimited entries: https://apptainer.org/docs/user/main/bind_paths_and_mounts.html#mount-examples
> Mount specifications are also read from then environment variable $APPTAINER_MOUNT. Multiple bind mounts set via this environment variable should be separated by newlines (\n).

There is also already a test for this: https://github.com/apptainer/apptainer/blob/e50ee8fb4916fae45681712c17cfe54d2f0d7af5/pkg/runtime/engine/apptainer/config/mount_test.go#L228-L243 it just use equal number of options for each row. 

Git blame points me to this commit: https://github.com/apptainer/apptainer/commit/d27ff7a6f5aa6da78302f5372e77ef6df63eb556 by you :)

This "just" allows these multiple entries to have a different number of options, which previously was not the case, as by default the csv reader enforces an equal amount of records per row.

DrDaveD:
Ah, I see.  I imported that from Sylabs long ago in apptainer/singularity#6283 and had forgotten about that capability.  It looks like their code (now in [pkg/util/bind](https://github.com/sylabs/singularity/blob/main/pkg/util/bind)) still probably has the same issue.

DrDaveD (changes requested):
Thanks for the contribution.  Please add a CHANGELOG.md entry in the `v1.5.x changes` section.  Make it user-focused, saying something like `Fixed a bug that prevented multiple mount requests in $APPTAINER_MOUNT from having different numbers of options`.

DrDaveD (changes requested):
Please rebase to fix the CHANGELOG conflict.

DrDaveD (approved):

DrDaveD:
Please also make a cherry-pick PR for release-1.5, although note that I'm also waiting on #3670 to do the same and if they're both done in parallel, whichever one is merged second will have to rebase again.

PR #3674, opened 2026-08-03
Add --writable-tmpfs-size argument

### LINKED ISSUE #1313: `--writable-tmpfs` space often too small
This is a continuation of apptainer/singularity#578.

Even though with apptainer-1.2.0 there is an [increase of the default size](#1116) for `--writable-tmpfs` from 16MB to 64MB, that's still not much space for people who want to use an overlay.  Possibly there should be an option for people to expand that size, or possibly a new `--writable-scratch` option that keeps the overlay in scratch disk space (default under `${APPTAINER_SCRATCHDIR:-${APPTAINER_SCRATCH:-${APPTAINER_TMPDIR:-/tmp}}}`) instead of memory.

mahnerak:

This PR adds `--writable-tmpfs-size` argument to control the size of the disposable overlay that is created, simialar how #1313 and https://github.com/apptainer/singularity/issues/5718 discussed.

This would be particularly useful for one-line commands. While `--overlay` offers similar functionality, it requires manually running `mkdir /tmp/overlay` (or `apptainer overlay create`), launching Apptainer, and then performing cleanup. The cleanup process is especially tricky with `--fakeroot`, as it can create files owned by sub-UIDs that a standard `rm` cannot delete (I have to do this multiple times a day).

In my experience, increasing the `tmpfs` size to 10GB in `apptainer.conf` was a helpful workaround, but it requires creating a custom configuration file and setting `APPTAINER_CONFIGDIR`.

It would be ideal to have direct control over the size via command-line arguments, especially since everything runs in userspace for non-setuid installations. A feature similar to `docker run --rm` that automatically cleans up leftovers would be incredibly helpful.

 - Fixes #1313

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 Changes since 1.5.x
 
+- Add a `--writable-tmpfs-size` option, and a corresponding
+  `APPTAINER_WRITABLE_TMPFS_SIZE` environment variable, to override the
+  `sessiondir max size` directive that limits `--writable-tmpfs`. It is
+  ignored when running setuid, where the session directory size is set by
+  the administrator.
 - Update minimum go version to 1.26.5.
 - Add support for variant to more architectures, but without verification.
   The previous arm32v5, arm32v6 and arm32v7 (= "arm") are still verified.
--- a/cmd/internal/cli/action_flags.go
+++ b/cmd/internal/cli/action_flags.go
@@ -85,6 +85,7 @@ var (
 	memorySwap        string // bytes
 	oomKillDisable    bool
 	pidsLimit         int
+	writableTmpfsSize uint32
 	unsquash          bool
 
 	ignoreSubuid      bool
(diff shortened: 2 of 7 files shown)


mahnerak:
@DrDaveD qq: is uint32 here fine? If we go with uint64 then new `registerUint64Var` needs to be added.

I have a `--writable-scratch` implementation on my side, but deliberately wanted to keep these as separate PRs.

DrDaveD:
Even though this is implementing a fix for an issue I made (long ago), I am extremely hesitant to make it too easy for people to use large `--writable-tmpfs` partitions.   In general it's not a very good or efficient way to use apptainer, and people can accidentally use up large amounts of RAM on shared systems.  Maybe if it were a hidden, undocumented option, or maybe if there was another apptainer.conf parameter that put a limit on how large of a tmpfs people could request ...

A better feature might be the other option you're talking about with `--writable-scratch`, an option to use temporary scratch space in $APPTAINER_TMPDIR rather than a tmpfs, although as you pointed out that could be problematic at cleanup time when using sub-UIDs.  Using an overlay ext3 partition helps with the cleanup, but it has the additional problem of having poor performance.

What are your use cases where you want large `--writable-tmpfs`?  Maybe a different approach to them would be better.

DrDaveD:
> a limit on how large of a tmpfs people could request

Probably that would be based on a percentage of system RAM, defaulting to maybe 20%.

DrDaveD:
Or maybe it would default to the amount of system RAM divided by the number of cores.

valentino-aguiar-gsa:
Hi, we effectively have created --writable-scratch internally via a custom slurm prolog + epilog. It utilises an ext3 overlay which gets created at job start and subsequently cleaned up at job end. I agree that the more useful thing for us would be --writable-scratch where we can use it disk backed rather than writable-tmpfs. Nothing stops you from using --writable-scratch pointed at a memory backed file path in any case.

Removing our custom logic would be awesome.

PR #3697, opened 2026-08-12
feat: Make ldconfig optional for GPU library discovery

### LINKED ISSUE #1894: --nv: make ldconfig optional in the nvliblist.conf
### Version of Apptainer

```
❯ apptainer --version
apptainer version 1.2.5
```

### Expected behavior

E.g. be able to manually specify - at build time or at runtime - a list of search paths for `--nv` to scan for the sonames listed in the `nvliblist.conf`. In this case, apptainer would not assume that `ldconfig` be in the `PATH`

### Actual behavior

Apptainer expects https://github.com/apptainer/apptainer/blob/dbaf1afa0e153e056c32dad2640b4d367a53ff14/internal/pkg/util/paths/resolve.go#L163 to exist and to be aware of the paths to the host's nvidia userspace driver libraries.

As far as I can tell, `ldconfig` isn't really a way to inspect the `ld.so`'s search paths, it's more of a way to generate the optional cache to speed up the search. Ldconfig would, for example, ignore the `LD_LIBRARY_PATH` and `LD_PRELOAD`, which the dynamic loader would have respected, so these are inherently different interfaces. It's maybe relevant that, although apptainer doesn't rely on following, the `/etc/ld.so.conf` isn't guaranteed to exist, and the `ld.so.cache` is internal to glibc (not necessarily intended as a part of the public interface)

I think the reasonable way forward would be to open a ticket in glibc's issue tracker and ask for their position on how to inspect ld.so's search paths, specifically how to locate "the host libraries that the dynamic loader would've used". The most obvious solution would be `dlopen` followed by `dlinfo(..., RTLD_DI_ORIGIN, ...)`, although that does execute the code (`.init` and constructors) from the said libraries. We'd also have to carefully explain why do we want to mount the host's libraries in the container, and that we're content with assuming that the container's libc is compatible with the host's.

```sh
❯ cat /etc/os-release 
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="24.05.20231224.5f64a12"
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 24.05 (Uakari)"
SUPPORT_URL="https://nixos.org/community.html"
VERSION="24.05 (Uakari)"
VERSION_CODENAME=uakari
VERSION_ID="24.05"
❯ nvidia-container-cli list
...
/nix/store/9x68sk5a6xc86q7fc6w8n40izdpwgv1q-nvidia-x11-545.29.06-6.1.69-lib32/lib/libcuda.so.545.29.06
...
❯ nvidia-container-cli info
NVRM version:   545.29.06
CUDA version:   12.3

Device Index:   0
Device Minor:   0
Model:          NVIDIA GeForce RTX 3090
...
❯ ls /run/opengl-driver/lib
...
libcuda.so  libcuda.so.1  libcuda.so.545.29.06
...
```

Via the NixOS configuration:

```
    programs.singularity.enable = true;
    programs.singularity.enableSuid = false;
    programs.singularity.package = pkgs.apptainer;
```

...but also just built separately: `nix build nixpkgs#apptainer`, `./result/bin/apptainer`

ehfd:
Reviewer: @DrDaveD 

Apptainer locates the GPU driver libraries named in nvliblist.conf and rocmliblist.conf by running "ldconfig -p". That fails on systems which do not populate an ld.so cache, or which do not ship ldconfig at all, such as NixOS and Guix, where the driver libraries live in unusual paths. On those hosts --nv and --rocm find nothing to bind.

Add a "gpu library path" option to apptainer.conf, holding a list of directories to search for those libraries. The configured directories are searched first, in order, and the ld.so cache becomes an optional secondary source: when ldconfig cannot be found or run, the configured directories are used on their own instead of the lookup coming up empty.

Entries which do not resolve, such as dangling symlinks, are skipped so that they cannot shadow a working library of the same name found in a lower priority location.

The warning issued when no libraries can be found now names the setting to configure, rather than only reporting that the ld cache is unreadable.

 - Fixes #1894

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,15 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 Changes since 1.5.x
 
+- Add a new `gpu library path` option to `apptainer.conf`, giving a list of
+  directories to search for the GPU driver libraries named in
+  `nvliblist.conf` and `rocmliblist.conf` when binding them with `--nv` or
+  `--rocm`. The directories are searched ahead of the ld.so cache, and make
+  `ldconfig` optional: when it cannot be found or run, the configured
+  directories are used on their own. This allows the host GPU libraries to be
+  found on systems that keep them outside the ld.so cache, or that do not
+  ship `ldconfig` at all, such as NixOS and Guix, where previously no host
+  GPU libraries would be found at all.
 - Add a new `--compat32` flag (action and build) which, alongside `--nv` or
   `--rocm`, additionally binds the 32-bit GPU driver libraries of the
   host into `/.singularity.d/libs32` in the container, and adds that
--- a/pkg/util/apptainerconf/config.go
+++ b/pkg/util/apptainerconf/config.go
@@ -106,6 +106,7 @@ type File struct {
 	AlwaysUseNv               bool     `default:"no" authorized:"yes,no" directive:"always use nv"`
 	UseNvCCLI                 bool     `default:"no" authorized:"yes,no" directive:"use nvidia-container-cli"`
 	AlwaysUseRocm             bool     `default:"no" authorized:"yes,no" directive:"always use rocm"`
+	GpuLibraryPath            []string `directive:"gpu library path"`
 	SharedLoopDevices         bool     `default:"no" authorized:"yes,no" directive:"shared loop devices"`
 	MaxLoopDevices            uint     `default:"256" directive:"max loop devices"`
 	SessiondirMaxSize         uint     `default:"64" directive:"sessiondir max size"`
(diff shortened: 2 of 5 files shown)


ehfd:
`Assisted-by:` added here as well.

DrDaveD (changes requested):
Please rebase on the current main branch.

ehfd:
Done.

DrDaveD (changes requested):
In addition to the unit tests, please add an e2e test or tests of the new functionality, mostly for completeness and to make sure the apptainer.conf directive works.  I guess it requires a GPU to be tested so it won't get run by the usual github action .... but that reminds me that we should probably be able to use an HPSF hosted runner with a GPU, which I should look into.

ehfd:
@DrDaveD Done.

pkg/util/apptainerconf/config.go
```diff
@@ -537,6 +538,23 @@ use nvidia-container-cli = {{ if eq .UseNvCCLI true }}yes{{ else }}no{{ end }}
 # environments).
 always use rocm = {{ if eq .AlwaysUseRocm true }}yes{{ else }}no{{ end }}
 
+# GPU LIBRARY PATH: [STRING]
+# DEFAULT: undefined
+# A comma separated list of directories to search for the GPU driver libraries
+# named in nvliblist.conf and rocmliblist.conf, when binding them into a
+# container with --nv, --rocm, or --compat32.  These directories are searched
```
DrDaveD: `--compat32` is not really an "or", right?  That still requires one of `--nv` or `--rocm` as I understand it.  If that's correct, then it should be dropped from this list.

CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 Changes since 1.5.x
 
+- Add a new `gpu library path` option to `apptainer.conf`, giving a list of
+  directories to search for the GPU driver libraries named in
+  `nvliblist.conf` and `rocmliblist.conf` when binding them with `--nv`,
+  `--rocm`, or `--compat32`. The directories are searched ahead of the ld.so
```
DrDaveD: Likewise here.

DrDaveD (changes requested):

ehfd:
@DrDaveD Both done.

ehfd:
@DrDaveD Anything left for me to do? Just asking.

DrDaveD (approved):

PR #3738, opened 2026-09-03
Run CDI device hooks and keep --device with --nv and --nvccli

ehfd:
## Description of the Pull Request (PR):

**This is a parity fix to https://github.com/NVIDIA/nvidia-container-toolkit/pull/1980 and https://github.com/NVIDIA/nvidia-container-toolkit/pull/1979.**

`--device` applies the mounts, device nodes and environment of a CDI device but runs none of its hooks. The specification `nvidia-ctk cdi generate` writes relies on two `createContainer` hooks: `nvidia-cdi-hook create-symlinks` creates the driver's soname links (`libcuda.so.1`, `libGLX_nvidia.so.0`, `gbm/nvidia-drm_gbm.so`, ...) and `nvidia-cdi-hook update-ldcache` refreshes the container's `ld.so.cache`. Without them the driver's files are in the container but nothing resolves them: GLX clients fail with "failed to create dri3 screen", Wayland compositors with "Failed to create EGL display", and `libcuda.so.1` is not found.

This PR:

- Runs the `createRuntime` and `createContainer` hooks a CDI device declares from the RPC server, which is in the container's mount namespace, once every mount is in place and before the chroot. The hooks are handed an OCI state on stdin whose bundle holds a `config.json` naming the container root (`<session>/final`), which is where the NVIDIA hooks read it from. The other hook kinds have no counterpart in this engine and are skipped. A device that declares hooks turns on `--writable-tmpfs` the way `--nvccli` does, since the hooks write into the container root.
- Keeps the hook's capabilities across the exec. The RPC server of an unprivileged installation is not uid 0 in its user namespace, so an exec'd hook lost every capability; `update-ldcache` clones new mount and PID namespaces and failed with `fork/exec /proc/self/exe: operation not permitted`. `exec.Hook` now raises the process's permitted capabilities that are within its bounding set into the child's ambient set, which for the RPC server is `CAP_SYS_ADMIN`, `CAP_IPC_LOCK` and `CAP_MKNOD` in that namespace: the same set a hook exec'd by a root RPC server would receive. A failing hook's stderr is now part of the error instead of being discarded.
- Does not run the hooks in the setuid flow, for the reason nvidia-container-cli is refused there, and warns instead; the device's mounts, nodes and environment still apply as before.
[...]

PR #3740, opened 2026-09-03
Bind the NVIDIA GBM backend and X server modules with --nv and --nvccli

ehfd:

`--nv` binds the host's NVIDIA driver libraries into one directory on the container's library path. That covers everything the dynamic loader opens by *name*, which is what CUDA and OpenGL need. It does not cover the parts of the driver that a program opens by *path*: Mesa's libgbm loads a GBM backend out of a backend directory, and an X server loads `nvidia_drv.so` and `libglxserver_nvidia.so` out of its module directory. Those files sit in directories the loader cache never lists, so `--nv` never found them, and `nvliblist.conf` had fallen behind the driver's current file set.

In practice, on a stock host, that means a Wayland compositor started with `--nv` cannot allocate a GPU buffer and falls back to software rendering, and an X server has no `nvidia` driver to load. `--nvccli` and CDI already handled this; plain `--nv` did not.

This PR makes `--nv` and `--nvccli` work for graphics as well as for compute:

- **The modules opened by path are found and bound.** The module directories next to the library directories are searched, and each module is bound both at its host path and among the libraries, so a container laid out like the host finds them where its X server and libgbm look. `GBM_BACKENDS_PATH` is set so the driver's backend is used ahead of the container's own.
- **`/dev/dri` is bound under `--contain`.** Those are the devices GBM, EGL and an X server open on the GPU; only `--rocm` bound them before. `--nvccli` binds the nodes of just the GPUs `NVIDIA_VISIBLE_DEVICES` selects, so a job given one GPU does not see the others.
- **Device nodes the host is missing are created before they are bound**, through `nvidia-modprobe`, the setuid helper the driver ships for this and that the driver's own libraries call on first use. A headless host does not create the modeset or unified memory nodes until something asks for them, and a container without them cannot present through Vulkan. A helper that is absent or refuses leaves the nodes as they are.
- **`nvliblist.conf` is brought up to the current driver**, and a configuration file it names is bound at the standard path the container's loaders read, wherever the host happens to keep it. The `libglx.so` entry is dropped: that was the driver's GLX module name before `libglxserver_nvidia.so`, and it would now shadow the X server's own `libglx.so`.
- **`--nvccli` accepts `NVIDIA_DRIVER_CAPABILITIES=all`**, the documented value for every capability. It previously refused the container with `unknown NVIDIA_DRIVER_CAPABILITIES value: all`.
- When a CDI device given alongside `--nv` or `--nvccli` mounts the same driver, the modules are left to that device, whose hooks install them; a bind mount in a symlink's place would block the hook.

This is the Apptainer counterpart of NVIDIA/nvidia-container-toolkit#1980, NVIDIA/nvidia-container-toolkit#1979 and NVIDIA/libnvidia-container#388, which make the same changes in the NVIDIA Container Toolkit.

## Testing

Host: 2x Tesla P100-SXM2-16GB, driver 580.178.04, Ubuntu 26.04, unprivileged user namespaces, NVIDIA Container Toolkit 1.20.0.

Inside `apptainer run --nv --contain` of a desktop image, before this PR: no GBM backend, no X server modules and no `/dev/dri`; a Wayland session reports `Failed to allocate GBM buffer` and falls back to software, and an X server has no `nvidia` module to load. After it: all of them present, and both a Wayland and an X11 session report the GPU on `/dev/dri/renderD128` using the driver's own GLX and EGL. `--nvccli` gives the same result, and `NVIDIA_DRIVER_CAPABILITIES=all` starts the container instead of failing.

Unit tests cover the module search, the configuration-file lookup, the host-path binds, the `NVIDIA_VISIBLE_DEVICES` forms, the DRM node selection and the `nvidia-modprobe` calls. `go vet`, `golangci-lint` and the unit tests of the touched packages pass; the GPU end-to-end group passes on the host above, and the added cases skip where the host lacks the file or the driver's report.

- None filed.

- [x] Made sure all commits are signed-off with `git commit -s`; see the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md#getting-started).
- [x] Either did not use AI or followed the AI Asistance Policy as found in the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md#ai-assistance-policy).
- [x] Added changes to the [CHANGELOG](https://github.com/apptainer/apptainer/blob/main/CHANGELOG.md), if necessary according to the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md#pull-requests-prs).
- [x] Added tests to validate this PR, linted with `make check` and tested this PR locally with a `make test`, and `make testall` if possible (see the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md#pull-requests-prs)). The unit tests of the touched packages and the GPU end-to-end group were run; the full `make test` and `make testall` were not.
- [x] Based this PR against the appropriate branch according to the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md#pull-requests-prs).


e2e/gpu/gpu.go
```diff
@@ -109,6 +109,22 @@ func (c ctx) testNvidiaLegacy(t *testing.T) {
...
+	backends, _ := filepath.Glob("/usr/lib*/gbm/nvidia-drm_gbm.so")
+	multiarch, _ := filepath.Glob("/usr/lib/*/gbm/nvidia-drm_gbm.so")
+	if len(backends)+len(multiarch) > 0 {
+		c.env.RunApptainer(
+			t,
+			e2e.AsSubtest("GBMBackend"),
+			e2e.WithProfile(e2e.UserProfile),
+			e2e.WithCommand("exec"),
+			e2e.WithArgs("--nv", imagePath, "sh", "-c", "ls /.singularity.d/libs/nvidia-drm_gbm.so && echo $GBM_BACKENDS_PATH"),
+			e2e.ExpectExit(0, e2e.ExpectOutput(e2e.ContainMatch, "/.singularity.d/libs:/")),
```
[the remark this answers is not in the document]
ehfd: Fixed in f2fda843a: the assertion is now the multiline regex, so it checks the directory leads the variable and accepts a single-entry value.

internal/pkg/util/fs/files/action_script.sh
```diff
@@ -174,6 +174,24 @@ if [ -d /.singularity.d/libs32 ] && [ "$(cd /.singularity.d/libs32 && echo *)" !
...
+# the dri backend is still found.
+for __backend__ in /.singularity.d/libs/*_gbm.so; do
+    if [ -e "$__backend__" ]; then
+        __gbm_dirs__="/.singularity.d/libs"
+        for __dir__ in $( (/sbin/ldconfig -Nv | /bin/sed '/^\t/d;s/:.*//') 2>/dev/null ); do
+            if [ -d "$__dir__/gbm" ]; then
+                __gbm_dirs__="$__gbm_dirs__:$__dir__/gbm"
+            fi
+        done
+        export GBM_BACKENDS_PATH="${GBM_BACKENDS_PATH:+$GBM_BACKENDS_PATH:}$__gbm_dirs__"
```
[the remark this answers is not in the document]
ehfd: Fixed in f2fda843a: the bound directory now comes first, ahead of the container's own gbm directories and then any value the image had set; measured with a preset value as /.singularity.d/libs:<container gbm>:<preset>.

DrDaveD:
I haven't had a chance yet to look at this in detail, but rebasing on the current main should fix the CI failure.

ehfd:
Done for both PRs, thanks for your tip @DrDaveD

DrDaveD:
Unfortunately I haven't had time to review this this week.  I'm mostly caught up with other stuff though so should be able to do it early next week.

DrDaveD (changes requested):
I'm having a very difficult time with understanding the description of this PR.  It's clearly AI-written, and hopefully you understand it, but much of it makes no sense to me.  Please go over it carefully and condense it down to the useful information that can be understood by users who are not GPU experts.

CHANGELOG.md
```diff
@@ -9,6 +9,43 @@ For older changes see the [archived Singularity change log](https://github.com/a
 
 Changes since 1.5.x
 
+- Bind the NVIDIA driver's GBM backend and X server modules with `--nv`, so
```
DrDaveD, line 12: This entry is much too long and detailed.  Please rewrite it, without an AI, to contain the essentials that users need to know to understand what the change is.

ehfd:
@DrDaveD Sorry about that; this is the result of porting PR body and comments from nvidia-container-toolkit where NVIDIA devs reviewed it. I'll edit it as accordingly.

CHANGELOG.md
ehfd: Rewritten, and carefully reviewed.

ehfd:
@DrDaveD Let me know if the descriptions are still confusing. This was a direct port of https://github.com/NVIDIA/nvidia-container-toolkit/pull/1980, so I didn't realize I had to use a different language to the one targeting NVIDIA maintainers.

etc/nvliblist.conf
```diff
@@ -73,14 +90,31 @@ libnvtegrahv.so
 libOpenCL.so
 libOpenGL.so
 libvdpau_nvidia.so
+# modules loaded by path, found in the module directories next to the libraries
```
DrDaveD: I don't understand how these can be distinguished as modules loaded by path, since they look like all the others.  Wouldn't there need to be a full path in order for apptainer to tell the difference?

internal/pkg/runtime/launch/launcher_linux.go
```diff
@@ -938,6 +941,37 @@ func (l *Launcher) setNvCCLIConfig() (err error) {
...
+		if l.driverStagedByDevices(libs) {
+			files = paths.WithoutModuleBinds(files)
+		}
+		if len(files) > 0 {
+			l.engineConfig.AppendFilesPath(files...)
+		}
+		platformLibs := []string{}
+		for _, lib := range libs {
+			base := filepath.Base(lib)
+			if strings.HasPrefix(base, "libnvidia-egl-") || strings.HasSuffix(base, "_gbm.so") {
```
DrDaveD: I don't like having exceptions explicit in the code like this.  We need a more generic mechanism somehow.

If a full path doesn't work, then maybe there needs to be some way of indicating this type of exception in nvliblist.conf, for example something starting with `**/` or `.../` to mark it as a special entry.

internal/pkg/util/fs/files/action_script.sh
```diff
@@ -174,6 +174,25 @@ if [ -d /.singularity.d/libs32 ] && [ "$(cd /.singularity.d/libs32 && echo *)" !
...
+        __gbm_dirs__="/.singularity.d/libs"
+        for __dir__ in $( (/sbin/ldconfig -Nv | /bin/sed '/^\t/d;s/:.*//') 2>/dev/null ); do
+            if [ -d "$__dir__/gbm" ]; then
+                __gbm_dirs__="$__gbm_dirs__:$__dir__/gbm"
+            fi
+        done
+        export GBM_BACKENDS_PATH="$__gbm_dirs__${GBM_BACKENDS_PATH:+:$GBM_BACKENDS_PATH}"
+        sylog debug "Setting GBM_BACKENDS_PATH to $GBM_BACKENDS_PATH"
+    fi
+    break
```
DrDaveD: I'm also not very happy with such a specific magic path name of `*_gbm.so` in this generic action script, although at the moment I don't have ideas about what to do about it.

internal/pkg/util/paths/resolve.go
```diff
@@ -281,10 +307,133 @@ func libraryCache() (map[string][]string, error) {
...
 
+// moduleSubdirs are the directories, relative to a library directory or its
+// parent, in which a driver installs the modules that are loaded by path
+// rather than through the dynamic loader, so the ld cache never lists them:
+// the GBM backend and the X server's driver and GLX modules.
+var moduleSubdirs = []string{
+	"gbm", "nvidia/xorg",
+	"xorg/modules/drivers", "xorg/modules/extensions",
+	"xorg/modules/updates/drivers", "xorg/modules/updates/extensions",
+}
```
DrDaveD: I think these should be configurable in nvliblist.conf somehow, not hardcoded.

internal/pkg/util/paths/resolve.go
```diff
@@ -281,10 +307,133 @@ func libraryCache() (map[string][]string, error) {
...
+		if strings.HasSuffix(dir, "/"+sub) {
+			return true
+		}
+	}
+	return false
+}
+
+// configRoots are the roots a distribution keeps the same configuration
+// trees under; a file named under one is looked for under the others.
+var configRoots = []string{"/etc", "/usr/share", "/usr/local/share"}
```
DrDaveD: It would be better for these to be runtime configurable too, but possibly we could do away without it.

DrDaveD (approved):
This description and changelog entries are much more understandable, thank you.  Now that I can see a little better what the goals are, I have more comments especially related to generic configurability.

etc/nvliblist.conf
ehfd: Right, the names alone could not tell them apart; the code went by the directory it had found them in, from a hard-coded list. Now the list itself says which entries are modules, by giving them as the path relative to a library directory, e.g. `gbm/nvidia-drm_gbm.so` or `xorg/modules/drivers/nvidia_drv.so`. So a bare name is looked up by name in the ld cache, an absolute path is a file, and a relative path is looked for under that path next to each library directory (and next to its parent, since Debian keeps `/usr/lib/xorg/modules` beside `/usr/lib/<multiarch>`). The plain `nvidia_drv.so` and `libglxserver_nvidia.so` entries are gone, and the file's header and the section comment explain the forms.

internal/pkg/runtime/launch/launcher_linux.go
ehfd: Agreed. The names are gone from the code: with `--nvccli` the launcher now runs `nvidia-container-cli list --libraries`, which prints everything the CLI would stage for the driver, and binds only the entries of `nvliblist.conf` that are not in that output (compared by resolved path), so nothing the CLI stages is shadowed, CUDA forward compatibility included. On my host that leaves the driver libraries to the CLI and binds the four `libnvidia-egl-*` platform libraries, the modules and the configuration files. One side effect to be aware of: the glvnd and OpenCL loader libraries that the list names are now bound under `--nvccli` as they always were under `--nv`. The changelog says so.

internal/pkg/util/fs/files/action_script.sh
ehfd: The glob is gone. A module now keeps its relative path under `/.singularity.d/libs`, so the backend lands in `/.singularity.d/libs/gbm/`, and the script only checks whether that directory exists, the same way it checks `libs` and `libs32` before adding them to `LD_LIBRARY_PATH`. The one thing it still has to know is that `gbm` is the directory Mesa's `GBM_BACKENDS_PATH` points at.

internal/pkg/util/paths/resolve.go
ehfd: Done: the directories now come from the entries in `nvliblist.conf` (see the thread on the conf file), and `moduleSubdirs`, `moduleDirs` and `isModulePath` are removed. A site whose driver keeps the X modules somewhere else adds a line to the list.

internal/pkg/util/paths/resolve.go
ehfd: Settled by removing them. A file the list names is looked for at its own path, and otherwise under the prefix of a `gpu library path` directory (`/usr/share/x` as `<prefix>/share/x`, `/etc/x` as `<prefix>/etc/x`) for a driver installed elsewhere; there is no other table of directories in the code. The list names each location a loader reads, `/etc` and `/usr/share`, as it already did for Vulkan, so where to look is entirely in `nvliblist.conf` and `apptainer.conf`.

ehfd:
@DrDaveD Thanks for the review. I have rebased on `main` (the CHANGELOG conflict with #3738) and reworked the parts you commented on so that nothing driver-specific is hard-coded: the modules are `nvliblist.conf` entries given relative to the library directory, the action script keys on the `gbm` directory that layout produces, `--nvccli` asks `nvidia-container-cli list` what it stages instead of naming exceptions, and the fixed configuration roots are gone. Details in the threads. Verified again on the P100 host (unit tests, lint, the GPU e2e group, and `eglinfo` reporting the NVIDIA GBM platform under both `--nv` and `--nvccli`), and CI is green. The documentation for the new `nvliblist.conf` entries and the graphics support is in apptainer/apptainer-admindocs#250 and apptainer/apptainer-userdocs#420.

ehfd:
GPU runner issues are unrelated to this PR; identical to `main`.

DrDaveD (approved):
This looks OK to me know but I would also like @cclerget and @afbjorklund to review before merging.

internal/pkg/util/fs/files/action_script.sh
DrDaveD: Alright I guess that's the best we can do.

PR #3752, opened 2026-09-06
Allow oras to upload multiple images with index

### LINKED ISSUE #3751: Implement multi-arch image push for ORAS
Currently `apptainer push` only takes one image:

`apptainer push [push options...] <image> <URI>`

It should be possible to push _multiple_ images, each for a different arch, to the _same_ URI:

`apptainer push [push options...] <image> ... <URI>`

----

Then each image should be pushed, almost as before*, and the digest (descriptor) recorded.
\* but with a new name or tag, it should either be `name-arch:*` - or alternatively `*:tag_arch`

At the end, a v1.ImageIndex should be created with all the v1.Image that were pushed before.
And finally, this v1.ImageIndex should be pushed to the original URI (that has to be `oras://`)

This replaces the current manual workflow: https://oras.land/docs/how_to_guides/multiarch

Instead, one can do a single: `apptainer push image-amd64.sif image-arm64.sif oras://image`

After that, **both** are available: `apptainer pull image.sif oras://image` (will pick the right arch!)

This means that the URI will now be multiarch (even though the actual SIF will be for a single arch)

The go-containerregistry code involved is similar to this crane command: https://github.com/google/go-containerregistry/blob/main/cmd/crane/recipes.md#create-a-multi-platform-image-from-scratch

```
crane index append -t ${IMAGE} \
  -m ubuntu@sha256:c985bc3f77946b8e92c9a3648c6f31751a7dd972e06604785e47303f4ad47c4c \
  -m ubuntu@sha256:61bd0b97000996232eb07b8d0e9375d14197f78aa850c2506417ef995a7199a7
```

----

The image index is a simple JSON file, see `oras manifest fetch` for the exact details (OCI spec)

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:ce54db371df60b95e3b7088ddbedd64c333424f60994dd3966a37bed51dbdecf",
      "size": 595,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:fe8e9d0b66194ae46bc0cb3f4c5da639842be98d8b9b3452c893ed72c2c59e6d",
      "size": 595,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    }
  ]
}
```

The temporary names/tags are not important, and could be deleted at the end....
The images are fetched using their digest, so they don't _need_ individual references.

Locally you could do something more simple, like a variable: `image-${ARCH}.sif`

EDIT: It is probably a bad idea to remove the arch name/tag, since the server will GC.
(Not all servers reflect what is used by ImageIndex, so they might delete the Image)

afbjorklund:
When uploading more than one image, create an image index.

Set platform in the index, but don't set the ArtifactType.

Allows uploading more than one SIF image to an ORAS URI:

`apptainer push [push options...] <image> ... <URI>`

```console
$ apptainer pull --arch amd64 alpine-amd64.sif docker://alpine
$ apptainer pull --arch arm64 alpine-arm64.sif docker://alpine
$ apptainer push alpine-amd64.sif alpine-arm64.sif oras://localhost:5000/alpine_sif
INFO:    Digest: sha256:6ce5455fc3cb4ccfafc711f0054e269e3d919397b37c57293dd2368643b18d2f (amd64)
3.6MiB / 3.6MiB [==================================================================================================================================================================================] 100 %0s
INFO:    Digest: sha256:7e887bf28123c79b3a98ab68e55a334805c0c5f2a7442ac0e226fc445418e514 (arm64)
3.9MiB / 3.9MiB [==================================================================================================================================================================================] 100 %0s
INFO:    Digest: sha256:10eb0c20b6ab70ccb73fb9f69020e998e4611bd7724c45b9007b7abafeec684e
INFO:    Upload complete
$ oras manifest fetch --pretty localhost:5000/alpine.sif@sha256:10eb0c20b6ab70ccb73fb9f69020e998e4611bd7724c45b9007b7abafeec684e
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 0,
      "digest": "sha256:6ce5455fc3cb4ccfafc711f0054e269e3d919397b37c57293dd2368643b18d2f",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "size": 0,
      "digest": "sha256:7e887bf28123c79b3a98ab68e55a334805c0c5f2a7442ac0e226fc445418e514",
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    }
  ]
}
$ apptainer search oras://localhost:5000/alpine_sif
Found 3 container images for amd64 matching "localhost:5000/alpine_sif":

	oras://localhost:5000/alpine_sif:latest

	oras://localhost:5000/alpine_sif:latest_amd64

	oras://localhost:5000/alpine_sif:latest_arm64
```

Seems to be some bug with search and arch, but...

~~EDIT: And the "size" seems to be wrong (0), as well~~ FIXED

 - Fixes #3751 

- [x] Made sure all commits are signed-off with `git commit -s`; see the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md#getting-started).
- [x] Either did not use AI or followed the AI Asistance Policy as found in the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md#ai-assistance-policy).
- [x] Added changes to the [CHANGELOG](https://github.com/apptainer/apptainer/blob/main/CHANGELOG.md), if necessary according to the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md#pull-requests-prs).
- [ ] Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md#pull-requests-prs)).
- [x] Based this PR against the appropriate branch according to the [Guidelines for Contributing](https://github.com/apptainer/apptainer/blob/main/CONTRIBUTING.md#pull-requests-prs).

### DIFF
--- a/internal/pkg/client/oras/oras.go
+++ b/internal/pkg/client/oras/oras.go
@@ -34,6 +34,8 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/empty"
 	"github.com/google/go-containerregistry/pkg/v1/layout"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/partial"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"golang.org/x/term"
 )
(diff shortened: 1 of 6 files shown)


DrDaveD:
Please rebase to fix ci issue

DrDaveD:
Would it be possible to upload different architecture images at different times to the same URL?  Or can it only be done in a single operation?

afbjorklund:
It would be possible, but it is not implemented here.

If you push one arch first, and then another arch - you would have to download them again in order to push the multi-arch, there is no way to reference an already existing digest (with apptainer)

You could always do it manually, though (the old way)

afbjorklund:
> Would it be possible to upload different architecture images at different times to the same URL?

If you wanted to squeeze it in, it could perhaps be added to the `apptainer tag` command

We could it add it as a separate issue, as a special case and a special registry optimization:

`apptainer tag oras://localhost:5000/alpine_sif:latest_amd64 oras://localhost:5000/alpine_sif:latest_arm64 latest`

Currently it would assume that both input images would use the same image name, though.

Or it would have to come up with some other way, like accepting an "absolute" (URI) for tag?

```
apptainer tag oras://localhost:5000/alpine_sif:latest_amd64 oras://localhost:5000/alpine_sif:latest_arm64 oras://localhost:5000/alpine_sif:latest
```

And try to parse "tag" as a reference instead:
```go
func TagImage(ctx context.Context, ref, tag string, ociAuth *authn.AuthConfig, noHTTPS bool, reqAuthFile string) error {

        ir, err := name.ParseReference(ref, opts...)

        it := ir.Context().Tag(tag)
```

EDIT: I remember now the problem with this approach, ORAS doesn't know the platform of a SIF image!

So you would have to download it first anyway, which negates the advantage of using the "tag" command.

internal/pkg/client/oras/oras.go
```diff
@@ -213,6 +215,142 @@ func UploadImage(ctx context.Context, path, ref, arch string, ociAuth *authn.Aut
 	return remote.Write(ir, im, remoteOpts...)
 }
 
+// UploadImages uploads the images specified by paths and pushes them along with an index to the provided oci reference,
+// it will use credentials if supplied
+func UploadImages(ctx context.Context, paths []string, ref string, archs []string, ociAuth *authn.AuthConfig, noHTTPS bool, reqAuthFile string, annotations []string) error {
```
DrDaveD, line 220: This looks like a lot of duplication with `UploadImage()`.  Can `UploadImage()` be re-written to invoke this function instead to avoid duplication?  (Or change the invocation(s) of `UploadImage()` to call this function directly).

internal/pkg/client/oras/oras.go
```diff
@@ -213,6 +215,142 @@ func UploadImage(ctx context.Context, path, ref, arch string, ociAuth *authn.Aut
 	return remote.Write(ir, im, remoteOpts...)
 }
 
+// UploadImages uploads the images specified by paths and pushes them along with an index to the provided oci reference,
+// it will use credentials if supplied
```
DrDaveD, line 219: This should be a separate sentence, not an extra piece tacked on to the previous sentence after a comma.  Yes the same problem exists on the `UploadImage()` comment..

DrDaveD (changes requested):

internal/pkg/client/oras/oras.go
afbjorklund: I think there is some common code that could be refactored out of both of them and be shared

The problem was that 1) there is some internal state that need to be kept 2) not exposing ggcr

internal/pkg/client/oras/oras.go
afbjorklund: I think the difference was only in the two "s" in the parameters

internal/pkg/client/oras/oras.go
afbjorklund: Now common code has been broken out, into helper functions.

DrDaveD (approved):

